PR #8, opened 2016-05-03
Issue/5

gogogadgetbioprocess:
closes issue and adds greater degree of composability by creating NewDNASequence_fromLookup and NewDNASequence_fromFile elements and demoing use with example workflows and inputs.


jmanart:
A reference to the issue you are closing is needed, not just closes issue.

Closes #5 would be the correct method.

jmanart:
There are way too many commits

They can either be simplified, merged into one in some cases.
The comment for each commit should be reviewed.

PR #16, opened 2016-05-10
Errors

### LINKED ISSUE #14: structured error handling in antha protocols
A better method for reporting errors instead of panicking.

Prototype:

  Errorf(format string, args... interface{})

ddn0:
Closes #14 

Uses go test, so 
```sh
go test -v github.com/antha-lang/antha/antha/component/lib
```
should run all example inputs. 

To run a specific set of inputs,
```sh
cd somedir && go test -v github.com/antha-lang/antha/antha/component/lib `pwd`
```

I tried to be fancy and the following cases for inputs are supported:
  * a directory with one workflow and one parameter file
  * a directory with N workflow and N parameter files
  * a directory with 1 workflow and N parameter files

The tests are run in parallel, so there will be some interleaving of different inputs on the console.

Once this is in, there will be a bunch of failing tests as the following cases are not going to be supported:
  * elements that require internet access; travis-ci restricts network access
  * elements that require input


gogogadgetbioprocess:
1. I fixed all so make test_workflows_old passes; 
2. To get those tests to work with lib_test.go I increased the timeout limit 
3. I temporarily removed testelementswithdefaults pending further refactoring so that we have the principal testing in place quicker without waiting to fix for all the default input cases just yet. 
4. There are errors in some of the workflows requiring files which cannot be found e.g. PipetteImage_fromPalette. 

I also noticed some strange behaviour for these also after running the test script:
images generated were found in completely different workflow folders to where they would normally be generated. e.g. cmyk3_plateformat.jpg was generated from workflows/PipetteImage_fromPallette/ but found in workflows/Loadgel/

Issue #15, opened 2016-05-10
Add support for setting asp/dsp speeds when mixing


m-i-sadowski:

gogogadgetbioprocess:
Issue now closed

PR #17, opened 2016-05-10
Ca protocol additions

m-i-sadowski:
Required additions for construct assembly protocol: 

- new parameters to allow setting of mix pipetting speeds + xyz offsets
- new dna mix class 
- mix on final addition of DNA to Construct Assembly MMX protocol
- new plate type: greiner V bottom SW

While I was there: 

- refactored well bottoms into consts from regular ints for added clarity
- a few bug fixes


m-i-sadowski:
closes #15

PR #22, opened 2016-05-17
Addxyoffsets

m-i-sadowski:
This branch
- Adds the ability to control X and Y offsets for aspirate, dispense and pre/post mixing

ASP[X/Y/Z]OFFSET
DSP[X/Y/Z]OFFSET
PRE_MIX_[X/Y/Z]
POST_MIX_[X/Y/Z]

- Fixes an issue when combining some consequents: pre-mixing and gentle moves, for instance

- Makes consequents dependent on values of parameters rather than presence of predicates, reducing possibility for errors

- Removes redundant moves when offsets, references etc. match up between mixes

### DIFF
--- a/antha/anthalib/wunit/wdimension.go
+++ b/antha/anthalib/wunit/wdimension.go
@@ -109,6 +109,10 @@ func (v Volume) Dup() Volume {
 	return ret
 }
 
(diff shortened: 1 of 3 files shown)


m-i-sadowski:
Deprecated by PR24

PR #24, opened 2016-05-19
Policydocs

m-i-sadowski:
This branch contains a number of bugfixes and updates for the liquid handling policy framework

additionally I have added a code call to enable the set of permissible liquid handling policy commands to be queried + a command line option to list

This has been tested against some lab protocols but we should ideally physically verify against any which use complex policy frameworks before merging into master

this inherits a lot of changes from addxyoffsets so I wil deprecate that PR in favour of this one. Any testing on that version should apply here also.


m-i-sadowski:
This branch breaks a timing estimate test on the internal microArch code... branch timingestimatechange should be merged after this to fix it

PR #27, opened 2016-05-25
Addplanningerrors

m-i-sadowski:
Mostly additions to liquid handling planner code to ensure that errors are passed back to caller instead of creating calls to logger.Fatal

target/Inst now has Error() which returns any error it has received
the only bullet bitten is in the generator for MixInsts -- in order to pass them back to the front end, mix tasks which generate errors now return MixInsts which do not return nil when Error() is called... attempts to look at generated files without checking this is safe will therefore fail


m-i-sadowski:
Updated to add new instruction type to handle errors

PR #28, opened 2016-05-31
allow serialization of LHRequest

ddn0:
Allow seriali(z/s)ation of LHRequest


m-i-sadowski:
This change is permissible but I warn you that this structure is still a bit volatile and has not been designed to be serializable so fields may need to change. Some fields are stable though, so if we need to agree on areas of volatility vs. stability we should discuss.

PR #39, opened 2016-06-15
dockerize

ddn0:
should simplify installs for some


gogogadgetbioprocess:
Are we going to recommend go 1.6 now?

we need to add brew install pkg-config and brew install go-opencv to install instructions

ddn0:
The Dockerfile already includes opencv.

gogogadgetbioprocess:
And in the native installation instructions do we need to add pkg-config?

brew install homebrew/science/glpk sqlite3 opencv

PR #40, opened 2016-06-15
use go15-style vendoring

ddn0:
Updated README to require go1.6 where GO15VENDOREXPERIMENT is always on


jmanart:
Looking at some internal routes in .an elements

PR #49, opened 2016-06-20
updated new readmes to cover all basic aspects of running antha elements

gogogadgetbioprocess:


CharlotteWard:
Hola - looks really good! The explanations are really clear and there is an obvious flow of what to go onto next. 👍 
There were only three things that I had questions on:
A_sample.an : L31 - I don't understand the difference between the "actual component" and "component itself"
C_samplefortotalvolume.an : L63 "therefoe" typo
A_singleelement/readme.txt : L17 after build may want to put "(recompile)" to keep the language consistent with L12 and avoid confusion.

PR #61, opened 2016-06-21
Fix dup ids

m-i-sadowski:
Migrated 'magic component' setup to new ID-aware system.


m-i-sadowski:
Accidentally incorporates some other changes... will merge with improved volume setup

PR #72, opened 2016-06-29
Find igem parts

gogogadgetbioprocess:
Cleaned up this element and the example search criteria to make it more reliable and the outputs more intuitive,


josi-asae:
The changes to the elements look great! I'm going to do a trial run with the VM to see how it looks than I'll merge.

josi-asae:
Tested, works great!

PR #76, opened 2016-07-05
Extern lib

ddn0:
Package reorg to support custom (statically linked) component libraries. See Synthace/component@3c41c58c53ae841e7df6865cfd1c3fb472993583 for an example of use.


m-i-sadowski:
Looks fine to me, pending conflict fixes

PR #102, opened 2016-07-25
estimate groups of concurrent manual instructions as max of each element

ddn0:


ddn0:
Closes Synthace/monolith#118

PR #107, opened 2016-08-10
Addfilenametype

m-i-sadowski:
Sugaring of output and input filenames... mostly for convenience of front-end

### DIFF
--- a/antha/anthalib/wtype/filename.go
+++ b/antha/anthalib/wtype/filename.go
@@ -1,4 +1,3 @@
 package wtype
 
-type InputFilename string
-type OutputFilename string
+type Filename string
--- a/antha/anthalib/wtype/filename.go
+++ b/antha/anthalib/wtype/filename.go
@@ -1,3 +1,4 @@
 package wtype
 
-type Filename string
+type InputFilename string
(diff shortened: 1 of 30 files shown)


antha/anthalib/wtype/filename.go
```diff
@@ -0,0 +1,4 @@
+package wtype
+
+type InputFilename string
+type OutputFilename string
```
gogogadgetbioprocess: Could the distinction of these into two types become cumbersome?

e.g. in situations where an outputfilename is an inputfilename to a subsequent step?

m-i-sadowski: Yeah, actually i'll just revert to a single type

ddn0:
To allow for expanding the data representation, can we do something like:

```go
package wtype

type File struct {
  Name string
  Data []byte
  ...
}
```

instead of `type Filename string` ?

RFC: changing the name to File because this will be the type we use to transfer data blobs between elements and to the database eventually.

gogogadgetbioprocess:
So What shall we do about this?

chr1sgrant:
I guess this can be deleted now

PR #113, opened 2016-08-25
removed blowout from dna and default

x-anderson:
i modified the dna lhpolicy to incorporate a postmix and remove the blowout. blowout was removed from defaults also. this was to ensure efficient mixing.


m-i-sadowski:
Superseded by other policy changes

PR #114, opened 2016-08-26
Handle intrinsic

ddn0:
Adds:
```go
  execute.Handle(ctx, condition string, *wtype.LHComponent) *wtype.LHComponent
```
to apply arbitrary conditions on components like:
```go
  dnaPart = Handle("Order DNA", dnaPart)
```
Handle instructions are inherently manual. The Human driver aggregates matching conditions in the generated `target.Inst` stream.


ddn0:
Required changing implementation of mixer.Sample and co in 6e6efea to change the ID for the sampled component.

Let me know if this change is okay. There needs to be a corresponding change in `antha-lang/antha/execute` to track the generation of components and pass the unit test.

PR #125, opened 2016-09-05
generate unique mixer file names

ddn0:


gogogadgetbioprocess:
Generates expected behaviour when sending protocols to Trilution through client device

PR #134, opened 2016-09-13
Handle selector

ddn0:


target/auto/execute.go
```diff
@@ -0,0 +1,83 @@
...
+		Type: inst.Files.Type,
+		Data: inst.Files.Tarball,
+	})
+	if err != nil {
+		return err
+	}
+
+	// Proof of concept
+	var errors []string
+	for range time.Tick(5 * time.Second) {
```
m-i-sadowski: [nitpicking] possible memory leak here? https://golang.org/pkg/time/#Tick

PR #162, opened 2016-11-07
Fix excess volumes

m-i-sadowski:
set all pcrplate well volumes to 200ul... fixes Synthace/elements#6 @josi-asae


m-i-sadowski:
@synthacebot test this please

PR #174, opened 2016-11-11
Bracemerge

gogogadgetbioprocess:
Planning to move all elements to antha/elements following this


gogogadgetbioprocess:
This is ready to merge apart from the new conflicts in the path pointing to the antha elements in each generated go file in component/lib. I've fixed this on the command line and will push that change now

ddn0:
@synthacebot test this please

PR #176, opened 2016-11-15
Fix mixnamed

m-i-sadowski:
Reinstated plate names


microArch/scheduler/liquidhandling/layoutagent2.go
```diff
@@ -378,9 +361,34 @@ func modpc(choice PlateChoice, nwell int) []PlateChoice {
 			fmt.Println("S: ", s, " E: ", e)
 			fmt.Println("L: ", len(choice.Assigned), " ", choice.Assigned)
 			fmt.Println("W: ", len(choice.Wells), " ", choice.Wells)
+			fmt.Println("I: ", ID)
 		*/
-		tx := strings.Split(choice.Name, "_")
-		nm := tx[0] + "_" + tx[1] + "_" + ID[0:6]
+		nm := choice.Name
```
ddn0: if choice.Name can't be postfixed with instances beforehand, how about:

```go
instances := make(map[string]int)

...

instance, seen := instances[choice.Name]
if seen {
  name = fmt.Sprintf("%s%d, choice.Name, instance)
} else {
  name = choice.Name
}
instances[choice.Name] += 1
```

If they can be postfixed already, at least move this to a function:
```go
seen := make(map[string]bool)
makeUniqueName := func(name string) (string, error) {
  if !seen[name] {
    seen[name] = true
    return name, nil
  }
  for k := 2; k < maxTries; k++ {
    candidate := fmt.Sprintf("%s%d", name, k)
    if !seen[candidate] {
      seen[candidate] = true
      return candidate, nil
     }
  }
  return "", fmt.Errorf("failed to name output plate after %d tries", maxTries)
}
```

m-i-sadowski:
@ddn0 test this please

microArch/scheduler/liquidhandling/layoutagent2.go
```diff
@@ -374,18 +357,40 @@ func modpc(choice PlateChoice, nwell int) []PlateChoice {
...
-		nm := tx[0] + "_" + tx[1] + "_" + ID[0:6]
+
+		nm := unique_plate_name(choice.Name, seen, 100)
+
 		r = append(r, PlateChoice{choice.Platetype, choice.Assigned[s:e], ID, choice.Wells[s:e], nm})
 	}
 	return r
 }
 
+func unique_plate_name(namein string, seen map[string]bool, maxtries int) string {
```
ddn0: camel case please.

PR #179, opened 2016-11-16
Generate paths relative to gopath

ddn0:
Required for matching elements with their definitions when elements are merged from various sources.

### DIFF
--- a/antha/compile/antha.go
+++ b/antha/compile/antha.go
@@ -30,6 +30,8 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"os"
+	"os/user"
 	"path/filepath"
 	"reflect"
 	"sort"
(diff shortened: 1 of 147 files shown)


antha/compile/antha.go
```diff
@@ -292,11 +294,47 @@ func sortKeys(m map[string]param) []string {
...
+// Return path relative to gopath if possible
+func relativeToGoPath(gopath string, name string) (string, error) {
+	prefixes := filepath.SplitList(gopath)
+	if len(prefixes) == 0 {
+		usr, err := user.Current()
+		if err != nil {
+			return "", err
+		}
+
+		prefixes = append(prefixes, filepath.Join(usr.HomeDir, "go"))
```
m-i-sadowski: Can we make this $GOPATH rather than assuming $GOPATH = $HOME/go ?

PR #184, opened 2016-11-21
support workflow+params as bundle

ddn0:
```sh
antharun --parameters params.json --workflow workflow.json
# now same as
jq -s '.[0] * .[1]' params.json workflow.json > bundle.json
antharun --bundle bundle.json
```

Along the way, get rid of unused and undocumented auto-expansion of workflows to parameter (e.g., n parameters for 1 element workflow expanded to n 1 parameter 1 element workflows)

Mutual dependence:
  - antha-lang/elements#3


m-i-sadowski:
antha/elements needs fixing to remain compatible

Issue #183, opened 2016-11-17
HVOffset not working ? 


gogogadgetbioprocess:
git checkout b62d149f0d67bc42c2bf0c7db9a5e8bf959312c6

see report for execution instructions: report20161117165242.md

 and csv file with description of plate test results: Arnoldpcrplateskirted20161117160504.csv

PR #185, opened 2016-11-21
Investigate hv offset

m-i-sadowski:
Modified the priority of the HV offset policy, made contingent on tip and platform, consequence is now an-offset-to-an-offset


m-i-sadowski:
fixes #183

PR #228, opened 2016-12-06
Reaches for eliminate

ddn0:
Reduce memory usage by calculating simpler node reachability versus eliminating nodes.


ddn0:
@synthacebot test this please

PR #230, opened 2016-12-07
Newelementtemplate

gogogadgetbioprocess:

PR #237, opened 2016-12-09
new element command

ddn0:
Replacement for:
  - antha-lang/antha#230


gogogadgetbioprocess:
Should the example parameter Volumes be type []Volume rather than []string ??

PR #250, opened 2016-12-15
Autoallocate mixes in place

m-i-sadowski:
This PR extends autoallocation to anonymous components coming in outside plates.

This may create confusion in some cases but hopefully will be clearer on aggregate.

Limitation: if the component is too big to fit into a single autoallocated well the system will refuse, throw an error, and go sulk in the corner

this may be confusing in some cases since it will work or not depending on what other autoallocated components are in the mix. Hence YMMV.

Please test and see how it goes.

### DIFF
--- a/microArch/scheduler/liquidhandling/autoallocate_test.go
+++ b/microArch/scheduler/liquidhandling/autoallocate_test.go
@@ -1,0 +1,114 @@
+package liquidhandling
+
+import (
+	"fmt"
+	"github.com/antha-lang/antha/antha/anthalib/mixer"
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"github.com/antha-lang/antha/antha/anthalib/wunit"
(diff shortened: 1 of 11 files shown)


microArch/scheduler/liquidhandling/autoallocate_test.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func testSetup(rbt *liquidhandling.LHProperties, expected map[string]float64, t *testing.T) {
+	for _, p := range rbt.Plates {
+		for _, w := range p.Wellcoords {
+			if !w.Empty() {
+				v, ok := expected[w.WContents.CName]
+
+				if !ok {
+					t.Fatal(fmt.Sprint("ERROR: unexpected component in plating area: ", w.WContents.CName))
```
ddn0: At least, `t.Fatalf`. `t.Errorf` is preferred so as to capture as many errors as possible in a test.

This generally applies throughout this file.

PR #253, opened 2016-12-19
return error nil at end of function

gogogadgetbioprocess:


gogogadgetbioprocess:
@synthacebot test this please

PR #256, opened 2016-12-22
increase default mix volume

chr1sgrant:


m-i-sadowski:
I have added changes which should reinstate override behaviour. For this to work mix volumes cannot be set in the default policy.

chr1sgrant:
This should solve a user issue

PR #257, opened 2017-01-05
DNASequence methods added for LHComponents

chr1sgrant:


antha/anthalib/wtype/lhcomponent.go
```diff
@@ -93,6 +93,244 @@ func (lhc *LHComponent) IsZero() bool {
 	return false
 }
 
+const SEQSKEY = "DNASequences"
+
+// Return a sequence list from a component.
+// Users should use GetDNASequences method.
+func (lhc *LHComponent) getSequences() (seqs []DNASequence, err error) {
```
m-i-sadowski: getDNASequences() if this is going to be so specific.

antha/anthalib/wtype/lhcomponent.go
```diff
@@ -93,6 +93,244 @@ func (lhc *LHComponent) IsZero() bool {
...
+		err = fmt.Errorf("Problem getting %s sequences. Sequences found: %+v; error: %s", lhc.Name(), seqsValue, err.Error())
+	}
+
+	return
+}
+
+// Add a sequence list to a component.
+// Any existing component list will be overwritten.
+// Users should use addDNASequence and UpdateDNASequence methods
+func (lhc *LHComponent) setSequences(seqList []DNASequence) error {
```
m-i-sadowski: please use *DNASequence everywhere for compatibility with BioSequence

antha/anthalib/wtype/lhcomponent.go
```diff
@@ -93,6 +93,244 @@ func (lhc *LHComponent) IsZero() bool {
...
+
+	lhc.Extra[SEQSKEY] = seqList
+
+	return nil
+}
+
+// Returns the positions of any matching instances of a sequence in a slice of sequences.
```
m-i-sadowski: Why the positions and not just the sequences themselves?

antha/anthalib/wtype/lhcomponent.go
```diff
@@ -93,6 +93,244 @@ func (lhc *LHComponent) IsZero() bool {
...
+// If checkSeqs is set to false, only the name will be checked;
+// if checkSeqs is set to true, matching sequences with different names will also be checked.
+func containsSeq(seqs []DNASequence, seq DNASequence, checkSeqs bool) (bool, []int) {
```
m-i-sadowski: This function would be better expressed as indexesOfSeq(seqs ... ), containsSeq really implies just yes or no, not where. However as noted above I'm not convinced we need to know where they are stored in the array so I would really suggest turning this into a getter method on wtype.DNASeqSet() (see below)

Would much prefer a refactor to add a method to the relevant type in wtype i.e. wtype.DNASeqSet.ContainsSeq(sq *DNASequence)

antha/anthalib/wtype/lhcomponent.go
```diff
@@ -93,6 +93,244 @@ func (lhc *LHComponent) IsZero() bool {
...
+			err = lhc.setSequences(seqList)
+			return err
+		}
+	}
+
+	return fmt.Errorf("Sequence %s did not previously exist in %s so could not be deleted.", seq.Name(), lhc.Name())
+}
+
+// Remove a DNA sequence from a specific position.
+// Designed for cases where FindDNASequnce() method returns multiple instances of the dna sequence.
```
m-i-sadowski: Why do we need to support this? I think it is much more sensible to restrict to names being locally unique

antha/anthalib/wtype/lhcomponent.go
chr1sgrant: This is only a private a function, the public method returns the sequences

antha/anthalib/wtype/lhcomponent.go
chr1sgrant: I personally disagree with this "This function would be better expressed as indexesOfSeq(seqs ... ), containsSeq really implies just yes or no, not where" 

The function is private anyway, and returns both true or false and position so indexes would also not quite describe the function accurately.

antha/anthalib/wtype/lhcomponent.go
chr1sgrant: I do agree but this would actually be a very big change which should be done separately since currently most functions take DNASequence not *DNASequence

PR #279, opened 2017-02-07
Feature/update xlsx

chr1sgrant:


chr1sgrant:
replaced by #280 and #281

PR #304, opened 2017-02-24
Plateheightsupdate

x-anderson:
*updated plate heights and physically tested


microArch/factory/make_plate_library.go
```diff
@@ -547,14 +547,14 @@ func makePlateLibrary() map[string]*wtype.LHPlate {
 	bottomtype = wtype.LHWBFLAT
 	xdim = 11.0
 	ydim = 11.0
-	zdim = 20.0
-	bottomh = 3.0
+	zdim = 19.0
+	bottomh = 2.0
```
m-i-sadowski: this is essentially not applicable here but we should probably revert it to its old values to prevent confusion

PR #338, opened 2017-04-10
New plate/idt plate

x-anderson:
Ran plate tests on the new IDT plate:

Residual volume was set at 200ul, which allows efficient aspirating/dispensing after testing. Tips unable to go to bottom of well due to tip head collision, hence high residual volume set. Likely with multichannel use, when implemented, tip head can go much lower

-High volume aspirating and dispensing tested OK

-Low volume aspirating and dispensing tested OK


chr1sgrant:
hmm tricky with such a high residual volume since this is a plate for DNA which will likely be far lower than 200ul

x-anderson:
Yes, unfortunately due to the physical limitations of the pipetmax there is not much we can do (similar to the deep 96 well plates) - the only way it will reach the bottom is with multichannel.

I could do a more physical testing to get a more specific residual volume?

PR #339, opened 2017-04-10
New plate/eppi lid clip

x-anderson:
Tested with plate height test:

-Low volume aspirating/dispensing OK
-High volume aspirating/dispensing OK


chr1sgrant:
Does it pipette ok to all columns and wells?

x-anderson:
Yes, i tested dispensation to every well and visually confirmed this.

PR #352, opened 2017-04-24
refactor pubchem library

chr1sgrant:


antha/AnthaStandardLibrary/Packages/Pubchem/Pubchem.go
```diff
@@ -270,23 +242,76 @@ type Properties struct {
...
+		return molecule, errors.New(fmt.Sprint("No property table for ", name, " got this from pubchem: ", pubchemtable))
+	}
+	molecule.CID = pubchemtable.Propertytable[0].CID
+	molecule.MolecularFormula = pubchemtable.Propertytable[0].MolecularFormula
+	molecule.MolecularWeight = pubchemtable.Propertytable[0].MolecularWeight
+	return molecule, nil
+}
+
+// Make an array of molecules based on molecule names
 func MakeMolecules(names []string) ([]Molecule, error) {
```
m-i-sadowski: I feel like with a convenience method like this it's more helpful to return as many valid entries as possible so I suggest (suggest only - not urgent to do it now) refactoring to return an array of errors rather than a single one OR concatenating errors as you go, either way I would prefer not to exit on first error

antha/AnthaStandardLibrary/Packages/Pubchem/Pubchem.go
```diff
@@ -270,23 +242,76 @@ type Properties struct {
 	CID              int     `json:"CID"`
 }
 
+// The principle type returned from querying the pubchem database if the molecule is not defined as a substance.
 type Molecule struct {
 	Moleculename     string
```
m-i-sadowski: MoleculeName?

chr1sgrant: There's a dependency in antha-lang/elements though so I'll do that tonight in sync with the elements PR. Actually maybe molecule.Name would be simpler.

antha/AnthaStandardLibrary/Packages/Pubchem/Pubchem.go
chr1sgrant: Ok good point, I'll do that.

PR #353, opened 2017-04-25
added additional MoClo custom overhangs

x-anderson:
also updated annotations


x-anderson:
Just to clarify - it is necessary for these parts and overhangs to be put in to allow the BSAI clonal parts to be designed by Antha

PR #371, opened 2017-05-10
Bug/volume issues

m-i-sadowski:
A few improvements: 
- better reporting when lower volume limit is breached
- rounded comparison operators for ConcreteMeasurement objects to sidestep numerical issues
- use of rounded comparison operators when choosing channels to avoid protocol failures


chr1sgrant:
closes #264

PR #374, opened 2017-05-12
Update/new egel48plate

x-anderson:


microArch/driver/liquidhandling/makelhpolicy.go
```diff
@@ -822,6 +822,12 @@ func MakeLVDNAMixPolicy() wtype.LHPolicy {
 	return dnapolicy
 }
 
+func MakeEPAGE48Policy() wtype.LHPolicy {
```
chr1sgrant: MakeEPAGE48Policy, MakeEGEL48Policy, MakeEGEL961Policy and MakeEGEL962Policy are all the same rule so  it might be better and less error prone to make one rule...

```go
func TurnOffBlowOut()  wtype.LHPolicy{
loadpolicy := make(wtype.LHPolicy, 1)
 +	loadpolicy["RESET_OVERRIDE"] = true
 +	return loadpolicy
}
```

x-anderson:
@chr1sgrant is it possible to refine the rules section into one (without it only applying the rule if all the conditions are met?)

e.g. combining

`//fix for removing blowout in DNA only if EGEL 48 plate type is used
	rule = wtype.NewLHPolicyRule("EPAGE48Load")
	rule.AddCategoryConditionOn("TOPLATETYPE", "EPAGE48")
	pol = TurnOffBlowoutPolicy()
	lhpr.AddRule(rule, pol)

	//fix for removing blowout in DNA only if EGEL 48 plate type is used
	rule = wtype.NewLHPolicyRule("EGEL48Load")
	rule.AddCategoryConditionOn("TOPLATETYPE", "EGEL48")
	pol = TurnOffBlowoutPolicy()
	lhpr.AddRule(rule, pol)`

chr1sgrant:
I don't think that's possible at present

PR #391, opened 2017-05-22
Fix/1.5mltubesracks

x-anderson:
Testing the tube racks out in master and on ui resulted in colliisons (wrong y-offset) - have adjusted these to fix.

### DIFF
--- a/microArch/factory/make_plate_library.go
+++ b/microArch/factory/make_plate_library.go
@@ -350,7 +350,7 @@ func makePlateLibrary() map[string]*wtype.LHPlate {
 
 	welltypesmallereppy := wtype.NewLHWell("1.5mlEpp", "", "", "ul", 1500, 50, eppy, wtype.LHWBV, 8.2, 8.2, 45, 4.7, "mm")
 
-	plate = wtype.NewLHPlate("eppendorfrack425_1.5ml", "Unknown", 4, 6, 45, "mm", welltypesmallereppy, wellxoffset, wellyoffset, xstart, ystart, zstart)
+	plate = wtype.NewLHPlate("eppendorfrack425_2ml", "Unknown", 4, 6, 45, "mm", welltypesmallereppy, wellxoffset, wellyoffset, xstart, ystart, zstart)
 	plates[plate.Type] = plate
 
 	// Eppendorfrack 424 with lid holders and using 2ml tubes
--- a/microArch/factory/make_plate_library.go
+++ b/microArch/factory/make_plate_library.go
@@ -332,7 +332,7 @@ func makePlateLibrary() map[string]*wtype.LHPlate {
 	wellxoffset = 18.0 // centre of well to centre of neighbouring well in x direction
 	wellyoffset = 18.0 //centre of well to centre of neighbouring well in y direction
 	xstart = 5.0       // distance from top left side of plate to first well
-	ystart = 5.0       // distance from top left side of plate to first well
+	ystart = 10.0      // distance from top left side of plate to first well
 	zstart = 6.0       // offset of bottom of deck to bottom of well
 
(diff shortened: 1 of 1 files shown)


microArch/factory/make_plate_library.go
```diff
@@ -332,34 +332,46 @@ func makePlateLibrary() map[string]*wtype.LHPlate {
...
+	ystart = 10.0      // distance from top left side of plate to first well
 	zstart = 6.0       // offset of bottom of deck to bottom of well
 
 	welltype2mleppy := wtype.NewLHWell("2mlEpp", "", "", "ul", 2000, 25, eppy, wtype.LHWBV, 8.2, 8.2, 45, 4.7, "mm")
 
 	plate = wtype.NewLHPlate("Kombi2mlEpp", "Unknown", 4, 2, 45, "mm", welltype2mleppy, wellxoffset, wellyoffset, xstart, ystart, zstart)
 	plates[plate.Type] = plate
 
-	// Eppendorfrack
+	// Eppendorfrack 425 for 2ml tubes
```
chr1sgrant: the description says // Eppendorfrack 425 for 2ml tubes 

but the definition is for 1.5ml tubes

microArch/factory/make_plate_library.go
```diff
@@ -332,34 +332,46 @@ func makePlateLibrary() map[string]*wtype.LHPlate {
...
 
 	wellxoffset = 18.0 // centre of well to centre of neighbouring well in x direction
 	wellyoffset = 18.0 //centre of well to centre of neighbouring well in y direction
-	xstart = 5.0       // distance from top left si z9fdJwde of plate to first well
-	ystart = 5.0       // distance from top left side of plate to first well
-	zstart = 7.0 - 1.0 // offset of bottom of deck to bottom of well
+	xstart = 4.5       // distance from top left side of plate to first well
+	ystart = 9         // distance from top left side of plate to first well
+	zstart = 5.0       // offset of bottom of deck to bottom of well
```
chr1sgrant: did you try these with both heads? 

We should pursue the head specific offset rule

microArch/factory/make_plate_library.go
chr1sgrant: It would be worth having definitions for both

x-anderson: Have added both (although i stuck with the 2ml as having the 1.5mls in there with the lids, its not really possibly to have many tubes)

microArch/factory/make_plate_library.go
x-anderson: Yes, I verified these with both heads.   i can look into the head specific offset rule.

microArch/factory/make_plate_library.go
chr1sgrant: The annotation on line 343 is for 2ml tubes but the plate specified from lines 345 - 354 depends on using 1.5ml eppendorfs. I think there should be both 1.5ml and 2ml dimensions specified for this rack as well as rack 424

x-anderson: Im not sure i understand - line 343 is the annotation for 2ml and line 351 specifies the 2ml plate - similarly lines 354 and 364 for the 1.5ml?

I didnt observe any differences needed in dimensions for using 1.5ml and 2ml tubes so the only differences between these are the volumes.

chr1sgrant: OK sorry, I see that now looking at the whole file, from the changes it was confusing. So the residual volume for the 2ml tube is 25ul and 50ul for the 1.5ml tube?

x-anderson: havent verified the residual volumes (they were already inputted there so I assumed they had been checked previously?). If not I can do this. However Id really like to get this in asap so I can run my workflow ideally today, so if its oK to merge now and I can update the residuals tomorrow/when i get back.

x-anderson: Also to clarify -  these 2ml tubes are the ones with screw caps - I can update the definition to reflect that and push up

PR #392, opened 2017-05-22
Fix/496 rack

x-anderson:


microArch/factory/make_plate_library.go
```diff
@@ -316,7 +316,12 @@ func makePlateLibrary() map[string]*wtype.LHPlate {
 
 	// pcr plate with 496rack
 
-	plate = wtype.NewLHPlate("pcrplate_with_496rack", "Unknown", 8, 12, 15.5, "mm", pcrplatewell, 9, 9, 0.0, 0.0, pcrtuberack496-2.5)
+	plate = wtype.NewLHPlate("pcrplate_with_496rack", "Unknown", 8, 12, 15.5, "mm", pcrplatewell, 9, 9, 0.0, 0.0, pcrtuberack496)
```
chr1sgrant: So these have different z offsets? 

Looks good, could add the manufacturer name instead of unknown.

x-anderson: Yes, the pcr_skirted place had a different offset. However, i wasnt able to verify with a non-skirted plate, which may have same offset as the 0.2ml tubes. Possibly we need to update the plate names (pcrplate_semi_skirted pcrplate_non_skirted etc instead of just pcrplate, as we encoutner more of these issues)

PR #394, opened 2017-05-24
renaming PCR to par

n-scholes:


antha/AnthaStandardLibrary/Packages/PCR/pcr.go
```diff
@@ -1,7 +1,7 @@
 // This is a package that contains types and function used for PCR reactions
```
chr1sgrant: It looks like the package folder has not been updated. you might have to git rm and git add again

antha/AnthaStandardLibrary/Packages/PCR/pcr.go
```diff
@@ -1,7 +1,7 @@
...
-package PCR
+package pcr
 
-type PCRReaction struct {
+type Reaction struct {
```
chr1sgrant: Can we make template and PrimerPairs DNASequence instead of string?

chr1sgrant: then two methods 
```go
func (r Reaction) TemplateName() string{
}
func (r Reaction) PrimerNames() [2]string{
}
```

chr1sgrant:
@synthacebot test this please

PR #412, opened 2017-06-22
Fix/src tgt

ddn0:
- antha-lang/elements#171
  - Synthace/elements#174
  - Synthace/monolith#276


workflow/workflow.go
```diff
@@ -38,8 +38,8 @@ type Process struct {
 }
```
chr1sgrant: I think more changes are needed to be compatible with the new structure:
```go
type Process struct {
	Component string `json:"component"`
}
```
to 
```go
type Process struct {
	Component string `json:"elementName"`
}
```

```go 
type Port struct {
	Process string `json:"process"`
	Port    string `json:"port"`
}
```

to 

```go 
type Port struct {
	Process string `json:"instanceName"`
	Port    string `json:"output"`
} 
```
and 
```go
type Port struct {
	Process string `json:"instanceName"`
	Port    string `json:"input"`
}
```
?

```json
 "Processes": {
        "AutoAssembly_multi_881860": {
            "elementName": "AutoAssembly_multi",
            "metadata": {
                "x": 503,
                "y": 136,
                "isSelected": false
            }
        },
```

```json
"source": {
                "instanceName": "ParseDesign_406172",
                "output": "ReactionMap"
            },
            "target": {
                "instanceName": "AutoAssembly_multi_881860",
                "input": "Reactiontopartseqs"
            }
```

PR #415, opened 2017-06-26
delete duplicate platereader data files from Parser package

chr1sgrant:


chr1sgrant:
@synthacebot test this please

PR #422, opened 2017-07-03
Fix/entrez lookup

chr1sgrant:


chr1sgrant:
fixes antha-lang/elements CI tests for https://github.com/antha-lang/elements/pull/150

PR #436, opened 2017-07-25
Feature/workflow test bundles

m-i-sadowski:
Includes robot instruction deserialization


execute/executeutil/unmarshal.go
```diff
@@ -23,43 +25,46 @@ type UnmarshalOpt struct {
...
-		execute.RawParams
-	}
+type Bundle struct {
+	workflow.Desc
+	execute.RawParams
+	workflowtest.TestOpt
+}
 
+// Parse parameters and workflow.
+func Unmarshal(opt UnmarshalOpt) (*workflow.Desc, *execute.RawParams, *workflowtest.TestOpt, error) {
```
ddn0: 4 return values is pretty unwieldy. Since you are already changing this, can you just return `Bundle`, i.e., `func Unmarshal(UnmarshalOpt) (*Bundle, error)`

PR #454, opened 2017-08-11
anthaFmt minor improvement

cachemoi:
Giving more comprehensive error message to anthafmt.

It now returns the path to the file in which errors occurs along with the error message.


ddn0:
add to whitelist

PR #485, opened 2017-09-26
Parse conc from assembly design

chr1sgrant:


jarpinosynthace:
We will now need to update any template XLSX files for assembly design to contain a concentration column!

PR #500, opened 2017-10-18
added SmartMix policy and adapted PreMix behaviour to look at source …

chr1sgrant:
…solution volume


microArch/driver/liquidhandling/makelhpolicy.go
```diff
@@ -874,6 +907,93 @@ func GetLHPolicyForTest() (*wtype.LHPolicyRuleSet, error) {
...
+	adjustPostMix = wtype.NewLHPolicyRule("mixInto100ul")
+	adjustPostMix.AddCategoryConditionOn("LIQUIDCLASS", "SmartMix")
+	adjustPostMix.AddNumericConditionOn("TOWELLVOLUME", 100.0, 200.0)
+	adjustVol100 := AdjustPostMixVolume(wunit.NewVolume(100, "ul"))
+
+	lhpr.AddRule(adjustPostMix, adjustVol100)
+
+	adjustPostMix = wtype.NewLHPolicyRule("mixInto200ul")
+	adjustPostMix.AddCategoryConditionOn("LIQUIDCLASS", "SmartMix")
+	adjustPostMix.AddNumericConditionOn("TOWELLVOLUME", 200.0, 500.0)
```
x-anderson: What would happen in this case in the case of well volumes >500ul?

microArch/driver/liquidhandling/makelhpolicy.go
```diff
@@ -874,6 +907,93 @@ func GetLHPolicyForTest() (*wtype.LHPolicyRuleSet, error) {
...
+
+	lhpr.AddRule(adjustPreMix, adjustPreMixVol100)
+
+	adjustPreMix = wtype.NewLHPolicyRule("premixFrom200ul")
+	adjustPreMix.AddCategoryConditionOn("LIQUIDCLASS", "PreMix")
+	adjustPreMix.AddNumericConditionOn("WELLFROMVOLUME", 200.0, 1000.0)
+
+	lhpr.AddRule(adjustPreMix, adjustPreMixVol200)
+
+	// NeedToMix
```
x-anderson: We need to deprecate the needtomix policy, and replace it with this proposed smartmix once tested

microArch/driver/liquidhandling/makelhpolicy.go
chr1sgrant: good point we can increase this value

microArch/driver/liquidhandling/makelhpolicy.go
chr1sgrant: Let's see after testing

chr1sgrant:
Accuracy test looks ok

PR #514, opened 2017-10-31
convert from file formats

ddn0:


chr1sgrant:
Just tried converting this bundle and it still got this error message when attempting to upload output.json into the UI:
`Import: could not parse the file` 

https://repos.antha.com/antha-ninja/elements-westeros/blob/master/antha-lang-elements/multi-element-workflows/MakePalette_PipetteImage_fromPalette/bundle.json

PR #539, opened 2017-11-22
Fix/get fullyqualifiednames

x-anderson:


m-i-sadowski:
Revised version to be submitted

PR #541, opened 2017-11-23
Feature/handle more unit factors

chr1sgrant:


chr1sgrant:
Have to merge this in to allow user testing in the UI: let's review later if that's ok

PR #561, opened 2017-12-08
use plate as argument in NextFreeWell

chr1sgrant:


chr1sgrant:
Breaks pipeline

jarpinosynthace:
do you need me to do anything here, Do we now need to update all elements that use the next free well function?

jarpinosynthace:
Potentially it is only my elements that have that function in them, unless you started using it.

PR #570, opened 2017-12-15
Normalise methods

chr1sgrant:


antha/anthalib/wtype/biology.go
```diff
@@ -134,7 +134,7 @@ func (seq DNASequence) Dup() DNASequence {
 
 func MakeDNASequence(name string, seqstring string, properties []string) (seq DNASequence, err error) {
 	seq.Nm = name
-	seq.Seq = seqstring
+	seq.Seq = upper(seqstring)
```
m-i-sadowski: It's common for people to use case to distinguish between different base states or when making alignments... it would make more sense to normalize when you do comparisons between sequences or other checks for which case is not meaningful, otherwise you lose information here

antha/anthalib/wtype/biology.go
```diff
@@ -162,19 +162,19 @@ func MakeDNASequence(name string, seqstring string, properties []string) (seq DN
 }
 func MakeLinearDNASequence(name string, seqstring string) (seq DNASequence) {
 	seq.Nm = name
-	seq.Seq = strings.ToUpper(seqstring)
+	seq.Seq = upper(seqstring)
```
m-i-sadowski: See above

chr1sgrant:
Upper case conversion of sequences has now been reverted.

PR #573, opened 2017-12-19
refactor RestrictionEnzyme type

chr1sgrant:
contains #571 and #570


chr1sgrant:
fixed by https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/223

PR #577, opened 2018-01-05
Job id type

chr1sgrant:


antha/compile/antha.go
```diff
@@ -236,6 +236,7 @@ func NewAntha(root *AnthaRoot) *Antha {
 		"FlowRate":             "wunit.FlowRate",
 		"Force":                "wunit.Force",
 		"HandleOpt":            "execute.HandleOpt",
+		"JobID":                "jobfile.JobID",
```
ddn0: need to add a default import too:

```
	p.addImportReq(&importReq{
		Path:    "github.com/antha-lang/antha/antha/anthalib/wtype",
		UseExpr: "wtype.FALSE",
	})
```

PR #616, opened 2018-01-31
New plate/nunc96 dw

x-anderson:
See attached document for testing details.

Overall functions fine.

Very similar to DSW96, so maintained the residual volume of those. Unilikely to be much different and is already very high. Need to look into how to get different heights for these plates at HV, as could go lower.

https://www.dropbox.com/s/9zpaxdxwv3i39up/Nunc96DeepWell.docx?dl=0

Passed on to Aleks for generation of UI data for plate


x-anderson:
@chr1sgrant 
Sure thing, next plates we add I will make sure to do this

PR #618, opened 2018-02-02
update zoffset of SRWFB plate, no physical testing

x-anderson:
@chr1sgrant Change here - no chance to physically test yet, but without something like this, the plate will 100% crash.


x-anderson:
updated with physical testing:

https://www.dropbox.com/s/qq4hcmvp8fa5aqb/SRWFB_updated%20height.docx?dl=0

chr1sgrant:
Would prefer to test with low volume tips too ideally

x-anderson:
@chr1sgrant Hi Chris, yes of course, I can do this this morning.

x-anderson:
Updated testing document to confirm no issues when tested with LV, or HV/LV on plates with no risers.

PR #623, opened 2018-02-05
addition of qPCR specific LH components to component library

CharlotteWard:


chr1sgrant:
is this still needed?

PR #628, opened 2018-02-08
Fixed counting error in CentreToQuery fn in bioinformatics.go

cosmonautspiff:
Was counting indicies wrong by 1 when writing the end portion of the query seq after the alignment portion. To see the effects of this, I've attached an the output file (originally the json file) from the Blastp element in the original prot-eng branch. These are the Query Centred Alignments. You can see that for subject sequences ending with gaps.... e.g., --", that the corresponding query sequences are incorrect at the end. 

[outputs_BLASTp_QCA_1000.txt](https://github.com/antha-lang/antha/files/1707138/outputs_BLASTp_QCA_1000.txt)

### DIFF
--- a/antha/anthalib/wtype/bioinformatics.go
+++ b/antha/anthalib/wtype/bioinformatics.go
@@ -196,14 +196,24 @@ func (aln SimpleAlignment) Column(i int) string {
 	return r
 }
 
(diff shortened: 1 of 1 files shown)


cosmonautspiff:
Changes here are included in the protein-engineering pull request

PR #629, opened 2018-02-10
Protein engineering

cosmonautspiff:
Added wtype.Mutation. A struct to hold data on mutations (for nucleotide and AAs). This branch follows directly from the fix-CentreToQuery branch, so you can just merge this one.

The fix-CentreToQuery branch was counting indicies wrong by 1 when writing the end portion of the query seq after the alignment portion. To see the effects of this, I've attached an the output file (originally the json file) from the Blastp element in the original prot-eng branch. These are the Query Centred Alignments. You can see that for subject sequences ending with gaps.... e.g., --", that the corresponding query sequences are incorrect at the end.

[outputs_BLASTp_QCA_1000.txt](https://github.com/antha-lang/antha/files/1717597/outputs_BLASTp_QCA_1000.txt)


antha/anthalib/wtype/bioinformatics.go
```diff
@@ -196,25 +196,37 @@ func (aln SimpleAlignment) Column(i int) string {
 	return r
 }
 
+// Trim aligned (subject) sequences to only those residues/nucleotides aligned
+// to those in the query. This removes inserts (wrt the query) from aligned
+// sequences. Gaps within the interior of the aligned sequence are already
+// represented by '-' characters, however gaps at the ends of the aligned
+// sequence must be added here as well. The resulting aligned sequences will 
+// have the same length as the query.
```
chr1sgrant: comments should start with the function name 

```
// CentreToQuery trims aligned sequences ... ```

File: antha/anthalib/wtype/mutation.go [hunk 2]
```diff
@@ -0,0 +1,7 @@
+package wtype
+
+type Mutation struct {
+  Residue int
+  Before  string
+  After   string
+}
```
chr1sgrant: We should think carefully about how we want to model this if it's going to be part of the standard library code. 

Is this what we want to do @m-i-sadowski ?

Do we want this in wtype? 

+ The type and each field should have comments.

antha/anthalib/wtype/bioinformatics.go
cosmonautspiff: Will fix.

antha/anthalib/wtype/mutation.go
cosmonautspiff: I can include this wtype.Mutation in the bioinformatics.go file. The reason I made the type, is to avoid passing two separate, but correspondingly ordered, arrays from one element to another (in the protein engineering workflow): one for residue number, and one for mutation AA-type. 

So I figured, might as well make a new type, include the "Before" field as well, and keep it agnostic to whether it's a nucleotide or amino acid mutation.

Open to other suggestions.

cosmonautspiff: I removed the Mutation type. Changes to CentreToQuery function still stand. Requesting merge for that.

chr1sgrant:
@m-i-sadowski are you happy to merge this in?

PR #632, opened 2018-02-15
Feature/simulator

haydnKing:
All the simulator stuff - should merge this in before drivers


antha/anthalib/wtype/lhinterfaces_test.go
```diff
@@ -0,0 +1,144 @@
...
+		"LHPlate":     &LHPlate{},
+		"LHTip":       &LHTip{},
+		"LHTipbox":    &LHTipbox{},
+		"LHTipwaste":  &LHTipwaste{},
+		"LHWell":      &LHWell{},
+	}
+	return ret
+}
+
+func getAssertionMap() map[string]func(interface{}) bool {
```
ddn0: for future reference, a slightly more robust way to do this is:
```
var _ Identifiable = (*LHComponent)(nil)
...
```

it's checked statically.

PR #635, opened 2018-02-19
Add check when mock devices are missing

m-i-sadowski:
Command-line usage segfaults without this fix

### DIFF
--- a/cmd/antha/cmd/run.go
+++ b/cmd/antha/cmd/run.go
@@ -198,15 +198,13 @@ func (a *runOpt) Run() error {
 	}
 
 	if targetConfig != nil {
-		if targetConfig.MockDevices != nil {
-			for _, mockDevice := range targetConfig.MockDevices {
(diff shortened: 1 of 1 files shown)


cmd/antha/cmd/run.go
```diff
@@ -196,13 +196,18 @@ func (a *runOpt) Run() error {
 			"cannot decode target-config file '%s' detail:%s",
 			a.TargetConfigFile, err)
 	}
-	for _, mockDevice := range targetConfig.MockDevices {
-		device, err := mockDevice.ToDevice()
-		if err != nil {
-			return fmt.Errorf("could not instatiate device from mock: %s", err)
+
+	if targetConfig != nil {
+		if targetConfig.MockDevices != nil {
```
ddn0: prefer removing this check as `for _, mockDevice := range nil { ... }` still works as expected.

jake-ccl:
Apologies, didn't check this.

PR #638, opened 2018-02-20
refactored absorbance data interface to simplify

chr1sgrant:


chr1sgrant:
This needs to come in for today's accuracy tests so merging in, already refactored Westeros branch of elements.

PR #639, opened 2018-02-21
Newplate/nunc96 u

chr1sgrant:


chr1sgrant:
[nunc96Utest_4_20180221175352.xlsx](https://github.com/antha-lang/antha/files/1748877/nunc96Utest_4_20180221175352.xlsx)

x-anderson:
@chr1sgrant Thanks for posting that!

A couple follow up questions:

1. On some of these runs, 'MultiChannel' column is indicated as 'true', but in rows 8-13, and 18-21, the user-inputted comments indicate 'no multi channel'. 
- Was this intended to do multichannel, and it did not? Or vice versa?

2. In rows 22 and 23, youve indicated that when multi-channeling, and it was too close to the bottom of the well, although it did dispense without error.  
- Was it closer than when single-channeling? 
- Was this an issue with aspirating liquids?

chr1sgrant:
1. The liquid was expected to dispense multichannel but did not. But that's not related to the plate, just an observation I didn't want to lose.
2. It was closer to the bottom than all other dispenses, but no collision with the bottom of the plate.

No aspiration was tested in this case.

x-anderson:
Sure - I expect you would have noticed aspiration issues when it did the SmartMi step.

PR #640, opened 2018-02-22
Plate library/cooler as incubator

chr1sgrant:
Refactored PCR plate holders/coolers to be treated as a riser which have a plateConstraints field to indicate that only certain plates can be added to them.

Added tests which ensure heights of previous plates in the library are maintained. This required the ability to specify SpecialOffsets for a plate/riser combination. This is undesirable but probably necessary since certain tubes will sit differently in a rack depending on the diameter of the tube, and in some cases there is a solid base which ensures all tubes sit at the same height, regardless of the tube diameter.

Untested in lab at present but any previous heights have CI tests added to check the heights are equivalent to before.

### DIFF
--- a/inventory/testinventory/devices.go
+++ b/inventory/testinventory/devices.go
@@ -2,6 +2,7 @@ package testinventory
 
 import (
 	"github.com/antha-lang/antha/antha/AnthaStandardLibrary/Packages/devices"
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
 )
 
 // heights in mm
@@ -14,7 +15,7 @@ const (
 	shallowriser18heightinmm                    = 18.75 - offset
 	coolerheight                                = 16.0
 	isofreezecoolerheight                       = 10.0
-	pcrtuberack496                              = 28.0
+	pcrtuberack496HeightInmm                    = 28.0
 	valueformaxheadtonotintoDSWplatewithp20tips = 4.5
 	bioshake96welladaptorheight                 = 4.5
 	bioshakestandardadaptorheight               = 5.0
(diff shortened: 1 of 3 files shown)


inventory/testinventory/devices.go
```diff
@@ -52,6 +53,33 @@ var defaultDevices = map[string]device{
...
+		Heightinmm:   pcrtuberack496HeightInmm,
+		Synonyms:     []string{"496rack"},
+		PlateConstraints: plateConstraints{
+			OnlyThesePlates: []plateWithConstraint{
+				plateWithConstraint{
+					Name:          "pcrplate",
+					SpecialOffset: -0.636,
+				},
+				plateWithConstraint{
+					Name:          "pcrplate_skirted",
```
x-anderson: remove this plate as not compatible with 496rack

inventory/testinventory/devices.go
```diff
@@ -89,6 +137,119 @@ var defaultDevices = map[string]device{
...
+				},
+			},
+		},
+		Properties: map[string]float64{
+			"Height": 0.0,
+		},
+		PositionConstraints: map[string][]string{},
+	},
+
+	"isofreeze_cooler": incubator{
```
x-anderson: we have to start depreciating anything using an isofreeze cooler because of functional issues - please dont include this in updated code.

inventory/testinventory/make_plate_library.go
```diff
@@ -440,23 +394,19 @@ func makeBasicPlates() (plates []*wtype.LHPlate) {
 	plate = wtype.NewLHPlate("Agarplateforpicking384", "Unknown", 16, 24, 14, "mm", welltype384, wellxoffset, wellyoffset, xstart, ystart, zstart)
 	plates = append(plates, plate)
 
-	// Onewell SBS format Agarplate with colonies on shallowriser (50ml agar) low res
+	// Onewell SBS format Agarplate with colonies on shallowriser (30ml agar) low res
```
x-anderson: is this volume change correct?

inventory/testinventory/make_plate_library.go
```diff
@@ -669,22 +612,66 @@ func makeBasicPlates() (plates []*wtype.LHPlate) {
...
+	pcrplatewell.SetAfVFunc(afs)
+
+	//LiquidLevel model for LL Following: vol_f estimates volume given height
+	vol_f := wutil.Quadratic{A: 0.402, B: 7.069, C: 0.0}
+	pcrplatewell.SetLiquidLevelModel(vol_f)
+
+	return pcrplatewell
+}
+
+func makePCRPlate() *wtype.LHPlate {
```
x-anderson: Looks good - very happy to see this moving over to individual functions

inventory/testinventory/devices.go
chr1sgrant: ok, will do.

inventory/testinventory/make_plate_library.go
chr1sgrant: Yes, there's also a 50ml version in the library. One day we'll model the agar separately I suspect but that might require modifying policies in elements.

PR #641, opened 2018-02-27
Fix/tip reuse with multi channeling try2

m-i-sadowski:
This PR contains a refactor of the liquid handling instruction hierarchy to 
change the TransferInstruction class to permit it to define multiple 
multichannel transfers where previously it could define at most one. 

This has the consequence that a single transfer can now generate a block
of multichannel instructions instead of only one. Since tip changes are always
generated at the start of a new block this was causing excessive tip usage. 
Consequently this refactor should allow for considerably more conservative
tip use.

### DIFF
--- a/microArch/driver/liquidhandling/compositerobotinstruction.go
+++ b/microArch/driver/liquidhandling/compositerobotinstruction.go
@@ -184,8 +184,6 @@ func (ins *SingleChannelBlockInstruction) Generate(ctx context.Context, policy *
 
 		tvs, err := TransferVolumes(ins.Volume[t], mergedchannel.Minvol, mergedchannel.Maxvol)
 
-		//fmt.Println("VOL ", ins.Volume[t].ToString(), " IN ", len(tvs), " GOES: ", mergedchannel.Minvol.ToString(), " MIN ", mergedchannel.Maxvol.ToString(), " MAX")
-
 		if err != nil {
 			return ret, err
 		}
@@ -210,18 +208,6 @@ func (ins *SingleChannelBlockInstruction) Generate(ctx context.Context, policy *
 			}
 
 			if change_tips {
-				/*
-					fmt.Println("CHANGING TIPS HERE ")
-					fmt.Println("THIS: ", this_thing.CName, " THAT: ", last_thing.CName)
-					fmt.Println("channels equal? ", channel == newchannel)
-					fmt.Println("tips same? ", tiptype == newtiptype)
-					fmt.Println("tip reuse over? ", n_tip_uses > pol["TIP_REUSE_LIMIT"].(int))
-					fmt.Println(n_tip_uses, " ", pol["TIP_REUSE_LIMIT"].(int), " NOT CAST: ", pol["TIP_REUSE_LIMIT"])
-					fmt.Println("dirty? ", dirty)
-				*/
-				// maybe wrap this as a ChangeTips function call
-				// these need parameters
-
 				tipdrp, err := DropTips(tt, prms, chanA)
 				if err != nil {
 					return ret, err
@@ -271,11 +257,6 @@ func (ins *SingleChannelBlockInstruction) Generate(ctx context.Context, policy *
 
 			if pol["DSPREFERENCE"].(int) == 0 && !ins.TVolume[t].IsZero() || premix && npre.(int) > 0 || postmix && npost.(int) > 0 {
 				dirty = true
-				/*
-					fmt.Println("DIRTY DIRTY DIRTY")
-					fmt.Println(pol["DSPREFERENCE"].(int) == 0, " ", ins.TVolume[t].ToString(), " ", "PRE: ", premix, " ", npre, " POST: ", postmix, " ", npost)
-					fmt.Println(ins.WellTo[t])
-				*/
 			}
 
 			ins.FVolume[t].Subtract(vol)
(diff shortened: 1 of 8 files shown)


m-i-sadowski:
Still some tests to do but functionally this looks good so far

microArch/driver/liquidhandling/transferblock.go
```diff
@@ -79,14 +79,35 @@ func (ti TransferBlockInstruction) Generate(ctx context.Context, policy *wtype.L
...
+		r := make([]RobotInstruction, len(inss))
+
+		for i, ins := range inss {
+			r[i] = RobotInstruction(ins)
+		}
+
+		return r
+	}
+
+	inss = fromTransfers(mergeTransfers(toTransfers(inss), policy))
```
haydnKing: Did we establish whether mergeTransfers is now working as expected? Previously it was failing to merge successive transfers (which allowed the planning numerical instability to end up causing the multichannel non-determinism)

microArch/driver/liquidhandling/transferblock.go
```diff
@@ -485,69 +505,51 @@ func (ti TransferBlockInstruction) GetParameter(p string) interface{} {
...
+	// merge only if the merge doesn't break either
 
-	// we just compare component names
+	ins3 := ins.Dup()
+	ins3.MergeWith(ins2)
 
-	m := a2Map(ins1.Components)
+	m1 := GetPolicyFor(policy, ins)["CAN_MULTI"].(bool)
+	m2 := GetPolicyFor(policy, ins2)["CAN_MULTI"].(bool)
+	m3 := GetPolicyFor(policy, ins3)["CAN_MULTI"].(bool)
```
haydnKing: There's some discussion about allowing elements to set LHPolicy items in the future, so we might want to think about error checking these, e.g. through implementing some accessors like `LHPolicy.getBool(name string) bool, err`

microArch/scheduler/liquidhandling/executionplanner3.go
```diff
@@ -52,8 +53,11 @@ func ExecutionPlanner3(ctx context.Context, request *LHRequest, robot *liquidhan
 
 			tfb := liquidhandling.NewTransferBlockInstruction(ch.Values)
 
+			fmt.Println("DEPTH ", depth, " LEN TFB: ", len(ch.Values))
```
haydnKing: Not sure if this is useful for users or leftover from debugging/development ;)

microArch/driver/liquidhandling/transferblock.go
m-i-sadowski: that would be safer, however the plan is to validate setting these items at source and only permit the correct datatype to be used. See aparam.go in wtype.

microArch/driver/liquidhandling/transferblock.go
m-i-sadowski: This is a total rewrite of how transfer merging happens. It should be considerably more robust. So far my tests have found it to perform better but it's worth trying against that same test case to confirm, I'll do that.

m-i-sadowski:
Tests pass locally.

microArch/driver/liquidhandling/transferblock.go
```diff
@@ -485,69 +505,51 @@ func (ti TransferBlockInstruction) GetParameter(p string) interface{} {
...
 
-	if len(m) > 1 {
-		return false
+	if !xor(m1, m2) {
```
ddn0:
```
if m1 == m2 {
  return m1 == m3
}
```

?

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -273,45 +228,47 @@ func plateTypeArray(ctx context.Context, types []string) ([]*wtype.LHPlate, erro
...
 
-	// the TransferBlock instruction takes into account the destinations being OK
-	// splits instructions into potentially multiable blocks on that basis
-	// and finds sources for them
+	return r
+}
 
-	// -- the transfer block ensures these instructions are at most multi long
-	// all of the below assumes we can't span multiple plates
+func (ins *TransferInstruction) testParallelSet(ctx context.Context, channel *wtype.LHChannelParameter, which int) bool {
```
ddn0: rename to `canParallelSet`? or `validateParallelSet`?

microArch/scheduler/liquidhandling/executionplanner3.go
```diff
@@ -54,6 +55,7 @@ func ExecutionPlanner3(ctx context.Context, request *LHRequest, robot *liquidhan
 
 			request.InstructionSet.Add(tfb)
 		}
+		depth += 1
```
ddn0: does anyone use this?

m-i-sadowski: Leftover debug... removed

microArch/driver/liquidhandling/transferblock.go
m-i-sadowski: Doh! Yes. That's simpler. Much much simpler. Face, meet palm.

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -387,17 +310,16 @@ func NewMultiChannelBlockInstruction() *MultiChannelBlockInstruction {
...
-	ins.PltFrom = append(ins.PltFrom, mct.PltFrom)
-	ins.PltTo = append(ins.PltTo, mct.PltTo)
-	ins.WellFrom = append(ins.WellFrom, mct.WellFrom)
-	ins.WellTo = append(ins.WellTo, mct.WellTo)
-	ins.Volume = append(ins.Volume, mct.Volume)
-	ins.FPlateType = append(ins.FPlateType, mct.FPlateType)
-	ins.TPlateType = append(ins.TPlateType, mct.TPlateType)
-	ins.FVolume = append(ins.FVolume, mct.FVolume)
-	ins.TVolume = append(ins.TVolume, mct.TVolume)
-	ins.Prms = mct.Channel
```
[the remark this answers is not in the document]
m-i-sadowski: Yes

microArch/driver/liquidhandling/transferinstruction.go
m-i-sadowski: done

PR #642, opened 2018-02-27
use latest version of go in ReadME

St3v3B:


README.md
```diff
@@ -25,7 +25,7 @@ for using both are stored here.
 
 ### OSX (Native)
 
-First step is to install or upgrade to go 1.7. Follow the instructions at the
+First step is to install or upgrade to go 1.10. Follow the instructions at the
```
chr1sgrant: Should we state the version of just say, latest or at least version 1.7??

St3v3B: I think Johan said the latest version when I showed hi the problem @johansglock ??

johansglock: yep, so that makes @chr1sgrant valid, less updating to do

PR #645, opened 2018-02-28
New plate/deep well96 1ml

x-anderson:
Testing files documented here:

https://www.dropbox.com/s/skklcqm0xanu1hb/Thermo_96_deepwell_1ml.docx?dl=0

https://www.dropbox.com/s/n8opkuur49mu4p6/Thermo_96_deepwell_1ml20180226143140.csv?dl=0

### DIFF
--- a/inventory/testinventory/make_plate_library.go
+++ b/inventory/testinventory/make_plate_library.go
@@ -757,7 +757,7 @@ func make96DeepWellLowVolumePlate() *wtype.LHPlate {
 	// These corrections are necessary to subtract from the official (correct) dimensions in order obtain correct pipetting behaviour.
 	xstartOffsetCorrection := 14.50
 	ystartOffsetCorrection := 11.50
-	zstartOffsetCorrection := 3.15
+	zstartOffsetCorrection := 2.5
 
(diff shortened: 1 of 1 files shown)


inventory/testinventory/make_plate_library.go
```diff
@@ -746,3 +749,51 @@ func makeGreinerVBottomPlateWithRiser() *wtype.LHPlate {
 	plate.WellZStart = 43.0
 	return plate
 }
+
+// Nunc™ 1.0 ml DeepWell™ Plates with Shared-Wall Technology Cat Num: 260251
```
chr1sgrant: Why does the comment say Nunc and the plate name is Thermo?

We should ideally aim to reassess the naming conventions and be consistent across the plate library

Perhaps standardising e.g: 

manufacturer_wells_wellshape_material_colour

nunc_96_mediumWellV_PP_Clear

x-anderson: Yes I agree about convention - I admit I havent been particularly srtict.

Thermo is the supplier, but Nunc is the brand,so it should be Nunc. Im happy to change this.

PR #648, opened 2018-02-28
New plate/fluidx700ul

chr1sgrant:


x-anderson:
@chr1sgrant  Was going to test this physically, but the linked plate and the description (8x12 wells) does not appear to relate to any FLuidX plates we have in stock - possible we need to obtain these plates, or these may be other plates, but im not totally sure which ones.

chr1sgrant:
there's a link to it in the comment: // http://fluidx.eu/0.7ml%2c-96-well-format-2d-barcoded-jacket-tube-with-external-thread.html
It's basically a rack with the small tubes.

x-anderson:
OK, we dont have any of these here so would need to order one to finalise the physical testing - shall I do this?

chr1sgrant:
[FluidXTest20180315165355.xlsx](https://github.com/antha-lang/antha/files/1819297/FluidXTest20180315165355.xlsx)

PR #649, opened 2018-03-02
Fix/washes omega

m-i-sadowski:
This Pull Request introduces a new intrinsic: SplitSample, which is a two-return version of the previous Sample method 

This was the most straightforward way to allow proper tracking of both the sample and the samplee post splitting. 

The major changes are 
- multiple returns to commandInsts and LHInstructions
- instruction sorting modifications to facilitate the new instruction type which ensure that the split 
   instructions go in their own layer AFTER use of the sample and BEFORE use of the remainder of the 
   samplee
- new robot instruction type (SampleBlock) which updates IDs of remaining components 

This is tested with its intended use and works fine so far, apart from any minor test failures not in sync with the compile changes this should not cause much collateral damage since it is opt-in based on use of the new intrinsic

### DIFF
--- a/antha/anthalib/mixer/mixer.go
+++ b/antha/anthalib/mixer/mixer.go
@@ -50,11 +50,22 @@ func Sample(l *wtype.LHComponent, v wunit.Volume) *wtype.LHComponent {
 	if l.Conc > 0 && len(l.Cunit) > 0 {
 		ret.SetConcentration(wunit.NewConcentration(l.Conc, l.Cunit))
 	}
-	//ret.Conc = l.Conc
-	//ret.Cunit = l.Cunit
+
 	ret.SetSample(true)
 
 	return ret
+}
+
+// SplitSample is a two-return version of sample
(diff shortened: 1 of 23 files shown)


antha/anthalib/mixer/mixer.go
```diff
@@ -50,13 +50,24 @@ func Sample(l *wtype.LHComponent, v wunit.Volume) *wtype.LHComponent {
...
 }
 
+// SplitSample is a two-return version of sample
+func SplitSample(l *wtype.LHComponent, v wunit.Volume) (moving, remaining *wtype.LHComponent) {
+	remaining = l.Dup()
+
+	moving = Sample(remaining, v)
+
+	remaining.Vol -= v.ConvertToString(remaining.Vunit)
+	remaining.ID = wtype.GetUUID()
```
haydnKing: I would have expected the ID of the part being moved to be different, rather than the part staying put - but this probably doesn't matter if elements never actually refer to samples by IDs

PR #656, opened 2018-03-09
Labware/add labcyte 384 pp std v

m-i-sadowski:
This PR adds one supported plate type for the Labcyte echo - the standard volume 384 plate used by their system

FYI @ddn0


m-i-sadowski:
If one of you (Xander most likely) can check the description I've forwarded on vs. the addition in the library it'd be a good idea.

inventory/testinventory/make_plate_library.go
```diff
@@ -830,3 +831,52 @@ func make96DeepWellLowVolumePlate() *wtype.LHPlate {
...
+	zstartOffsetCorrection := 2.5
+
+	plateName := "Labcyte_384PP_StdV"
+	wellName := "Labcyte_384PP_StdV"
+	manufacturer := "Labcyte"
+
+	numberOfRows := 16
+	numberOfColumns := 24
+
+	wellShape := "squre cylinder"
```
x-anderson: "square cylinder" - spellcheck to allow future proofing/sorting

inventory/testinventory/make_plate_library.go
```diff
@@ -830,3 +831,52 @@ func make96DeepWellLowVolumePlate() *wtype.LHPlate {
 
 	return plate
 }
+
```
x-anderson: Is it possible to include the URL to the place where you got the dimensions from, so easy to reference if needed?

inventory/testinventory/make_plate_library.go
m-i-sadowski: Fixed this - good spot  thanks!

inventory/testinventory/make_plate_library.go
m-i-sadowski: I've added the link to the PDF from labcyte in the comment above the method.

PR #658, opened 2018-03-13
relax stringency for string type to be a string

chr1sgrant:


chr1sgrant:
Encountered by a user wanting to name liquids using numbers: this should be permitted

PR #660, opened 2018-03-14
Add Gilson filter tips

haydnKing:
New Tips: Gilson DF200, DF30, and DF10
Gilson Driver returns only D200 and DL10 (unfiltered tips) as tiptype preferences, so users have to override this through UI or command line to change existing behaviour


microArch/driver/liquidhandling/choosechannel_test.go
```diff
@@ -169,6 +169,10 @@ func SetUpTipsFor(lhp *LHProperties) *LHProperties {
 
 	for _, tb := range testinventory.GetTipboxes(ctx) {
 		if tb.Mnfr == lhp.Mnfr || lhp.Mnfr == "MotherNature" {
+			//ignore filter tips
+			if tb.Type == "DF200 Tip Rack (PIPETMAX 8x200)" || tb.Type == "DFL10 Tip Rack (PIPETMAX 8x20)" || tb.Type == "DF30 Tip Rack (PIPETMAX 8x20)" {
```
ddn0: not a fan of this. Should filter tips be a property on tips?

PR #661, opened 2018-03-15
Maximum transfer volumes are now respected for the liquid handler

m-i-sadowski:
Labcyte Echo driver will now only see transfers of at most the maximum volume


microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -554,3 +560,59 @@ func (ins *TransferInstruction) Generate(ctx context.Context, policy *wtype.LHPo
...
 }
+
+func (ins *TransferInstruction) ReviseTransferVolumes(prms *LHProperties) error {
+	newTransfers := make([]MultiTransferParams, len(ins.Transfers))
+
+	for _, mtp := range ins.Transfers {
+		//newMtp := make(MultiTransferParams, len(mtp))
+		newMtp := NewMultiTransferParams(mtp.Multi)
+		for _, tp := range mtp.Transfers {
+			if tp.What == "" {
```
haydnKing: safeTransfers already handles this case?

m-i-sadowski: Needs handling here to prevent empty transfers... I left the other check in to stop safeTransfers crashing if it ever gets called in a different context

PR #663, opened 2018-03-16
Tests/multichannel tip reuse

haydnKing:
Tidied up some existing test code and added three new tests:
TestMultiChannelTipReuseGood - Test that moving a column of water to two different wells doesn't generate a tip change in between
TestMultiChannelTipReuseBad - Test that moving a column of water wells then separately ethanol generates a tip change
TestMultiChannelTipReuseUgly - Test that mixing water and ethanol in the same column generates tip changes ('ugly' because it doesn't necessarily need to change tips since contamination would have no real effect)


microArch/driver/liquidhandling/transferblock_test.go
```diff
@@ -629,3 +617,88 @@ func testNegative(ctx context.Context, ris []RobotInstruction, pol *wtype.LHPoli
...
+//TestMultiChannelTipReuseGood Move water to two columns of wells - shouldn't need to change tips in between
+func TestMultiChannelTipReuseGood(t *testing.T) {
+	ctx := testinventory.NewContext(context.Background())
+
+	inss, err := getMixInstructions(ctx, 16, []string{inventory.WaterType}, []float64{50.0})
+	if err != nil {
+		panic(err)
+	}
+
+	checkTipUsage(t, ctx, inss, 8)
```
m-i-sadowski: This is OK but I find it a little indirect - I would like to see an additional check to ensure that tip reuse is definitely occurring, i.e. we don't see load asp dsp unload load asp dsp... 

an additional check or two on single channel tip reuse might be useful also. 

One good way to be sure things is to modify TIP_REUSE_LIMIT in the policy supplied to 0 and rerun, in which case you should see the above is happening.

PR #667, opened 2018-03-19
Additional work for base qPCR driver

matttgregg:
Allow support for job id insertion.
Use multiple input LHC elements.
Clean ups to functionality and messaging

### DIFF
--- a/target/qpcrdevice/qpcrdevice.go
+++ b/target/qpcrdevice/qpcrdevice.go
@@ -10,7 +10,6 @@ import (
 	quantstudio "github.com/antha-lang/antha/driver/antha_quantstudio_v1"
 	"github.com/antha-lang/antha/target"
 	"github.com/golang/protobuf/proto"
-	"golang.org/x/net/html/atom"
 	"os"
 )
 
--- a/target/qpcrdevice/qpcrdevice.go
+++ b/target/qpcrdevice/qpcrdevice.go
@@ -2,12 +2,16 @@ package qpcrdevice
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"github.com/antha-lang/antha/ast"
 	"github.com/antha-lang/antha/driver"
 	framework "github.com/antha-lang/antha/driver/antha_framework_v1"
 	quantstudio "github.com/antha-lang/antha/driver/antha_quantstudio_v1"
 	"github.com/antha-lang/antha/target"
 	"github.com/golang/protobuf/proto"
+	"golang.org/x/net/html/atom"
+	"os"
 )
 
 // QPCRDevice defines the state of a qpcr device device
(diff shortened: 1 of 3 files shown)


target/qpcrdevice/qpcrdevice.go
```diff
@@ -36,6 +39,62 @@ func (a *QPCRDevice) MoveCost(from target.Device) int {
 	return 0
 }
 
+// singleCallForInstruction generates a device call, based on a single qPCR insruction.
+func singleCallForInstruction(inst *ast.QPCRInstruction) (driver.Call, error) {
+	message := &quantstudio.TemplatedRequest{
+		SessionInstrument: &quantstudio.SessionInstrument{
+			Session: &quantstudio.Session{
+				Id: "",
```
ddn0: Remove, since `Id: ""` is the same as not initializing the field at all.

target/qpcrdevice/qpcrdevice.go
```diff
@@ -36,6 +39,62 @@ func (a *QPCRDevice) MoveCost(from target.Device) int {
...
+			},
+			Instrument: &quantstudio.Instrument{
+				Id: "",
```
ddn0: ditto

target/qpcrdevice/qpcrdevice.go
```diff
@@ -36,6 +39,62 @@ func (a *QPCRDevice) MoveCost(from target.Device) int {
...
+			},
+		},
+		TemplateFile: &quantstudio.ExperimentFile{
+			Url: inst.Definition,
+		},
+		Barcode: &quantstudio.Barcode{
+			Barcode: inst.Barcode,
+		},
+		OutputPath: "",
```
ddn0: ditto

target/qpcrdevice/qpcrdevice.go
```diff
@@ -36,6 +39,62 @@ func (a *QPCRDevice) MoveCost(from target.Device) int {
...
+	}
+
+	messageBytes, err := proto.Marshal(message)
+	if err != nil {
+		return driver.Call{}, err
+	}
+
+	jobId := inst.TagAs
+	if jobId == "" {
```
ddn0: small preference for `len(jobID) == 0`

and go style says that the name should be `jobID`; there is an exception for machine generated code, but not for human-written code.

target/qpcrdevice/qpcrdevice.go
```diff
@@ -46,61 +105,38 @@ func (a *QPCRDevice) Compile(ctx context.Context, nodes []ast.Node) ([]target.In
...
+		if inst.Definition == "" {
+			return nil, errors.New("Blank experiment file for qPCR.")
 		}
 
-		messageBytes, err := proto.Marshal(message)
-		if err != nil {
-			return nil, err
+		if len(experimentFile) > 0 {
+			if experimentFile != inst.Definition {
+				return nil, fmt.Errorf("Unexpected multiple experiment files %s, %s", experimentFile, inst.Definition)
```
ddn0: error messages should start lower case to facilitate appending like so `fmt.Errorf("bad stuff: %s", err)`

target/qpcrdevice/qpcrdevice.go
```diff
@@ -46,61 +105,38 @@ func (a *QPCRDevice) Compile(ctx context.Context, nodes []ast.Node) ([]target.In
...
-			TemplateFile: &quantstudio.ExperimentFile{
-				Url: inst.Definition,
-			},
-			Barcode: &quantstudio.Barcode{
-				Barcode: inst.Barcode,
-			},
-			OutputPath: "",
+		// Note that we only accept the first value at present. (i.e. a single qPCR run within a workflow.)
+
+		if inst.Definition == "" {
```
ddn0: slight preference for `len(...) == 0`

target/qpcrdevice/qpcrdevice.go
```diff
@@ -46,61 +105,38 @@ func (a *QPCRDevice) Compile(ctx context.Context, nodes []ast.Node) ([]target.In
 	}
 
 	var calls []driver.Call
+	experimentFile := ""
```
ddn0: var experimentFile string

target/qpcrdevice/qpcrdevice.go
```diff
@@ -36,6 +39,62 @@ func (a *QPCRDevice) MoveCost(from target.Device) int {
...
+		return driver.Call{}, err
+	}
+
+	jobId := inst.TagAs
+	if jobId == "" {
+		jobId = os.Getenv("METADATA_JOB_ID")
+	}
+
+	instID := ""
+	if len(inst.ComponentIn) > 0 {
```
ddn0: if not, should that be an error?

matttgregg: It's not necessarily an error for the qPCR functionality - we don't actually use the input components *yet* - so partially minded to leave to the framework to enforce for the moment?

ddn0: sounds reasonable.

PR #672, opened 2018-03-20
make PCR parser robust to missing fields

chr1sgrant:


antha/AnthaStandardLibrary/Packages/Parser/parseXLSX_test.go
```diff
@@ -389,3 +392,69 @@ func TestParseExcel(t *testing.T) {
...
+	parsePCRDesignXLSXTest{
+		designFileName:     "PCRSetUpWorklist.xlsx",
+		expectedErrMessage: "",
+	},
+	parsePCRDesignXLSXTest{
+		designFileName:     "PCRSetUpWorklist_noTemplateSet.xlsx",
+		expectedErrMessage: "no template specified for PCRReaction24\nno primer 1 specified for PCRReaction24\nno primer 2 specified for PCRReaction24\nPlease specify all parts for reaction PCRReaction24 in Sheet1.",
+	},
+	parsePCRDesignXLSXTest{
+		designFileName:     "PCRSetUpWorklist_missingSequences.xlsx",
```
x-anderson: In this test case, would an error be expected if there are missing Sequences, as suggested by the filename?

chr1sgrant: The intention, at the moment, is that errors are not to be triggered by missing sequence information, since in the current basic PCR elements a sequence is not strictly required, only as an option for validation if the user requires. 

Prior to this PR an ungraceful failure occurred if no sequence was specified.

chr1sgrant: lines 92 and 97 prevent the index out of range error which was previously triggered when no sequence was specified. 

```go 
if len(sequenceField) > 1 {
	pcrReaction.PrimerPair[0].Seq = sequenceField[1]
}
```

PR #674, opened 2018-03-20
New plate/ab 384plate 2

CharlotteWard:
Update to previous PR to add in Applied Biosystems 384 well plate, addressing issues raised (min/max volume and deletion of labcyte plate). On different branch due to problem with labcyte plate having been deleted in the previous one.


chr1sgrant:
```inventory/testinventory/make_plate_library.go:989:47: cannot use minVolume (type int) as type float64 in argument to wtype.NewLHWell```

PR #690, opened 2018-03-28
Workaround for float discrepancies

johansglock:
0.0001 is elsewhere also used to check for float discrepancies so this
should be in line with other checks.


johansglock:
antha-tester passes successfully as well

PR #695, opened 2018-04-05
permit capital L for litre

chr1sgrant:


chr1sgrant:
fixes https://synthace.atlassian.net/browse/ANTHA-1332

PR #697, opened 2018-04-06
Fix well overfilling issues caused by indescrimitately adding the carry volume

haydnKing:
This involved quite a cleanup of LHWell, rationalising (and unit-testing!) the accessor methods for volumes as well as the AddComponent and RemoveVolume methods. 

Also LHWell.Empty() is renamed to LHWell.IsEmpty() to make it clear that it's a question (unlike LHTipWaste.Empty() which actually empties the tipwaste)

Carry Volume is still no longer added to wells during the `revise_volumes` function as this usually causes wells to be overfilled. Instead it's removed from the amount each well can contribute during the  `input_plate_setup` method. 

Carry volume still isn't accounted for properly, so i've created [ANTHA-1367](https://synthace.atlassian.net/browse/ANTHA-1367) and assigned it to technical debt to record that.


haydnKing:
(still getting some test failures that I'll go through with Sid on Monday)

PR #702, opened 2018-04-09
add element name to Errorf messages

ddn0:
`Errorf("some error")` will print `element XXX: some error`.

If you do `fmt.Printf("%+v", err)` you'll get a stack track too. Whether that should be the default, I can handle in a separate PR.

### DIFF
--- a/execute/error.go
+++ b/execute/error.go
@@ -3,19 +3,45 @@ package execute
 import (
 	"context"
 	"fmt"
+
+	"github.com/pkg/errors"
 )
 
 // An Error reported by user code
 type Error struct {
-	Message string
+	message string
 }
 
 // Error returns the error message
 func (a *Error) Error() string {
-	return a.Message
+	return a.message
 }
 
 // Errorf reports an execution error. Does not return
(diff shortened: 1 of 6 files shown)


execute/error.go
```diff
@@ -3,19 +3,45 @@ package execute
...
 
 // Errorf reports an execution error. Does not return
 func Errorf(ctx context.Context, format string, args ...interface{}) {
-	panic(&Error{Message: fmt.Sprintf(format, args...)})
+	userMsg := fmt.Sprintf(format, args...)
+	elementName := getElementName(ctx)
+
+	msg := userMsg
+	if len(elementName) != 0 {
+		msg = "element " + elementName + ": " + userMsg
```
chr1sgrant: Some error messages are very long now so might be best to separate the user message to a new line:

```go 
msg = "error in " + elementName + ": \n" + userMsg
```

ddn0: the presentation of error messages should be separate from the generation of them. Can add a field for introspection, but adding newlines is not a good idea at this point.

PR #703, opened 2018-04-09
Fixes to splitsample

m-i-sadowski:
- there was an error when the "stationary" component (i.e. the remainder after sampling) was not used which led to mixes and splits being combined into a single node of the execution graph. This causes downstream issues and so these are now separated into two nodes wherever this happens.

- remainders were also not registering as having locations set if their parent component's location is assigned automatically since these are copies rather than references. Invoking the sample tracker's ID-forwarding mechanism during the SplitSample command definition was sufficient to get the right behaviour.

### DIFF
--- a/microArch/driver/liquidhandling/getComponents.go
+++ b/microArch/driver/liquidhandling/getComponents.go
@@ -255,12 +255,6 @@ func (lhp *LHProperties) GetComponents(opt GetComponentsOptions) (GetComponentsR
 	rep := newReply()
 	// build list of possible sources -- this is a list of ComponentVectors
 
-	for _, cmp := range opt.Cmps {
-		if cmp == nil {
-			continue
-		}
-	}
-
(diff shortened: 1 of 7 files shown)


microArch/driver/liquidhandling/getComponents.go
```diff
@@ -255,6 +255,12 @@ func (lhp *LHProperties) GetComponents(opt GetComponentsOptions) (GetComponentsR
 	rep := newReply()
 	// build list of possible sources -- this is a list of ComponentVectors
 
+	for _, cmp := range opt.Cmps {
+		if cmp == nil {
+			continue
+		}
+	}
```
haydnKing: is this doing something?

m-i-sadowski: Doh... not any more! Removed.

PR #704, opened 2018-04-10
handle capital L for concentration units also

chr1sgrant:


chr1sgrant:
related to https://synthace.atlassian.net/browse/ANTHA-1332

chr1sgrant:
the characters I tested do not need UI updates (ng/uL) but this may apply to some of the new ones added

PR #705, opened 2018-04-10
antha metadata add-tags: initial

ddn0:
adds a new subcommand `antha metadata add-tags`.

For example, running
```
antha metadata add-tags MVL
```
will add the tag MVL to all the metadata for elements in the current directory and below.


chr1sgrant:
ace

PR #707, opened 2018-04-11
Use csvutil.BetterNewReader

cosmonautspiff:
Uses BetterNewReader function from "github.com/antha-lang/toolbox/csvutil" to deal with carriage returns.


target/mixer/parse_plate.go
```diff
@@ -2,17 +2,16 @@ package mixer
 
 import (
 	"context"
-	"encoding/csv"
 	"fmt"
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
```
ddn0: looks like you are not using gofmt or goimports.

cosmonautspiff: fmt is being used in parse_plate.go. What don't I understand?

ddn0:
Agreed, but that test case probably should be in the github.com/antha-lang/toolbox rather than here.

chr1sgrant:
why not in both?

cosmonautspiff:
There is a test case with carriage return. I've actually included small .csv files.

chr1sgrant:
Ahh I meant just to add a test example which has the example with carriage returns scoped within the existing test list for TestParsePlate. The added test looks more appropriate for the github.com/antha-lang/toolbox package as a direct unit test of the NewTolerantReader() func

cosmonautspiff:
@chr1sgrant I've updated the parse_plate test as we discussed. It reads in an external .csv file in order to preserve the carriage returns.

PR #708, opened 2018-04-12
Refactor of tip/well alignment checks

m-i-sadowski:
Two things

1) refactoring tip/well alignment checks used by multichannel planning to have a better signature: now depends on robot type, head of robot to be used, plate type and wells to be addressed

2) explicitly checking for clashes using well geometry for nonstandard plate types

rolling this out further and getting max efficiency in the case of, e.g., big wells has to be left as part of another story as it has several considerations. Hence one test is now deprecated and skipped pending picking up this subsequent story.


m-i-sadowski:
Purported fix for https://synthace.atlassian.net/browse/ANTHA-1292

PR #711, opened 2018-04-16
Fix/split sample revisions

m-i-sadowski:
Main fix in this branch is improving the volume fixing logic to account for split samples, which previously were ignored, causing a test failure for one bundle using the revised SerialDilutions protocol.

The mechanism is similar to the one for prompts in that the ID change is handled when assessing the amount of component required, additionally there is logic to accumulate if the new ID is already known when planning. 

One small fix comes in with this: previously inventory components were returned with a constant ID. This has not previously been consequential but was clearly incorrect behaviour. Now this is fixed.

Also as a freebie there is an additional benchmark test added to the transferblock test set as a convenience for profiling a substantial part of the system.

### DIFF
--- a/microArch/driver/liquidhandling/splitblock.go
+++ b/microArch/driver/liquidhandling/splitblock.go
@@ -39,6 +39,14 @@ func (sp SplitBlockInstruction) Generate(ctx context.Context, policy *wtype.LHPo
 		// if Components is a sample we'll probably want to change ParentID instead
 		// that may not work
 		robot.UpdateComponentID(ins.Components[0].ID, ins.Results[1])
+
+		/*
+			question over whether this is needed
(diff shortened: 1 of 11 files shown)


microArch/driver/liquidhandling/splitblock.go
```diff
@@ -39,6 +39,14 @@ func (sp SplitBlockInstruction) Generate(ctx context.Context, policy *wtype.LHPo
 		// if Components is a sample we'll probably want to change ParentID instead
 		// that may not work
 		robot.UpdateComponentID(ins.Components[0].ID, ins.Results[1])
+
+		/*
+			question over whether this is needed
```
haydnKing: I'd say... not? ;-)

PR #714, opened 2018-04-18
MixNamed no longer fragments plates

m-i-sadowski:
This fixes a reported issue with the unique plate naming - post the fix the system was splitting samples to multiple wells excessively. 

The fix was to add a check for an eligible plate with the same type and name. Uniqueness of plate naming assigned in this way is enforced in a later check.


m-i-sadowski:
Passes CI tests run locally

PR #717, opened 2018-04-19
Cleanup/serialized inventory

m-i-sadowski:
Profiling some slow tests showed that inventory generation is very costly since for each plate it is necessary to create at worst several thousand UUID and the inventory contains hundreds of plate types. Therefore it is highly inefficient to create the entire plate library on every simulation. 

This PR changes the inventory generation process from one of initializing every plate when the inventory is made to one of loading plate data and initializing plates only when they are requested from the inventory. 

Profiling indicates that this creates a substantial speedup for the antha-lang CI tests (since many of them generate their own inventory) and a moderate speedup for general execution - the latter is a consequence of removing a large-but-fixed cost operation. 

As a side-effect it slightly complicates addition of plates to the library however this is hopefully to be phased out shortly and will therefore be of much lower impact than the current issues around execution speed.


m-i-sadowski:
@ddn0 this is mostly FYI (although comments / corrections of course welcome)
@haydnKing @chr1sgrant please look at the test and see if you think we're missing any fields out. I will be running this through the bundles locally and I'll let you know if that goes OK.

ddn0:
the binary file `inventory/testinventory/serialize/serialize` probably should be removed from the PR

antha/anthalib/wtype/lhplate.go
```diff
@@ -878,12 +882,35 @@ func exportCSV(records [][]string, filename string) (File, error) {
...
+	}
+
+	return strings.Replace(s, CONSTRAINTMARKER, "", -1)
+}
+
+func isConstraintKey(s string) bool {
+	return strings.HasPrefix(s, CONSTRAINTMARKER)
+}
+
 func (p *LHPlate) SetConstrained(platform string, positions []string) {
```
haydnKing: What does being constrained actually signify?

haydnKing:
Looks good to me - the only other thing I can think of is string comparing the serialised list of original and new plates as an extra sanity check? unlikely to catch anything new though

antha/anthalib/wtype/lhplate.go
m-i-sadowski: Indicates that a plate can only be placed on certain positions on certain robots... it's clumsy, but it does the job. 

I'll consider if any sort of serialized comparison might work too.

m-i-sadowski:
I've been through all the plate and well type code and added tests for anything I didn't immediately see. The tests pass and I am confident this is ready to come in now.

antha/anthalib/wtype/lhplate.go
chr1sgrant: maybe should change to `SetPositionConstrained` to make this clear

inventory/testinventory/serialPlate.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+)
+
+var (
+	vunit = "ul"
+	lunit = "mm"
+)
+
+type PlateForSerializing struct {
```
chr1sgrant: Full annotation of `PlateForSerializing` and all fields would be extremely helpful for anyone trying to add plates

PR #719, opened 2018-04-20
Moved aggregation to proper place

m-i-sadowski:
Instruction aggregation mistakenly excluded single channel instructions 

Proposed fix for 1259


m-i-sadowski:
Passes CI tests... will add unit tests in same style as the multichannel tip reuse ones

PR #723, opened 2018-04-24
fix cases where solution name contains concentration unit as suffix

chr1sgrant:


cosmonautspiff:
Cool. Works fine for me from the command line.

PR #735, opened 2018-05-03
Accept sources which are only smaller due to numerical error

haydnKing:
This time I definitely ran it though antha-tester and it passed.

key change is in `AllVolsPosOrZero()` in file driver/liquidhandling/getComponents.go - the issue is as we suspected, the carryVol was protecting us from numerical instabilities that we're now exposed to. I played around with matching volumes in matchComponents as well, but it added extra complexity that wasn't actually necessary

### DIFF
--- a/antha/anthalib/wtype/componentvector.go
+++ b/antha/anthalib/wtype/componentvector.go
@@ -17,7 +17,7 @@ func (cv ComponentVector) String() string {
 //DeleteAllBelowVolume set all components whose volume is below vol to nil
 func (cv ComponentVector) DeleteAllBelowVolume(vol wunit.Volume) {
 	for i := 0; i < len(cv); i++ {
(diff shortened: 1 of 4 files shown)


antha/anthalib/wtype/componentvector.go
```diff
@@ -16,7 +16,7 @@ func (cv ComponentVector) String() string {
 
 func (cv ComponentVector) DeleteAllBelowVolume(vol wunit.Volume) {
 	for i := 0; i < len(cv); i++ {
-		if cv[i] != nil && cv[i].Volume().LessThan(vol) {
+		if v := cv[i].Volume(); v.LessThan(vol) && !wunit.SubtractVolumes(vol, v).IsZero() {
```
m-i-sadowski: Can you please add a comment here to explain this... I don't think the rationale will be easy to see at a glance!

PR #775, opened 2018-06-04
Antha 1529/update primer errors and seqsavoid function

x-anderson:
To support updates to the Primer Design Multi elements:

- Improved formatting of errors
- Updated evaluation of `tempoligoseq` in `seqstoavoid` using new function `PartialInStrings`. Previously, the `tempoligoseq` had to match exactly to the `seqstoavoid` to be flagged. However, user feedback indicated the expected function of this was to assess the presence of a `tempoligoseq` as a subset of`seqstoavoid` (i.e. does not match exactly in length).
- Added test and updated previous ones


antha/AnthaStandardLibrary/Packages/search/InStrings.go
```diff
@@ -73,3 +73,24 @@ func PositionsInStrings(list []string, target string, options ...Option) []int {
...
+// PartialInStrings seaches for a target string in a slice of strings and returns a boolean.
+// True is returned if the target is a subset of a value in the slice of strings (i.e. value "foobar", target "bar" will return True).
+// If the IgnoreCase option is specified the strings will be compared ignoring case.
+func PartialInStrings(list []string, target string, options ...Option) bool {
+
+	ignore := containsIgnoreCase(options...)
+
+	for _, entry := range list {
+		if ignore {
+			if strings.Contains(entry, target) {
```
chr1sgrant: is this the right way round?
shouldn't it be 

```go
if ignore {
if strings.Contains(strings.ToUpper(strings.TrimSpace(entry)), strings.ToUpper(strings.TrimSpace(target))){
  return true
}
} else {
if strings.Contains(entry, target) {
return true
}
}
```

antha/AnthaStandardLibrary/Packages/sequences/oligos/oligos.go
```diff
@@ -281,7 +275,7 @@ func REVOligoSeq(seq wtype.DNASequence, maxGCcontent float64, minlength int, max
 				}
 			}
 
-			if temppercentage <= maxGCcontent && minmeltingtemp.SIValue() < meltingtemp.SIValue() && maxmeltingtemp.SIValue() > meltingtemp.SIValue() && bindingsites == 1 && !search.InStrings(seqstoavoid, tempoligoseq) && !overlapthresholdfail {
+			if temppercentage <= maxGCcontent && minmeltingtemp.SIValue() < meltingtemp.SIValue() && maxmeltingtemp.SIValue() > meltingtemp.SIValue() && bindingsites == 1 && !search.PartialInStrings(seqstoavoid, tempoligoseq) && !overlapthresholdfail {
```
chr1sgrant: the PartialInStrings should be fixed to use the ignoreCase the right way round and then use updated here:
```
!search.PartialInStrings(seqstoavoid, tempoligoseq, search.IgnoreCase) ```

antha/AnthaStandardLibrary/Packages/search/InStrings.go
chr1sgrant: Not sure about the term PartialInStrings ... slight preference for ContainedInStrings or an alternative would be to 
make an Option IncludePartialMatches to use in the search.InStrings func (and the other funcs such as InSequences). I'm ok with `PartialInStrings` though just wanted to bring this up for discussion

PR #777, opened 2018-06-05
Add plateCache to context in liquidhandler.Plan(...)

haydnKing:
re enable caching of temporary plates, but add the cache to the context in Liquidhandler.Plan, which means the caller doesn't have to remember to add it beforehand to avoid crash


microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/antha-lang/antha/antha/anthalib/wutil"
 	"github.com/antha-lang/antha/inventory"
 	"github.com/antha-lang/antha/inventory/cache"
+	"github.com/antha-lang/antha/inventory/cache/plateCache"
```
ddn0: oh the naming of this package...

https://golang.org/doc/effective_go.html#package-names

PR #787, opened 2018-06-08
Turn things up to 11

chr1sgrant:


x-anderson:
Yes! 

Does the DEFAULTPIPETTESPEED also override the pre- and post-mix rates:

`PRE_MIX_RATE`
`POST_MIX_RATE`

both could be changed to 3.7?

chr1sgrant:
I believe this does set Pre and Post mix rate (if not explicitely set)

PR #798, opened 2018-06-15
Fix/antha 1842

haydnKing:
as per comment in [ANTHA-1842](https://synthace.atlassian.net/browse/ANTHA-1842), slices of components to be multichannelled are now split such that only components of the same volume are multichanneled simultaneously for a non-independent head

unit test derived from ANTHA-1842 ticket, tested on command line against bundle and with antha-tester

### DIFF
--- a/microArch/driver/liquidhandling/convertinstructions.go
+++ b/microArch/driver/liquidhandling/convertinstructions.go
@@ -24,6 +24,7 @@ package liquidhandling
 
 import (
 	"context"
+	"fmt"
 	"github.com/antha-lang/antha/antha/anthalib/wtype"
 	"github.com/antha-lang/antha/antha/anthalib/wunit"
 )
@@ -118,7 +119,9 @@ func convertInstructions(inssIn LHIVector, robot *LHProperties, carryvol wunit.V
 		l = inssIn.MaxLen()
 	}
 
+	fmt.Printf("for i := 0; i < l = %d; i++\n", l)
 	for i := 0; i < l; i++ {
+		fmt.Printf("  i := %d\n", i)
 		var inssToUse LHIVector
 		var cmps []*wtype.LHComponent
 		if horiz {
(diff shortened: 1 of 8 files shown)


microArch/driver/liquidhandling/convertinstructions.go
```diff
@@ -179,6 +201,50 @@ func convertInstructions(inssIn LHIVector, robot *LHProperties, carryvol wunit.V
...
+	for i := 0; i < len(inComponentsByVolume); i++ {
+		c, ins := segmentByVolume(inComponentsByVolume[i], inInstructionsToUse[i])
+		retComponentsToMove = append(retComponentsToMove, c...)
+		retInstructionsToUse = append(retInstructionsToUse, ins...)
+	}
+
+	return retComponentsToMove, retInstructionsToUse
+}
+
+//split up the idsets such that each instruction in the set has the same volume
```
ddn0: is it always the case that the inputs are sorted? might be good to update the function doc to note this.

microArch/driver/liquidhandling/convertinstructions.go
```diff
@@ -179,6 +201,50 @@ func convertInstructions(inssIn LHIVector, robot *LHProperties, carryvol wunit.V
...
+func segmentByVolume(components []*wtype.LHComponent, instructions LHIVector) ([][]*wtype.LHComponent, []LHIVector) {
+
+	var retComponents [][]*wtype.LHComponent
+	var retInstructions []LHIVector
+
+	var currComponents []*wtype.LHComponent
+	var currInstructions LHIVector
+
+	lastVolume := wunit.ZeroVolume()
+	for i, component := range components {
```
ddn0: Nit: this loop seems more complicated than necessary.

parallel iteration of more than once slice is a little cleaner like this:
```
for i := range components {
  curComponent = components[i]
  curInstruction = instructions[i]
  ...
```

and I feel like you can get rid of the epilogue to the loop but I can't see how yet :D

PR #800, opened 2018-06-19
remove automatic fix of plate heights for LV with no riser when using…

chr1sgrant:
… greiner384 and pcrplate skirted


m-i-sadowski:
This should now be superseded by the plate library changes

PR #801, opened 2018-06-20
Make physical simulation errors fatal by default

haydnKing:
two effects:
1) make errors in physical simulation fatal
2) add a config option, "disablePhysicalSimulation" to bypass in the case of false positives

Goal is that physical simulation will be able to catch using low volume tips without risers (currently in a branch), as well as getting feedback on catching physical collisions


haydnKing:
Pause on merging this for now - there are some issues with physical collisions in antha-tester to get to the bottom of

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -3575,3 +3594,48 @@ func checkAndSaften(proposed, min, max float64, overrideIfOutOfRange bool) (floa
 
 	return proposed, nil
 }
+
+//make ZOffset safe
```
m-i-sadowski: A bit more here, please! This code is much longer than I had expected

microArch/simulator/liquidhandling/errors.go
```diff
@@ -80,3 +94,171 @@ func (self *GenericError) setInstruction(index int, ins driver.TerminalRobotInst
...
+
+func (self *CollisionError) SetInstructionDescription(d string) {
+	self.description = d
+}
+
+func (self *CollisionError) GetStateAtError() string {
+	return self.stateAtError
+}
+
+func (self *CollisionError) CollisionDescription() string {
```
m-i-sadowski: Comment please

microArch/simulator/liquidhandling/robotstate.go
```diff
@@ -167,11 +172,13 @@ func (self *ChannelState) GetCollisions(channelClearance float64) []wtype.LHObje
 			}
 		}
 	} else {
-		//reporting that we've collided with a well is a bit silly since wells are empty space
-		//but since channels shouldn't be inside wells, report collision with the plate instead
 		for _, obj := range objects {
 			if well, ok := obj.(*wtype.LHWell); ok {
-				ret = append(ret, well.GetParent())
+				//wells are empty space, so we can't collide with them, but we could collide with the parent
```
m-i-sadowski: Do we consider the walls of the wells?

microArch/simulator/liquidhandling/robotstate.go
```diff
@@ -222,6 +244,99 @@ func NewAdaptorState(name string,
...
+			pTips(len(loaded)),
+			strings.Join(tipTypes, ","),
+			pLengths(len(tipLengths)),
+			strings.Join(tipLengths, ","),
+			summariseChannels(loaded))
+	}
+	return fmt.Sprintf("%s: <no tips>", self.name)
+}
+
+func (self *AdaptorState) SummarisePositions(channelsColliding []int) string {
```
m-i-sadowski: Something explaining the output of this would be useful!

microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -261,6 +262,17 @@ func (this *Liquidhandler) Simulate(request *LHRequest) error {
 	}
 	logger.Info(strings.Join(logLines, "\n"))
 
+	//return the worst error if it's actually an error
```
m-i-sadowski: I'd consider including information about other errors as well... possibly user-configurable? It would be preferable not to have to resimulate repeatedly. Possibly not now but a future enhancement to produce a full summary of collisions caused by particular entities might be very useful.

msackman-synthace: Yeah (sorry to wade in here), I really don't understand why pkg/errors doesn't have a `type errors []error` - the number of times I've implemented that ... it's pretty trivial to do. And yes, as a UX issue, giving as many sensible errors as possible (especially when we can tell each error is independent to some extent) is preferable IME.

microArch/simulator/liquidhandling/robotstate.go
haydnKing: I'd personally consider the walls of the wells part of the plate, though that's inconsistent with our interpretation of bottomh :s

microArch/scheduler/liquidhandling/liquidhandler.go
haydnKing: Agree entirely. I held off from both because currently we just spew out the errors to the console in the front end and have no method for dealing with a list of errors - printing out all of the errors it likely to just be overwhelming for most users. In practice simulation errors tend to be very repetitive - if a particular plate configuration is causing problems it'll do so repeatedly. It would be nice to be able to filter out repeats of collisions caused by the same underlying cause and group them together so the front end can collapse repeats of the same type, but I think that's a new feature

PR #803, opened 2018-06-22
track subcomponents for all products of mixes

chr1sgrant:
This will: 
1. Ensure sub component concentrations are tracked following mixes.
2. Rename components with concentrations in a standardised way and ensure the name is accurate


antha/anthalib/mixer/mixer.go
```diff
@@ -197,6 +199,7 @@ func GenericMix(opt MixOptions) *wtype.LHInstruction {
 			}
 		}
 
+		solutions.UpdateComponentDetails(r.Results[0], opt.Components...) //nolint
```
m-i-sadowski: Is there a *good* reason to ignore the error here?

chr1sgrant: they're warnings really, of the sort you mention above "No componentlist found for water, no concentration found for water so defaulting to 0.5 v/v water"

PR #824, opened 2018-07-12
ANTHA-1939: Add sub components to plate csv

chr1sgrant:
The representation of sub components for Liquids in Antha is currently not supported in our plate file format but needs to be.

Testing performed: addition of a test case where sub components are specified in the CSV file.

There may be additional changes needed for the plate download step in the plate prep page of the UI, if it does not call this code.


target/mixer/parse_plate.go
```diff
@@ -268,6 +268,24 @@ func parsePlateCSVWithValidationConfig(ctx context.Context, inData io.Reader, vc
 		cmp.Conc = concentration.RawValue()
 		cmp.Cunit = concentration.Unit().PrefixedSymbol()
 
+		if len(rec) > 7 {
```
haydnKing: Can we check for "sub components" in the header instead? Will be a bit more robust if the columns change

target/mixer/parse_plate.go
```diff
@@ -268,6 +268,24 @@ func parsePlateCSVWithValidationConfig(ctx context.Context, inData io.Reader, vc
...
+			for k := 7; k < len(rec); k = k + 2 {
```
haydnKing: how about `const numberOfStaticCSVColumns = 7`, or `k := len(getCSVHeader())-1`

target/mixer/parse_plate.go
chr1sgrant: Yes I agree that would be better, I left it consistent with how the other column headers are handled (purely by position ... I think this is due to the fact that the first two "headers" are in fact not headers but the plate type and the plate name). Interestingly, the SubComponents header is a special case since it represents the start of where several columns may be specified... so extending this to the other properties would be now more problematic.

PR #825, opened 2018-07-12
Simplify declaration of aparams.

msackman-synthace:
This is just a consequence of me reading code and seeing things that can be tidied a little. If this is not helpful that please say and we can just throw this away completely.
By using literal maps, and moving away from strings a bit, the code is a bit clearer to read, and it caught a bug too (ASPREFERENCE was defined twice (line 80 and line 125)).


m-i-sadowski:
Thanks, that is indeed very helpful!

PR #826, opened 2018-07-13
Fix nil pointer exception in instruction ordering edge case

haydnKing:
When a mixture of mix and splits are sorted to the last IChain, splitting the IChain failed

PR fixes and adds unit testing to prevent a regression


microArch/scheduler/liquidhandling/executionhelpers_test.go
```diff
@@ -0,0 +1,208 @@
...
+	mix, _ := GetMixForTest("firstMix", waterSample, lemonJuice)
+
+	test := setOutputOrderTest{
+		Instructions:  []*wtype.LHInstruction{split, mix},
+		OutputSort:    true,
+		ExpectedOrder: []string{"firstMix", "theSplit"},
+		ChainHeight:   2,
+	}
+
+	test.Run(t)
```
msackman-synthace: Personally, I'd drop the test var completely and just tack the .Run(t) onto the end. But that's just me...

microArch/scheduler/liquidhandling/executionhelpers_test.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+func TestSetOutputOrdering_SplitMixes(t *testing.T) {
+
+	water := GetLiquidForTest("water", 250.0)
+	concentrate := GetLiquidForTest("concentratedSquash", 5000.0)
+	vodka := GetLiquidForTest("vodka", 50.0)
```
msackman-synthace: Need to consult cocktail recipes book... ;)

microArch/scheduler/liquidhandling/executionhelpers_test.go
```diff
@@ -0,0 +1,208 @@
...
+	expectedIDOrder = append(expectedIDOrder, mix2.ID)
+	expectedIDOrder = append(expectedIDOrder, split.ID)
+	instructions = append(instructions, mix2)
+
+	// mix the static component with some more water
+	c7 := GetLiquidForTest("water", 200.0)
+	mix3, _ := GetMixForTest("mix3", c4a, c7)
+
+	expectedIDOrder = append(expectedIDOrder, mix3.ID)
+	instructions = append(instructions, mix3)
```
msackman-synthace: Hmm. I don't really mind building up the expectedIdOrder and Instructions as you go along. Otoh, the order is deterministic, and you don't use expected or instructions until right at the end. So you could still just have full inline declarations within your setOutputOrderTest. It would save a few lines ... but I can see arguments for each way. Don't really mind.

microArch/scheduler/liquidhandling/ichain.go
```diff
@@ -226,10 +225,10 @@ func (it *IChain) splitMixedNode() {
 	ch := NewIChain(it)
```
msackman-synthace: ok, I've been trying to figure out what I don't like about this... I think I've found it. IMO, the NewIChain constructor should do more - if it receives a non-nil parent, then it should also set parent.child to be the new ichain. I.e. if we're maintaining a doubly-linked-list, which is what we are doing, then we should always mutate that completely rather than having to do the tidy up work in here. So if NewIChain changes as I suggest, then we then would have something like:

```
grandChild := it.Child
ch := NewIChain(it) // now it.Child points to ch, and ch.Parent points to it
if grandChild != nil {
  ch.Child, grandChild.parent = grandChild, ch
}
```

I feel that's clearer.

microArch/scheduler/liquidhandling/executionhelpers_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+import (
+	"github.com/antha-lang/antha/antha/anthalib/mixer"
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"github.com/antha-lang/antha/antha/anthalib/wunit"
+	"testing"
+)
+
+func GetMixForTest(id string, input ...*wtype.Liquid) (*wtype.LHInstruction, *wtype.Liquid) {
+	output := input[0].Dup()
```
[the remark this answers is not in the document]
haydnKing: good change - I did it this way because I was simply refactoring how it was originally written in the test

microArch/scheduler/liquidhandling/executionhelpers_test.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	mix.AddProduct(output)
+
+	return mix, output
+}
+
+func GetSplitForTest(id string, input *wtype.Liquid, volume float64) (*wtype.LHInstruction, *wtype.Liquid, *wtype.Liquid) {
+	split := wtype.NewLHSplitInstruction()
+	split.ID = id
+	moving, remaining := mixer.SplitSample(input, wunit.NewVolume(volume, "ul"))
```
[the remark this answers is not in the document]
haydnKing: yeah... there's a lot of things wrong with the unit library that need a good refactor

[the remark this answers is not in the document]
haydnKing: #teamintegervolume

microArch/scheduler/liquidhandling/ichain.go
haydnKing: how about
```
//NewIChain create a new item in a doubly linked list
func NewIChain() *IChain {...}

//Append add a new link after this element and before the next
func (this *IChain) Append(next *IChain) {
  //sets child & parent correctly for this, this.child and the next ichain, updates depth
}

//Prepend add a new link before this element
func (this *IChain) Prepend(prev *IChain) {
  //as above
}
```
This is more the kind of interface I'd be expecting from a linked list

msackman-synthace: Agreed - it's far less surprising to have an API that matches what we all expect from a linked list.

Personally, I'd drop prepend on the basis that the caller can always just call append with the args the other way around. Also, I'm not sure append is the right verb - if you call append on the head of a list, you might expect that to recurse down the entire chain and append it onto the very tail. I think insertAfter or something like that would be more obvious?

msackman-synthace: Ahh, the other thing I guess is that I might suggest calling the constructor NewIChain is confusing as it almost suggests it is creating an empty container, rather than a node itself. I wonder whether it should be renamed to NewIChainNode?

haydnKing: InsertAfter it is! I think this change would bloat this PR a bit too much though - I can refactor IChain in another branch

msackman-synthace: Ok, fine by me.

microArch/scheduler/liquidhandling/executionhelpers_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	return split, moving, remaining
+}
+
+func GetPromptForTest(message string, inputs ...*wtype.Liquid) (*wtype.LHInstruction, []*wtype.Liquid) {
+	ret := wtype.NewLHPromptInstruction()
+	ret.ID = message //ID will be overwritten, set the message as well for testing
+	ret.Message = message
+	for _, input := range inputs {
+		output := input.Cp()
```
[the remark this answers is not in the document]
haydnKing: Yeah, turns out liquids have `Dup` and `Cp` which do slightly different things. Seems like a bunch of elements depend on the behaviour though so not a trivial thing to fix

PR #828, opened 2018-07-13
Refactor robot instructions.

msackman-synthace:
The main goal here has been to move away from so much reliance of strings, which tend to be highly fragile and prone to mis-typing. In addition, by using anonymous embedding of structs, we can start introducing default behaviours, and reduce the amount of boilerplate, and reduce some of the level of casting going on (for example, see how I've moved down the merging of MOV instructions with ASP/DSP/MIX).
The tests pass.

I've chosen to stop at the point where the instruction parameters start coming out of the LHPolicy because it's not clear to me how to unify them, and this PR is big enough already. There's a good chance I've made some mistaken changes somewhere which the tests won't catch, and sadly this PR is fairly big so it may be a bit tricky to review.


msackman-synthace:
Also, check out the changes to timerfactory.go - got rid of all those magic numbers!

msackman-synthace:
Ok, with help from @chr1sgrant and @haydnKing I'm now running antha-runner and I can see there are some breakages in there too, so I'll fix those up... which means we're going to have at least two linked PRs in different repos.... yay.

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -859,22 +833,20 @@ type ChangeAdaptorInstruction struct {
...
+		NewAdaptorType:  newad,
+		Platform:        platform,
+	}
+	v.BaseRobotInstruction = NewBaseRobotInstruction(v)
+	return v
+}
+
+func (ins *ChangeAdaptorInstruction) GetParameter(name InstructionParameter) interface{} {
 	switch name {
 	case "HEAD":
```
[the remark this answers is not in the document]
msackman-synthace: Just testing ;)

msackman-synthace:
Needless to say, the build fails because it doesn't know to track the same branch in the other repo...

microArch/driver/liquidhandling/robotinstruction.go
```diff
@@ -47,116 +48,174 @@ type TerminalRobotInstruction interface {
...
+	MVM = NewInstructionType("MVM", "MoveMix")
+	MBL = NewInstructionType("MBL", "MoveBlowout")
+	RAP = NewInstructionType("RAP", "RemoveAllPlates")
+	APT = NewInstructionType("APT", "AddPlateTo")
+	RPA = NewInstructionType("RPA", "RemovePlateAt")
+	SPB = NewInstructionType("SPB", "SplitBlock")
+)
+
+type InstructionType struct {
+	MachineName string `json:"Type"`
```
m-i-sadowski: Can this be renamed to something like InternalName? I find the use of the term Machine here makes me think about the liquid handler

microArch/driver/liquidhandling/robotinstruction.go
```diff
@@ -501,14 +529,21 @@ func concatboolarray(a []bool) string {
...
+func NewBaseRobotInstruction(ins RobotInstruction) BaseRobotInstruction {
+	return BaseRobotInstruction{
+		Ins: ins,
+	}
+}
+
+func (bri BaseRobotInstruction) Check(rule wtype.LHPolicyRule) bool {
 	for _, vcondition := range rule.Conditions {
-		v := gri.Ins.GetParameter(vcondition.TestVariable)
+		// todo - this cast to InstructionParameter is gross, but we're
```
m-i-sadowski: I'm not clear on what this comment means - what work are you intending to do later on?

microArch/driver/liquidhandling/robotinstruction.go
msackman-synthace: Ok. I meant MachineName as opposed to HumanName. On the other hand, in the course of doing this, I think I came across zero uses of the HumanName anyway so I wonder if it's just better to kill off the whole HumanName thing, and then it can just become Name?

microArch/driver/liquidhandling/robotinstruction.go
msackman-synthace: Well. The whole GetParameter thing feels odd to me because you have nice structs with sensibly named and typed fields, and then you're allowing totally dynamic string-based access to that returning an interface{} which then requires dynamic type inspection... it's all a bit not-ideal ;) I have a feeling that the way to go is the Visitor Pattern, which would then allow all the GetParameter stuff to go away, and avoid use of interface{} completely.

But that would be a bigger refactor, and I would need to understand LHPolicies and how they work and what they do, which I currently don't. So there are currently I think 2 places where we cast from a LHPolicy-side string through to an InstructionParameter, which seems like a sensible boundary for the time being.

microArch/driver/liquidhandling/robotinstruction.go
m-i-sadowski: Yeah I realised of course, but I will forget and then it will confuse me again. Just having Name makes sense to me

microArch/driver/liquidhandling/robotinstruction.go
m-i-sadowski: OK cool, I'll be interested to see what you have in mind

microArch/driver/liquidhandling/robotinstruction.go
msackman-synthace: Dang, just doing this change, and I found HumanName is used in a couple of places. I'll just change MachineName to Name.

msackman-synthace:
Ok, I think this is ready for another go around.

PR #831, opened 2018-07-17
Don't attempt to compile LHInstructions if element execution fails

haydnKing:
Previously if an element calls Errorf(...), the trace will attempt to compile the issued LHInstructions before stopping the simulation with the error. If compiling the instructions fails, the new error would override the element error, leaving element developers with no way to correct the error.

This change prevents compilation if an element calls Errorf, ensure that the error is seen in the output.

Tested against antha-tester with no errors

### DIFF
--- a/trace/pool.go
+++ b/trace/pool.go
@@ -112,18 +112,12 @@ func getPool(ctx context.Context) *poolCtx {
 
 //tryUnblockWithLock will attempt to compile any LHInstructions in the trace
 func (pctx *poolCtx) tryUnblockWithLock(tr *trace) {
-	if pctx == nil {
-		return
-	}
 	if err := tr.signal(pctx); err != nil {
 		pctx.cancelWithLock(err)
 	}
 }
 
 func (pctx *poolCtx) decrement(tr *trace, delta int, err error) {
-	if pctx == nil {
-		return
-	}
 	pctx.lock.Lock()
 	defer pctx.lock.Unlock()
 	pctx.alive -= delta
--- a/trace/pool.go
+++ b/trace/pool.go
@@ -110,7 +110,11 @@ func getPool(ctx context.Context) *poolCtx {
 	return c
 }
 
-func tryUnblock(tr *trace, pctx *poolCtx) {
+//tryUnblockWithLock will attempt to compile any LHInstructions in the trace
+func (pctx *poolCtx) tryUnblockWithLock(tr *trace) {
+	if pctx == nil {
+		return
+	}
 	if err := tr.signal(pctx); err != nil {
 		pctx.cancelWithLock(err)
 	}
@@ -127,8 +131,7 @@ func (pctx *poolCtx) decrement(tr *trace, delta int, err error) {
 	if err != nil {
 		pctx.cancelWithLock(err)
 	} else {
-		//tryUnblock will attempt to compile any LHInstructions in the trace
-		tryUnblock(tr, pctx)
+		pctx.tryUnblockWithLock(tr)
 	}
 
 	if pctx.alive == 0 {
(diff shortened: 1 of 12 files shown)


trace/pool.go
```diff
@@ -120,17 +120,16 @@ func decrement(pctx *poolCtx, tr *trace, delta int, err error) {
...
 
-	var cancel error
 	if err != nil {
-		cancel = err
-	}
-	if pctx.alive == 0 && cancel == nil {
-		cancel = errPoolDone
+		pctx.cancelWithLock(err)
+	} else {
+		//tryUnblock will attempt to compile any LHInstructions in the trace
```
msackman-synthace: should this comment be on tryUnblock itself rather than here? Also, it does seem to me that both tryUnblock and decrement should be methods on poolCtx rather than package functions.

haydnKing: Agreed for decrement, does tryUnblock belong on the trace instead though?

msackman-synthace: Who knows? Also, tryUnblock should, for consistency, probably be called tryUnblockWithLock, which would make it less surprising that it calls cancelWithLock. The thing is though that it's holding the pool lock, and not the trace lock, so on that basis, I think the receiver should be the pool.

trace/pool.go
```diff
@@ -110,22 +110,28 @@ func getPool(ctx context.Context) *poolCtx {
 	return c
 }
 
-func tryUnblock(tr *trace, pctx *poolCtx) {
+//tryUnblockWithLock will attempt to compile any LHInstructions in the trace
+func (pctx *poolCtx) tryUnblockWithLock(tr *trace) {
+	if pctx == nil {
```
msackman-synthace: hmm, we didn't have these extra checks before - in the old version, if pctx was nil, it would have just blown up. What's motivating these extra checks? (same in decrement)

haydnKing: trying to avoid blowing up ;)
thought the convention was that functions should tolerate nil receivers? (though there's lots of places were we don't follow that)

msackman-synthace: Unless it's explicitly required, I would avoid trying to cope with nil receivers - to me it seems dangerous to have code that you would expect to do something turn into a noop because of a mistake further up the call stack - i.e. I would have thought that most of the time, a nil receiver is a mistake, and certainly I tend to code for it rarely (though it is useful from time to time). Given how the code was before, unless we now know that we should be coping with these methods being called with a nil pool, I don't see why we would attempt to silently swallow such a situation.

haydnKing: yeah, guess the question to ask is "does nil have a meaning here?"

trace/pool.go
```diff
@@ -110,27 +110,26 @@ func getPool(ctx context.Context) *poolCtx {
...
-		cancel = errPoolDone
+		pctx.cancelWithLock(err)
+	} else {
+		pctx.tryUnblockWithLock(tr)
 	}
-	if cancel != nil {
-		pctx.cancelWithLock(cancel)
+
+	if pctx.alive == 0 {
+		pctx.cancelWithLock(errPoolDone)
```
msackman-synthace: Ok, so looking at this again, we now have a path through where cancelWithLock will be called twice - err != nil and alive is now 0. This wouldn't happen before. So I think these three cases need combining into one block:

```
if err != nil {
  pctx.cancelWithLock(err)
} else if pctx.alive == 0 {
  pctx.cancelWithLock(errPoolDone)
} else {
  pctx.tryUnblockWithLock(tr)
}
```

haydnKing: hmm, previously we'd call tryUnblockWithLock before cancelWithLock(errPoolDone), so I kept that the same

PR #832, opened 2018-07-17
Don't Panic() when building the instruction graph fails, return error instead

haydnKing:
Building the topological graph can now fail if SplitSample is used incorrectly. This change returns error instead of calling panic

### DIFF
--- a/microArch/scheduler/liquidhandling/executionhelpers_test.go
+++ b/microArch/scheduler/liquidhandling/executionhelpers_test.go
@@ -77,14 +77,9 @@ func (self *setOutputOrderTest) Run(t *testing.T) {
 	rq.Options.OutputSort = self.OutputSort
 
 	err := setOutputOrder(rq)
-	if self.ExpectingError {
-		if err == nil {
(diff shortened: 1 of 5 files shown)


microArch/scheduler/liquidhandling/executionhelpers_test.go
```diff
@@ -75,7 +76,14 @@ func (self *setOutputOrderTest) Run(t *testing.T) {
 
 	rq.Options.OutputSort = self.OutputSort
 
-	if err := setOutputOrder(rq); err != nil {
+	err := setOutputOrder(rq)
+	if self.ExpectingError {
```
msackman-synthace: there's a neat way of writing this once you spot that != is logical xor.

```
if errFound := err != nil; errFound != self.ExpectingError {
  t.Fatalf("Error Expected: %v; Error encountered: %v", self.ExpectingError, err)
  return
}
```

haydnKing: had a feeling there was a better way

PR #837, opened 2018-07-19
Remove promises.

msackman-synthace:
The promises implementation was basically not being used - everything being issued as a promise was in fact a constant value, and so the entire promises mechanism was about an extra 1000 lines of code creating confusion and complexity. It can always be brought back when it's needed - and simply too - as can be seen from the diff here, the changes to intrinsics are trivial so can easily be undone.

local tests pass, and the fuller antha-runner/antha-tester tests pass too.


execute/trace.go
```diff
@@ -0,0 +1,54 @@
...
+
+// This is pretty gross. What we do here is add into a context a
+// container into which we later inject instructions.  Eventually,
+// when the stack unwinds, we grab all these instructions. This is an
+// abuse of the intent of context, but to fix this will require
+// changes to the code generation of elements.
+
+type Trace struct {
+	lock   sync.Mutex
+	instrs []*commandInst
```
haydnKing: instructions? the rs confuses me

execute/trace.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (tr *Trace) Issue(instr *commandInst) {
+	tr.lock.Lock()
+	defer tr.lock.Unlock()
+
+	tr.instrs = append(tr.instrs, instr)
+}
+
+func (tr *Trace) Instrs() []*commandInst {
```
haydnKing: [Get|Clone|Copy]Instructions()?

execute/trace.go
```diff
@@ -0,0 +1,54 @@
...
+	tr.lock.Lock()
+	defer tr.lock.Unlock()
+
+	clone := make([]*commandInst, len(tr.instrs))
+	copy(clone, tr.instrs)
+	return clone
+}
+
+func Issue(ctx context.Context, instr *commandInst) {
```
haydnKing: e.g. //Issue store the instruction for later compilation

workflow/workflow.go
```diff
@@ -186,37 +189,27 @@ func makeRoots(nodes map[string]*node) ([]*node, error) {
...
-			}
+	for len(worklist) > 0 {
+		n := worklist[0]
+		worklist = worklist[1:]
 
-			if len(newRoots) == 0 && len(a.nodes) > 0 {
-				return errCyclicWorkflow
-			}
-			roots = newRoots
+		if moreWork, err := a.run(ctx, n); err != nil {
```
haydnKing: personally I find 
```
moreWork, err := a.run(ctx, n)
if err != nil {
	return fmt.Errorf("cannot run process %q: %s", n.Process, err)
}
worklist = append(worklist, moreWork...)
```
easier to follow as the "happy path" is more obvious - is there a reason to do otherwise?

workflow/workflow.go
```diff
@@ -186,37 +189,27 @@ func makeRoots(nodes map[string]*node) ([]*node, error) {
...
-			var newRoots []*node
-			// TODO: Parallelize this loop
-			for _, n := range roots {
-				rs, err := a.run(ctx, n)
-				if err != nil {
-					return fmt.Errorf("cannot run process %q: %s", n.Process, err)
-				}
-				newRoots = append(newRoots, rs...)
-			}
+	for len(worklist) > 0 {
```
haydnKing: hmm, nervous that this is an infinite loop in some circumstances - e.g. a badly written element calling itself recursively - but users wouldn't be able to ctrl-c out...
I'm not immediately sure if we can do anything about it though

msackman-synthace:
```
for {
			part = NewComponent("dna_part")
}
```

We will never be able to stop the element writer "doing the wrong thing" so we probably shouldn't try (too hard)...?

workflow/workflow.go
msackman-synthace: I agree that what you suggest is easier to read, and I'll happily change it to that if you want. However, I have got used to the other form because of a couple of reasons - 1) it limits scope better, which can help avoid shadowing issues; 2) when these constructs get larger, it starts to look and feel quite like the error monad, eg:

```
if a, err := foo(); err != nil {
  return nil, err
} else if b, err := bar(a); err != nil {
  return nil, err
} else if c, err := baz(a,b); err != nil {
  return nil, err
} else {
  return c, nil
}
```

this both minimises scope, and makes it completely clear by quick visual inspection that all errors are being dealt with, and it minimises the loc.

But I do agree that in the simpler case of a single if-else, it can be overkill.

haydnKing: Fair enough, clearly sensible in that example and not too big a deal either way

workflow/workflow.go
haydnKing: true, I guess it's up to the environment to time out after a certain amount of time

msackman-synthace: (FWIW there are languages out there that are "Total" for which you can prove at compile time that they will terminate. I would be in favour of such things...)

PR #841, opened 2018-07-23
Refactor liquidhandler.Plan(...) in microArch/scheduler

haydnKing:
lh.Plan(...) is the key function in our liquidhandling planning scheme, but is quite hard to follow as much of the 'business logic' is hidden and it's not clear what parts of the LHRequest are actually set where.
This refactor should not change overall behaviour and focusses on the GetInputs() function, which estimates the volumes required for each liquid given the high level instruction, with the goal of making its behaviour and responsibilities clearer


microArch/scheduler/liquidhandling/getinputs.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	volsSupplied := make(map[string]wunit.Volume, len(volsRequired))
+	volsWanting := make(map[string]wunit.Volume, len(volsRequired))
+
+	for _, k := range allinputs {
+		// volSupplied: how much comes in
+		volSupplied := wunit.NewVolume(0.00, "ul")
```
msackman-synthace: You used 0.0 above. 0.00 looks ... surprising.

microArch/scheduler/liquidhandling/getinputs.go
```diff
@@ -0,0 +1,154 @@
...
+		for _, cmp := range inputSolutions[k] {
+			volSupplied.Add(cmp.Volume())
+		}
+		volsSupplied[k] = volSupplied
+
+		// volWanted: how much extra we wanted
+		volWanted := wunit.SubtractVolumes(volsRequired[k], volSupplied)
```
msackman-synthace: this api doesn't thrill me...

microArch/scheduler/liquidhandling/getinputs.go
```diff
@@ -0,0 +1,154 @@
...
+
+		if volWanted.GreaterThanFloat(0.0001) {
```
msackman-synthace: const?

microArch/scheduler/liquidhandling/getinputs.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/antha-lang/antha/microArch/logger"
+)
+
+//InputSolutions properties to do with the input Liquids required for the mix
+type InputSolutions struct {
+	Solutions       map[string][]*wtype.Liquid
+	Order           []string
+	VolumesSupplied map[string]wunit.Volume
+	VolumesRequired map[string]wunit.Volume
+	VolumesWanting  map[string]wunit.Volume
```
msackman-synthace: I'd quite like some comments on these fields - particularly noting invariants. For example, is Required always <= Supplied? Which fields can have keys that are not in other fields? etc

microArch/scheduler/liquidhandling/ichain.go
```diff
@@ -173,6 +173,7 @@ func (it *IChain) ProductIDs() string {
 	return s
 }
 
+//Flatten get the instruction IDs in order
 func (it *IChain) Flatten() []string {
```
msackman-synthace: This is more efficient and I feel cleaner expressed with an accumulator. Plus its current name is ... misleading.
```
func (it *IChain) FlattenValueIDs() []string {
  return it.flattenIDs(nil)
}

func (it *IChain) flattenValueIDs(acc []string) []string {
  if it == nil {
    return acc
  } else {
    for _, v := range it.Values {
      acc = append(acc, v.ID)
    }
    return it.Child.flattenValueIDs(acc)
  }
}
```

microArch/scheduler/liquidhandling/ichain.go
```diff
@@ -189,8 +190,23 @@ func (it *IChain) Flatten() []string {
...
+func (it *IChain) GetOrderedLHInstructions() []*wtype.LHInstruction {
+	var ret []*wtype.LHInstruction
+
+	if it == nil {
+		return ret
+	}
+
+	ret = append(ret, it.Values...)
+
+	ret = append(ret, it.Child.GetOrderedLHInstructions()...)
```
msackman-synthace: Yeah, I'd switch this to an accumulator form too.

microArch/scheduler/liquidhandling/ichain.go
```diff
@@ -825,3 +841,28 @@ func addNewNodesTo(ic *IChain, newNodes *IChain) *IChain {
 
 	return ic
 }
+
+func (self *IChain) getInstructionTypes() map[string]bool {
```
msackman-synthace: Hmm. This is where I think the IChain gets a bit confusing - many of the methods are acting on the chain from the current node to the tail. So having the method on the chain makes sense. But this one is clearly working only on the node. Maybe just rename to getNodeInstructionTypes ?

microArch/scheduler/liquidhandling/input_plate_setup.go
```diff
@@ -81,10 +81,10 @@ func (is InputSorter) Less(i, j int) bool {
 func input_plate_setup(ctx context.Context, request *LHRequest) (*LHRequest, error) {
 	st := sampletracker.GetSampleTracker()
 	// I think this might need moving too
-	input_platetypes := (*request).Input_platetypes
+	input_platetypes := request.Input_platetypes
```
msackman-synthace: could we drop the _ in these variable names at least?

microArch/scheduler/liquidhandling/layoutagent2.go
```diff
@@ -308,7 +308,7 @@ func get_and_complete_assignments(request *LHRequest, order []string, s []PlateC
 
 	// inconsistent plate types will be assigned randomly!
 	//	for k, v := range request.LHInstructions {
```
msackman-synthace: what is this commented-out code here for?

microArch/scheduler/liquidhandling/lhrequest.go
```diff
@@ -360,3 +338,43 @@ func (request *LHRequest) AllPlates() []*wtype.Plate {
...
+//previously been detrmined by setOutputOrder
+func (request *LHRequest) GetOrderedLHInstructions() ([]*wtype.LHInstruction, error) {
+	//if these aren't the same length then there was probably an issue with setOutputOrder
+	if len(request.OutputOrder) != len(request.LHInstructions) {
+		return nil, errors.Errorf("request OutputOrder has length %d but %d LHInstructions", len(request.OutputOrder), len(request.LHInstructions))
+	}
+
+	ret := make([]*wtype.LHInstruction, 0, len(request.OutputOrder))
+	for _, instructionID := range request.OutputOrder {
+		instruction, ok := request.LHInstructions[instructionID]
```
msackman-synthace: request and ret are visually quite similar. request => self ?

microArch/scheduler/liquidhandling/lhrequest.go
```diff
@@ -360,3 +338,43 @@ func (request *LHRequest) AllPlates() []*wtype.Plate {
...
+			return ret, errors.Errorf("request has invalid OutputOrder, no instruction with id %s", instructionID)
+		}
+		ret = append(ret, instruction)
+	}
+	return ret, nil
+}
+
+//GetUnorderedLHInstructions get a slice containing all the LHInstructions in
+//arbitrary order
+func (request *LHRequest) GetUnorderedLHInstructions() []*wtype.LHInstruction {
```
msackman-synthace: Why is this method needed? The map is already unordered - is there really any place where we make life easier by suppling this slice rather than the map?

microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -1407,21 +1179,21 @@ func removeDummyInstructions(rq *LHRequest) *LHRequest {
 		return rq
 	}
```
msackman-synthace: we have ins.IsMixInPlace - i.e. the method is on the instruction. But we don't have ins.IsDummy - please?

microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -1384,20 +1158,10 @@ func (lh *Liquidhandler) fix_post_names(rq *LHRequest) error {
...
-	}
-
-	return false
-}
-
 func removeDummyInstructions(rq *LHRequest) *LHRequest {
 	toRemove := make(map[string]bool, len(rq.LHInstructions))
 	for _, ins := range rq.LHInstructions {
-		if dummy(ins) {
+		if ins.IsDummy() {
```
msackman-synthace: I'm afraid the only code coverage we have in this method (both local tests, and via -runner) returns false for IsDummy(), so toRemove is always empty, so half of this method never gets run.

microArch/scheduler/liquidhandling/getinputs.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+//GetInputs
+func GetInputs(orderedInstructions []*wtype.LHInstruction, inputSolutions map[string][]*wtype.Liquid, carryVolume wunit.Volume) (*InputSolutions, error) {
```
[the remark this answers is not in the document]
m-i-sadowski: microArch/driver/liquidhandling/lhivector.go is your boy there (might want to move it to wtype?)

antha/anthalib/wtype/lhinstruction.go
```diff
@@ -55,7 +55,13 @@ type LHInstruction struct {
 }
 
 func (ins LHInstruction) String() string {
-	return fmt.Sprint(ins.InsType(), " G:", ins.Generation(), " ", ins.ID, " ", ComponentVector(ins.Components), " ", ins.PlateName, " ID(", ins.PlateID, ") ", ins.Welladdress, ": ", ins.ProductIDs())
+	ret := fmt.Sprintf("%s G: %d %s %v %s ID(%s) %s: %s", ins.InsType(), ins.Generation(), ins.ID, ComponentVector(ins.Components), ins.PlateName, ins.PlateID, ins.Welladdress, ins.ProductIDs())
```
jacksongabbard: In terms of maintainability, long lines like this are pretty painful. Suppose someone changes Welladdress to be a couple of fields because it makes sense architecturally. In this case, they'll have to edit in amongst this giant parameter list. Line breaks and spacing would help out massively:

```
ret := fmt.Sprintf(
  "%s G: %d %s %v %s ID(%s) %s: %s", 
  ins.InsType(), 
  ins.Generation(), 
  ins.ID, 
  ComponentVector(ins.Components),
  ins.PlateName, 
  ins.PlateID, 
  ins.Welladdress, 
  ins.ProductIDs()
)
```

microArch/scheduler/liquidhandling/getinputs.go
```diff
@@ -0,0 +1,152 @@
...
+	VolumesSupplied map[string]wunit.Volume    //the volumes of the solutions explicitly supplied to the protocol
+	VolumesRequired map[string]wunit.Volume    //the estimated volumes of solutions required to carry out the protocol
+	VolumesWanting  map[string]wunit.Volume    //the estimated shortfall between the supplied and required volumes which must be auto-allocated
+}
+
+//GetInputs calculate the volumes required for each input solution by looking
+//through the high level liquidhandling (LH) instructions.
+//inputSolutions gives the solutions explicitly provided to the protocol, and
+//is used to calculate the shortfall, i.e how much of each must be auto-allocated
+func GetInputs(orderedInstructions []*wtype.LHInstruction, inputSolutions map[string][]*wtype.Liquid, carryVolume wunit.Volume) (*InputSolutions, error) {
```
jacksongabbard: Same comment as above:

```
func GetInputs(
  orderedInstructions []*wtype.LHInstruction, 
  inputSolutions map[string][]*wtype.Liquid, 
  carryVolume wunit.Volume,
) (*InputSolutions, error) {
```

msackman-synthace:
So, having now discussed this in person, I feel I can put this on github:

> This means that reasoning about this history of these files is pretty much impenetrable unless you fine this particular PR in amongst the many commits and follow its evolution.

Which we do. There have been several cases in the last few weeks where I've used git blame to find out a commit id, and then xref'd that into github to find out the PR. That's a pretty quick job and not overly strenuous. The most important point is that we are not squashing any commits, which means we are not throwing away commit comments. So the commits that are 100% cosmetic (eg renamings) remain just that and will not cause future problems in git-blame or similar.

The fact is that in reviewing this code, there were changes that I wanted to see before this code could make it to master, that required larger scale cosmetic changes. Inevitably, that has made the complete composite change, devoid of its context, harder to review. Nevertheless you certainly can, in github, view arbitrary diffs if you want, as well as look at individual commits, which makes things simpler.

I feel a wider discussion of this maybe warranted, but that this may not be the ideal venue :)

haydnKing:
I'd like to get this merged in this week if poss. 
Have just run `antha-tester` on this branch and am getting some errors along the lines of `self OutputOrder has length 102 but 246 LHInstructions` - but am also getting these errors on master so I don't think it's related to this change

microArch/scheduler/liquidhandling/liquidhandler.go
haydnKing: this was done :)

microArch/scheduler/liquidhandling/liquidhandler.go
haydnKing: added TestEP3DummyInstruction to hit this code

PR #842, opened 2018-07-23
use the gilson tip chute by default

haydnKing:
Previously represented this as the gilsontipwaste (i.e. the on deck box) but this is considerably taller than the chute and so several collisions are predicted with the box which do not occur when using the chute


haydnKing:
have verified that this introduces no new errors with `antha-tester`

PR #844, opened 2018-07-23
cmd to update bundle file based on mapping details from metadata file

chr1sgrant:
Designed to facilitate automated upgrading of element bundles based on a defined mapping.

For example, running the command with the following metadata file against a test bundle would upgrade `Aliquot` to `Aliquot_Liquid` and  the output `Aliquots` would be updated to `NewLiquids` and the input `Solution` to `Liquid`.

```
{
  "defaults": {
    "ByRow": false,
    "NumberofAliquots": 8,
    "OutPlate": "SRWFB96_riser18",
    "Solution": "water",
    "VolumePerAliquot": "50ul"
  },
  "name": "Aliquot",
  "tags": [
    "MVL"
  ],
  "new-element-mapping-details":{
    "new-element-name": "Aliquot_Liquid",
    "new-parameter-mapping": {
      "Aliquots": "NewLiquids",
      "Solution": "Liquid"
    },
    "old-element-name": "Aliquot"
  }
}
``` 

Tested locally and working as expected.

### DIFF
--- a/cmd/antha/cmd/convert_all_bundle_parameters.go
+++ b/cmd/antha/cmd/convert_all_bundle_parameters.go
@@ -99,7 +99,7 @@ func convertAllBundles(cmd *cobra.Command, args []string) error {
 
 			if !c.Empty() {
 				for _, bundle := range bundles.Bundles {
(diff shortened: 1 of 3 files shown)


johansglock:
(Pending fixing of build errors of course)

PR #847, opened 2018-07-23
small tweaks to MicroAmp Optical 384 plate dimensions

CharlotteWard:
I've made some small changes to the 384 well PCR plates due to problems found with these during testing. Thanks :)


inventory/testinventory/serialize/make_plate_library.go
```diff
@@ -941,8 +941,8 @@ func make384wellplateAppliedBiosystems() *wtype.Plate {
 	numberOfRows := 16
 	numberOfColumns := 24
 
-	wellShape := "cylinder"
-	bottomtype := wtype.VWellBottom
+	wellShape := "CIRCLE"
```
m-i-sadowski: I think this needs to be reverted back to "cylinder" - AFAIK "CIRCLE" does not exist

PR #851, opened 2018-07-29
Fix/misc cleanings

msackman-synthace:
Minor tidying as I was reading through the frontend cmd stuff for antha.


jacksongabbard:
This PR is a mess. One thesis, it is not, sir.

msackman-synthace:
Ok, compile has been tested manually and is still working fine.

For pretty/run.go changes, `go install ./cmd/...` in the `antha-lang/elements` repo, and then `antha run --driver go://github.com/Synthace/PipetMaxDriver/server --bundle ~/Downloads/1x\ Aliquot.json` This is also working the same before and after.

For format, I have run `antha format -w /home/matthew/src/Go_external_1.10.3/src/repos.antha.com/antha-ninja/elements-westeros/Elements/Molecular_Biology_Starter/Construct_Assembly` (with and without -w). I've altered the input .an to be badly formatted, and checked that running the command corrects the formatting.

cmd/antha/cmd/format.go
```diff
@@ -57,64 +57,40 @@ func (w *walker) Walk(path string, fi os.FileInfo, err error) error {
...
-
-	out, err := format.Source(src)
-	if err != nil {
+	} else if out, err := format.Source(src); err != nil {
 		return err
-	}
-
-	if !w.Write {
+	} else if !w.WriteToFile {
 		os.Stdout.Write(out) // nolint
```
haydnKing: is there a reason to use `// nolint` here?

msackman-synthace: Well `Write` can return an error. I assume that's what the `// nolint` is for - really sod all we can do if we can't write to stdout. I'll check, hang on.

msackman-synthace:
Yup, `cmd/antha/cmd/format.go:67:18:warning: error return value not checked (os.Stdout.Write(out)) (errcheck)`

PR #853, opened 2018-07-30
Modified tip box offset by 0.1mm

m-i-sadowski:


m-i-sadowski:
Urgent 0.1mm fix based on consistency with pcrplate dimensions

PR #855, opened 2018-08-01
Overhaul tips well alignment check

haydnKing:
The planner needs to be able to determine whether a particular head can service a particular set of well coordinates simultaneously as one multichannel operation or whether they need to be split up into multiple operations.
Previously, this was done logically based on well addresses, with no checking of plate geometry or head capabilities. This led to situations where this check would succeed (and thus allow multichanneling) only for physical simulation (which checks the actual geometry) to reveal an error.
The new function `CanReach` takes *LHHead as receiver and introduces checks on plate geometry, and begins to allow for different head capabilities as we move towards proper support for independent multichannel, as well as introducing more thorough testing and commenting.

Nb. Since geometry based alignment test requires knowledge of well targets (i.e. how multiple channels should fit into a single well) a method for calculating well targets based on well and head geometry on the fly has been added since it cannot be guaranteed that well targets will have been set when `CanReach` is called. This method may end up replacing the current method (where targets are hidden away in the well's `Extra` field), but I believe that change would be out of scope for this PR


antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+	if head.GetParams().Orientation == LHVChannel {
+		channelOffset.Y = channelStep
+	} else {
+		channelOffset.X = channelStep
+	}
+
+	return channelOffset
+}
+
+//GetLargestChannelOffset get the largest possible distance between successive channels
```
m-i-sadowski: Hate to say it but the channels are a property of the Adaptor... these methods should be in LHAdaptor

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+
+	return true
+}
+
+//distanceOutsideSquare return a lower bound on how far position is outside the
+//square defined by the corners lowerLeft,topRight
+//where a is the bottom left and b is the top right corner
+//value will be negative if position is inside the square
+//z-value for all arguments is ignored
+func distanceOutsideSquare(lowerLeft, topRight, pos Coordinates) float64 {
```
msackman-synthace: You don't enforce that lowerLeft is both lower and lefter than topRight. If we had a type for 2D squares, then you could enforce that in construction?

antha/anthalib/wtype/lhhead.go
haydnKing: Yeah good call, I'll split out LHAdaptor into it's own file and move these there

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	var params *LHChannelParameter
+	var adaptor *LHAdaptor
+	if keepIDs {
+		params = head.Params.DupKeepIDs()
+		adaptor = head.Adaptor.DupKeepIDs()
+	} else {
+		params = head.Params.Dup()
+		adaptor = head.Adaptor.Dup()
+	}
```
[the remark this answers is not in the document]
haydnKing: Completely agree it would make more sense that way - thing is the current pattern of `Dup()`+`DupKeepIDs()`+`dup(keepIDs bool)` implemented on a bunch of things in wtype so would need changing everywhere for consistency (in a separate PR). I'd be in favour of doing this as it would remove a whole load of extra if...elses, what do you think @m-i-sadowski ?

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+	h := NewLHHead(head.Name, head.Manufacturer, params)
+	h.Adaptor = adaptor
+	h.TipLoading = head.TipLoading
+	return h
+}
+
+func (head *LHHead) Equal(rhs *LHHead) bool {
```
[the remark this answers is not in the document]
haydnKing: Yeah I thought is was odd, will investigate (copied from lhtypes.go)

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+func (head *LHHead) Dup() *LHHead {
+	return head.dup(false)
+}
+
+func (head *LHHead) DupKeepIDs() *LHHead {
+	return head.dup(true)
+}
+
+func (head *LHHead) dup(keepIDs bool) *LHHead {
+	if head == nil {
```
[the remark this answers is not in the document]
haydnKing: ^ ditto

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+
+//GetWellTargets get the offset from the center of the well for each channel that
+//can access the well simultaneously
+//returns nil if the well cannot fit multiple channels
+func (head *LHHead) GetWellTargets(well *LHWell) []Coordinates {
+	channelPositions := head.GetMostCompactChannelPositions()
+	channelRadius := 3.0 //should come from head
+
+	//total size of channels, including radius
+	channelSize := channelPositions[len(channelPositions)-1].Add(Coordinates{X: 2 * channelRadius, Y: 2 * channelRadius})
```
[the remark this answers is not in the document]
haydnKing: seems like a good idea, this also assumes that the first channel is at the origin, which is only the case by convention at the moment

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if !head.GetParams().Independent {
+		//non-independent heads can only use contiguous channels
+		seenNil := false
+		for _, crd := range coords {
+			if seenNil && crd != nil {
+				return false
+			}
+			seenNil = seenNil || crd == nil
+		}
```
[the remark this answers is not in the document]
haydnKing: Hmm, on reflection I think you're right. For example the gilson can't actually handle nil in the first channel, since that would require picking up a tip in a channel other than the first, but not the first, which is impossible for that device. However, in that case one could just remove leading nils and treat the first non-nil address as the address for channel 0...

Perhaps the simplest thing for non-independent heads would be to drop leading and trailing nils from the off, then any further nills must be within a non-nil section and thus not allowed

antha/anthalib/wtype/lhhead.go
haydnKing: types for 2d geometry is a thing we don't have in general... you're right, it's time

antha/anthalib/wtype/lhhead.go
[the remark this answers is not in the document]
msackman-synthace: Oh interesting (I had assumed eg gilson could do any contiguous block that includes _either_ the first or last head - making assumptions about how it can align the head "off the edge" of a tip box - apparently not the case). Does this mean that we should actually have an invariant on this method of addresses not starting or ending with nil(s) (err empty strings)? Or should we just trim any nils of either end as a first step?

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if wellSize := well.GetSize(); wellSize.X < channelSize.X || wellSize.Y < channelSize.Y {
+		return nil
+	}
+
+	center := Coordinates{}
+	for _, pos := range channelPositions {
+		center = center.Add(pos)
+	}
+	center = center.Divide(float64(len(channelPositions)))
```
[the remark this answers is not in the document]
haydnKing: Is that odd? I guess <vector type>.Scale(float) might be clearer, but I'd expect element wise division to be the exception (e.g. <vector>.ElementWiseDivide(<vector>)) rather than the default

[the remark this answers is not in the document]
msackman-synthace: Ok, maybe it's just me. In my head (possibly addled from Haskell abuse), all basic arithmetic operations are homogenous in type `(Num a) => (+) a -> a -> a`, same for `-`, `*`, `/` - so if our type is a vector, I would expect the basic ops would always be `Vec -> Vec -> Vec`. So deviation from that would be explicitly named, eg `DivideByScalar` because essentially it's a `map`. I can easily believe this is just a issue in my head...

antha/anthalib/wtype/lhhead.go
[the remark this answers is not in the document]
haydnKing: It's clearly something the hardware can do, but trilution insists on loading from the top. My feeling is that that particular requirement (i.e. addresses[0] != nil) is most likely a gilson quirk rather than a general property of non-independent liquid handlers, so I think trimming the nils (i.e. remapping channels to start at zero) is the more general thing to do since all the channels are the same in a non-independent machine... thoughts @m-i-sadowski ?

antha/anthalib/wtype/lhhead.go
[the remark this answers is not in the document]
haydnKing: Hmm, yeah I suspect it's something there are different conventions for... i guess since go forces you to specify what you want (unlike say `operator*` in c++, or numpy's funky broadcasting) it's at least explicit

antha/anthalib/wtype/lhhead.go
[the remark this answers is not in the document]
haydnKing: unused, removing

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,226 @@
...
+		//we're not particularly interested in the z dimension, so just use topreference
+		crd, ok := plate.WellCoordsToCoords(address, TopReference)
+		if !ok {
+			//can't address a well which doesn't exist
+			return false
+		}
+
+		//add well target offset if the well supports multiple tips at once
+		if wellTargets != nil {
+			crd = crd.Add(wellTargets[channel%len(wellTargets)])
```
haydnKing: note to self: replace channel with number of previous appearances of this address to support e.g. ["A1", "A1", "B1", "B1"]

antha/anthalib/wtype/lhhead.go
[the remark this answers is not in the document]
haydnKing: yup, the only time this is called is from LHProperties.Dup(), where a nil head is certainly an error

haydnKing:
think we can have another run at this, so long as Sid agrees with the discussion above

antha/anthalib/wtype/geometry.go
```diff
@@ -127,3 +135,102 @@ func (ps PointSet) CentreTo(c Coordinates) PointSet {
...
+}
+
+//String string representation of the coordinate
+func (self Coordinates2D) String() string {
+	return fmt.Sprintf("%.1fx%.1f mm", self.X, self.Y)
+}
+
+//Equals return true if the two coordinates are equal
+func (self Coordinates2D) Equals(rhs Coordinates2D) bool {
+	return self.X == rhs.X && self.Y == rhs.Y
```
msackman-synthace: you can actually just rely on structural equality here - literally `return self == rhs`

antha/anthalib/wtype/geometry.go
```diff
@@ -127,3 +135,102 @@ func (ps PointSet) CentreTo(c Coordinates) PointSet {
...
+
+//Abs return the L2 norm of the coordinate
+func (self Coordinates2D) Abs() float64 {
+	return math.Sqrt(self.X*self.X + self.Y*self.Y)
+}
+
+//a rectangle
+type Rectangle struct {
+	LowerLeft  Coordinates2D
+	UpperRight Coordinates2D
```
msackman-synthace: I would be tempted to make these fields private to the struct.

antha/anthalib/wtype/geometry.go
```diff
@@ -127,3 +135,102 @@ func (ps PointSet) CentreTo(c Coordinates) PointSet {
...
+
+//Height the height of the rectangle
+func (self Rectangle) Height() float64 {
+	return self.UpperRight.Y - self.LowerLeft.Y
+}
+
+//distanceOutside return a lower bound on how far position is outside the
+//rectangle.
+//The return value will be negative if position is inside the square
+func (self Rectangle) distanceOutside(pos Coordinates2D) float64 {
```
msackman-synthace: I think this is misnamed. Consider:

```
x      |
       |
       |
-------+--+
       |  |  <-box
       +--+
```

I think what your method will return is not the diagonal distance. Instead, it'll return the distance to the tangent of the closest infinitely-extended edge of the box. I'm sure there's a proper way of expressing that... I think there's nothing for it but to construct the vector directly, and call Abs on it. (or rename the method if the current behaviour is what you want)

antha/anthalib/wtype/geometry.go
haydnKing: neato

antha/anthalib/wtype/geometry.go
haydnKing: yeah I wasn't sure how to express that, hence "lower bound" in the comment. I think it'll actually give the distance to the furthest tangent - the purpose of the method is to allow for some small numerical error in the head position calculations, so it didn't seem worth the extra computational effort handling the corner cases separately.
I think I can express what I'm trying to do by first growing the rectangle by a small amount and then checking whether the point is inside...

msackman-synthace: Ahh ok - I don't think I'd figured out your intent. Yep, I think expanding the box by your positionAccuracy, and then checking for inclusion is the clearer approach.

antha/anthalib/wtype/lhadaptor.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return Coordinates2D{
+		X: self[len(self)-1].X - self[0].X + 2*channelRadius,
+		Y: self[len(self)-1].Y - self[0].Y + 2*channelRadius,
+	}
+}
+
+//Add return a new set of channel positions with the offset added
+func (self ChannelPositions) Add(rhs Coordinates2D) ChannelPositions {
+	ret := make([]Coordinates2D, len(self))
```
msackman-synthace: Neat feature of Go: `make(ChannelPositions, len(self))`

antha/anthalib/wtype/lhadaptor.go
```diff
@@ -0,0 +1,229 @@
...
+	for i, crd := range self {
+		ret[i] = crd.Add(rhs)
+	}
+	return ret
+}
+
+//Subtract return a new set of channel positions with the offset subtracted
+func (self ChannelPositions) Subtract(rhs Coordinates2D) ChannelPositions {
+	ret := make([]Coordinates2D, len(self))
```
msackman-synthace: same.

antha/anthalib/wtype/lhadaptor.go
```diff
@@ -0,0 +1,229 @@
...
+	return ret
+}
+
+//GetCenter return the mean of the channel coordinates
+func (self ChannelPositions) GetCenter() Coordinates2D {
+	ret := Coordinates2D{}
+	for _, crd := range self {
+		ret = ret.Add(crd)
+	}
+	return ret.Divide(float64(len(self)))
```
msackman-synthace: I'm confused - for eg a vertical head, are we saying the width is 0 (in which case this is all fine) or the width is 1 (or some non-0 value), in which case this isn't right? I think this is all originating in getSmallestChannelOffset, in which case width is 0....

msackman-synthace: However, as soon as we get to heads which are not a single line, this will go wrong. Maybe just note that in some comments somewhere (maybe around `type ChannelPositions`...)?

antha/anthalib/wtype/lhhead.go
```diff
@@ -0,0 +1,168 @@
...
+
+//CanReach return true if the head can service the given addresses in the given object
+//simultaneously.
+//addresses is a slice of well addresses which should be serviced by successive channels of
+//the head, eg. ["A1", "B1", "", "D1",] means channels 0, 1, and 3 should address wells
+//A1, B1, and D1 respectively and channels 2 and 4-7 should not be used.
+//Repeated addresses (e.g. ["A1", "A1", "A1"]) imply multiple tips per well, with
+//exact positioning for each tip calculated with LHHead.GetWellTargets().
+//Addresses are not reordered, and so ["A1", "B1"] != ["B1", "A1"].
+func (head *LHHead) CanReach(plate *LHPlate, addresses WellCoordSlice) bool {
```
msackman-synthace: Excellent docs :)

antha/anthalib/wtype/lhadaptor.go
msackman-synthace: Ok, here's the thing, once you use either .Add or .Subtract, .GetCenter is no longer correct to use. I think to solve this you're going to have to have ChannelPositions is both a rectangle with ll and ur, and then a list of coordinates with in that rectangle, probably relative to ll. And this level of complexity is now making me wonder whether your original version was much better...

antha/anthalib/wtype/lhadaptor.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+//GetTip Return the tip at channel_number, nil otherwise
+func (lha *LHAdaptor) GetTip(channel_number int) *LHTip {
+	return lha.Tips[channel_number]
+}
+
+//AddTip Load a tip to the specified channel
+func (lha *LHAdaptor) AddTip(channel_number int, tip *LHTip) {
+	lha.Tips[channel_number] = tip
```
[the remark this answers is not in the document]
haydnKing: Hmm, yeah I wondered about that, I need to check where this code is actually used and figure out what the correct behaviour actually is

antha/anthalib/wtype/geometry.go
```diff
@@ -127,3 +135,100 @@ func (ps PointSet) CentreTo(c Coordinates) PointSet {
...
+}
+
+//Expand return a new Rectangle with the same center point but whose width and
+//height is increased by the given positive amount
+func (self Rectangle) Expand(amount float64) Rectangle {
+	delta := Coordinates2D{X: amount / 2.0, Y: amount / 2.0}
+	return Rectangle{
+		lowerLeft:  self.lowerLeft.Subtract(delta),
+		upperRight: self.upperRight.Add(delta),
+	}
```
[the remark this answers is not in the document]
haydnKing: yeah I see what you mean - I think that's something that's often left to users to keep track of, at least in my experience of libraries like numPy etc

haydnKing:
fourth time's a charm :-)

antha/anthalib/wtype/lhadaptor.go
```diff
@@ -0,0 +1,213 @@
...
+		return self.GetSmallestChannelOffset()
+	}
+
+	//completely arbitrary for now since we don't report this
+	return self.GetSmallestChannelOffset().Multiply(2.0)
+}
+
+//GetMostCompactChannelPositions get the relative channel positions for the adaptor
+//in the most tightly bunched layout supported
+func (self *LHAdaptor) GetMostCompactChannelPositions() ChannelPositions {
```
msackman-synthace: Maybe rename this GetMostCompactChannelOffsets because it's built out of GetSmallestChannelOffset?

antha/anthalib/wtype/wellcoords.go
```diff
@@ -375,3 +375,22 @@ func HumanizeWellCoords(coords []WellCoords) string {
...
+	var start, end int
+	for i, wc := range self {
+		if !wc.IsZero() {
+			if start == end { //only true if we've not seen a nil
+				start = i
+			}
+			end = i + 1
+		}
+	}
+	if start < 0 {
```
msackman-synthace: impossible. Default value for int is 0. Do you mean `start == end`?

antha/anthalib/wtype/lhadaptor.go
haydnKing: Hmm you're right positions vs offset aren't clear, I think `%S/Get(Smallest|Largest)ChannelOffset/Get\1ChannelSpacing/` might be more clear though

PR #856, opened 2018-08-01
update linux installation instructions

chr1sgrant:


README.md
```diff
@@ -69,8 +69,8 @@ replacing the go binary with one that corresponds to your platform and
 replacing ``apt-get`` with your package manager.
 ```bash
 # Install go
-curl -O https://storage.googleapis.com/golang/go1.7.3.linux-amd64.tar.gz
-sudo tar -C /usr/local -xzf go1.7.3.linux-amd64.tar.gz
+curl -O https://storage.googleapis.com/golang/go1.10.3.linux-amd64.tar.gz
+sudo tar -C /usr/local -xzf go1.10.3.linux-amd64.tar.gz
```
msackman-synthace: Fine by me. But I wonder - do we have an official "these are the versions of Go we support" list? The text above this sorta suggests we just want the most recent version, in which case maybe to avoid having to update this document all the time, we should just say "if your linux distribution's version of Go is too old, then head to https://golang.org/dl/ to download the latest version" ?

chr1sgrant: I believe that currently the minimal requirement is go 1.10. You're suggestion sounds good to me; I'll just point them to the golang site to get the latest version.

PR #859, opened 2018-08-01
[ANTHA-1972] Expect Data

msackman-synthace:
Initial implementation. There is more to be done in elements, and in antha-runner.

However, I'd love feedback on this initial implementation attempt, hence creating this PR sooner rather than later.


msackman-synthace:
Not required now, so closing.

PR #871, opened 2018-08-16
add basic autoformatting to exported doe files

chr1sgrant:


antha/AnthaStandardLibrary/Packages/doe/export.go
```diff
@@ -230,5 +230,42 @@ func XLSXFileFromRuns(runs []Run, outputfilename string, dxorjmp string) (xlsxfi
...
+			w := float64(len(v)+1) * estimatedCharWidth
+			if w > sheet.Col(c).Width {
+				sheet.Col(c).Width = w
+			}
+		}
+	}
+}
+
+func autoFormatHeader(sheet *xlsx.Sheet) {
+	for i, row := range sheet.Rows {
```
haydnKing: how about
```
if len(sheet.Rows) == 0 {
  return
}
row := sheet.Rows[0]
...
```
seems clearer to me than iterating?

haydnKing:
So I think xlsx formatting is inherently a bit hacky* in go (at least for the moment), so I was wondering if we could move these functions to a specific package so we don't end up replicating this hack all over - (I've already done something similar in antha-runner) Maybe `antha/AnthaStandardLibrary/Packages/spreadsheet`?

*specifically having to guess an average character width and set column width manually, and know that borders have to be set to "thick"

chr1sgrant:
I've made these changes now and exported the AutoFormat function to the spreadsheet package.

PR #873, opened 2018-08-20
[ANTHA-2031] Improve the target.Inst interface

msackman-synthace:
Introduce an AppendDependsOn method to simplify a number of uses of the current SetDependsOn.

Local tests and antha-runner tests pass as before.

### DIFF
--- a/target/inst.go
+++ b/target/inst.go
@@ -220,9 +220,11 @@ type TimedWait struct {
 	Duration time.Duration
 }
 
+type Insts []Inst
+
 // SequentialOrder takes a slice of instructions and modifies them
 // in-place, resetting to sequential dependencies.
-func SequentialOrder(insts ...Inst) {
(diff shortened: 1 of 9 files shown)


target/inst.go
```diff
@@ -213,17 +220,16 @@ type TimedWait struct {
...
 
-// SequentialOrder takes a set of instructions with out any dependencies and
-// modifies them to follow sequential order
-func SequentialOrder(insts ...Inst) []Inst {
-	for idx, inst := range insts {
-		if idx == 0 {
-			continue
+// SequentialOrder takes a slice of instructions and modifies them
+// in-place, resetting to sequential dependencies.
+func SequentialOrder(insts ...Inst) {
```
haydnKing: you could
```
type Instructions []Inst
func (insts Instructions) SequentialOrder() {
    ...
}
```
maybe not worth it but i know you'd have raised it ;)

PR #877, opened 2018-08-23
[ANTHA-2038] Visit Robots.

msackman-synthace:
So quite a lot of boiler plate here to set it all up, but to see it in use, I've converted by lhtimer, and tip_box_setup, which at least to my eyes, now look much much nicer.

If you agree that this is worth pursuing, I will chase out all uses of GetParameter and then we should be able to delete all of those functions, which may well lead to a net reduction of code.


haydnKing:
Looks like a worthwhile endeavour to me. I had originally been put off by the amount of boiler plate necessary, but it's actually pretty neat and easy to follow... Think the real cleanness will come in examples where there are switches for different instruction types

msackman-synthace:
@haydnKing @m-i-sadowski Given the size of this, I think it's actually best to stop this one here. `GetParameters` still exists, but is only used now by LHPolicy stuff. Given how extraordinarily cross-cutting all that stuff is, this PR would easily double in size to support that - I'm sure how it could be done without reworking all the wtype.aparam stuff - would love to hear any suggestions!

microArch/driver/liquidhandling/compareinstructions.go
```diff
@@ -58,47 +30,42 @@ func mergeMovs(ris []RobotInstruction) []RobotInstruction {
...
+	strs := make([]string, len(es))
+	for i, err := range es {
+		strs[i] = err.Error()
+	}
+	return strings.Join(strs, "\n")
 }
 
-func CompareInstructionSets(setA, setB []RobotInstruction, opt ComparisonOpt) ComparisonResult {
-
+func CompareInstructionSets(setA, setB []RobotInstruction, comparators []RobotInstructionComparatorFunc) errorSlice {
```
haydnKing: I quite like the form
`func CompareInstructionSets(setA, setB []RobotInstruction, comparators ...RobotInstructionComparatorFunc) errorSlice`
so they you can do 
`CompareInstructionSets(a,b)` for some default behaviour or (compare all is this case), or
`CompareInstructionSets(a,b, comparison1, comparison2,...)` for specific behaviour
but since this is only used for testing as far as I can see it's probably not worth it

microArch/driver/liquidhandling/compareinstructions.go
```diff
@@ -240,168 +223,166 @@ func (mm MovMix) Check(lhpr wtype.LHPolicyRule) bool {
...
 
-		if !reflect.DeepEqual(p1, p2) {
-			errors = append(errors, fmt.Errorf("Instructions at index %d type %s parameter %s differ (%v %v)", index, ins1.Type().Name, prm, p1, p2))
+	var errs errorSlice
+	ric := &robotInstructionComparator{a: a, b: b}
+	for _, comp := range comparators {
+		ric.aFields = ric.aFields[:0]
+		ric.bFields = ric.bFields[:0]
+		ric.cur = ric.a
+		ric.cur.Visit(comp(ric))
```
haydnKing: could we do away with needing to set `ric.cur` by having 
`type RobotInstructionComparatorFunc func([]interface{}) *RobotInstructionBaseVisitor`
with e.g.
```
func CompareAspReference(out []interface{}) *RobotInstructionBaseVisitor {
	return &RobotInstructionBaseVisitor{
		HandleMovAsp: func(ins *MovAsp) { out = append(out, ins.Mov.Reference, ins.Asp.What) },
	}
}
```
and then just do `ric.a.Visit(comp(ric.aFields))`?

microArch/driver/liquidhandling/compareinstructions.go
```diff
@@ -240,168 +223,166 @@ func (mm MovMix) Check(lhpr wtype.LHPolicyRule) bool {
...
+func CompareVolumes(ric *robotInstructionComparator) *RobotInstructionBaseVisitor {
+	return &RobotInstructionBaseVisitor{
+		HandleAspirate: func(ins *AspirateInstruction) { ric.appendFieldValues(ins.Volume, ins.What) },
+		HandleMovAsp:   func(ins *MovAsp) { ric.appendFieldValues(ins.Asp.Volume, ins.Asp.What) },
+		HandleDispense: func(ins *DispenseInstruction) { ric.appendFieldValues(ins.Volume, ins.What) },
+		HandleMovDsp:   func(ins *MovDsp) { ric.appendFieldValues(ins.Dsp.Volume, ins.Dsp.What) },
+		HandleBlowout:  func(ins *BlowoutInstruction) { ric.appendFieldValues(ins.Volume, ins.What) },
+		HandleMovBlo:   func(ins *MovBlo) { ric.appendFieldValues(ins.Blo.Volume, ins.Blo.What) },
+		HandleMix:      func(ins *MixInstruction) { ric.appendFieldValues(ins.Volume, ins.What) },
+		HandleMovMix:   func(ins *MovMix) { ric.appendFieldValues(ins.Mix.Volume, ins.Mix.What) },
```
haydnKing: this is wonderful

microArch/driver/liquidhandling/compareinstructions.go
msackman-synthace: ahh, yes, nice idea, I think that would work. Provided we can get the late binding of out to still happen should be fine.

microArch/driver/liquidhandling/compareinstructions.go
msackman-synthace: I was *so* pleased when I got the code down to the same one-line-per-ins as it was before. But I think this is the case that shows your idea above won't work (passing in the `out []interface{}` - it needs to be a _different_ `out []interface{}` for each instruction type.

microArch/driver/liquidhandling/robotinstruction.go
```diff
@@ -90,7 +89,6 @@ var (
 	MBL = NewInstructionType("MBL", "MoveBlowout")
 	RAP = NewInstructionType("RAP", "RemoveAllPlates")
 	APT = NewInstructionType("APT", "AddPlateTo")
-	RPA = NewInstructionType("RPA", "RemovePlateAt")
```
msackman-synthace: Oh yes @haydnKing @m-i-sadowski I found no evidence RPA had any implementation or was used anywhere, so I removed it. Please shout if that was the wrong thing to do :)

microArch/driver/liquidhandling/compareinstructions.go
msackman-synthace: Ok, so after further thinking and discussion, this won't work, because there's no way to get the modified `out` out of the method. We could do it by passing in a `*[]interface{}` but that way gets confusing very quickly, so I don't think it's worth it.

msackman-synthace:
Ok, testing update: tests in antha and antha-tester work. I've run antha with makeTestBundle and checked that still works and appears to behave as before too. I just want to raised the code coverage a little further.

microArch/driver/liquidhandling/robotinstruction.go
haydnKing: I very much doubt this is used - it might be something we want to do someday (e.g. plate movement) - but unlikely to be useful in the immediate future I wouldn't think

m-i-sadowski: Yeah it's not currently used.

PR #878, opened 2018-08-24
Reimplement wunit library

haydnKing:
Goal of this PR is to improve the wunit library and to increase test coverage, while keeping the external API (i.e. calls made by elements) as similar as possible.

Prior to this PR, wunit has 46.8% test code coverage with several known issues. The PR increases test coverage to 80.1%, with a roughly 700 line reduction in non-test code and supports considerably more units.

I deleted the README... is it worth writing a new one?


antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
+package wunit
+
+import "math"
+
+type unitDef struct {
```
msackman-synthace: Hmm, I'm not sure the `Def` suffix adds much. I would be happier with either `unit` and `derivedUnit` or `baseUnit` and `derivedUnit`

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+	Name             string
+	Symbol           string
+	SISymbol         string //the canonincal form for the unit which can include a prefix, defaults to Symbol
+	Prefixes         []string
+	ExponentMinusOne int //such that default exponent is one
```
msackman-synthace: I think the potential confusion added by the MinusOne is not justified by the loc saved, sorry.

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+	},
+}
+
+type derivedUnitDef struct {
+	Name             string
+	Symbol           string
+	Prefixes         []string
+	ExponentMinusOne int //such that default exponent is one
+	TargetSymbol     string
+	NewUnitInTargets float64
```
msackman-synthace: I don't think `Unit` is quite the right word. It's more "scaling factor when converting to Target". Maybe `TargetScale` ?

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+	"Volume": {
+		{
+			BaseSymbol: "L",
+			BaseTarget: "l",
+			Prefixes:   allPrefixes,
+		},
+	},
+	"Length": {
+		{
+			BaseSymbol: "\"",
```
msackman-synthace: you could use backticks and avoid the backslash in this case.

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+		{
+			Name:             "per hour",
+			Symbol:           "/h",
+			TargetSymbol:     "/s",
+			NewUnitInTargets: 1 / 3600.0,
+		},
+	},
+}
+
+type aliasDef struct {
```
msackman-synthace: I'm trying to figure out: "is it better to have a distinct mechanism for doing aliases, as you have here, or should the other unit types have a slice of Symbols to express aliases, rather than a single Symbol?". What do you think?

antha/anthalib/wunit/unit.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+// Unit everything we need to know about a unit to support it
+type Unit struct {
+	name       string   //common name for the unit
+	symbol     string   //symbol of the unit
+	base       string   //the SI unit for this dimension, or derived one if there isn't one
+	prefix     SIPrefix //the SI prefix which is applied to the symbol
+	multiplier float64  //value to multiply by to convert "symbol"s to "base", e.g. 60 for min (SI unit = s)
+	exponent   int      //the exponent for the prefix, 1 unless the prefix is inclided in a power, e.g. exponent=2 for "m^2"
```
msackman-synthace: 1. spelling - inclided
1. you write "exponent for the prefix", but then give an example where the exponent is of the unit, not the prefix. It feels like it's of the unit to me. Also, can this ever be negative? I presume it can - eg acceleration m.s^-2 ?

antha/anthalib/wunit/unit.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/pkg/errors"
+	"math"
+)
+
+// Unit everything we need to know about a unit to support it
+type Unit struct {
```
msackman-synthace: Oh... why do we need a different set of structs to declare the definitions in then?

antha/anthalib/wunit/unit.go
```diff
@@ -0,0 +1,140 @@
...
+	symbol     string   //symbol of the unit
+	base       string   //the SI unit for this dimension, or derived one if there isn't one
+	prefix     SIPrefix //the SI prefix which is applied to the symbol
+	multiplier float64  //value to multiply by to convert "symbol"s to "base", e.g. 60 for min (SI unit = s)
+	exponent   int      //the exponent for the prefix, 1 unless the prefix is inclided in a power, e.g. exponent=2 for "m^2"
+}
+
+// MarshalJSON marshal the unit as a JSON string
+func (self *Unit) MarshalJSON() ([]byte, error) {
+	fmt.Println("MarshalJSON")
```
msackman-synthace: wot.

antha/anthalib/wunit/wunit.go
```diff
@@ -58,6 +58,9 @@ type PrefixedUnit interface {
 	BaseSISymbol() string
 	// returns conversion factor from *this* unit to the other
 	ConvertTo(pu PrefixedUnit) float64
+	// CompatibleWith returns true iff the PrefixedUnit can be converted to rhs
+	// if false then calling ConvertTo will cause a panic()
+	CompatibleWith(PrefixedUnit) bool
 }
```
msackman-synthace: Ok, I'm not sure I see the advantage of having a separate BaseUnit and PrefixedUnit - a BaseUnit is just a PrefixedUnit with a Prefix of 1. However, I can well imagine that fixing this would extend the scope of this PR substantially, so please push back on this if so.

antha/anthalib/wunit/unit.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return self.BaseSIConversionFactor() / pu.BaseSIConversionFactor()
+}
+
+// CompatibleWith returns true if the units can be converted to the supplied units.
+// If this function returns false then calling ConvertTo with the same units will
+// case a panic
+func (self *Unit) CompatibleWith(pu PrefixedUnit) bool {
+	return self.base == pu.BaseSIUnit()
+}
```
msackman-synthace: Ok, so you have an API here with 2 methods - compatibleWith and convertTo, and to be safe, they always have to be called in that order one after another. So I would suggest it's much better to have a single API - convertTo that can return an error if they're not compatible. That way, no one can forget to call compatibleWith => a simpler and smaller API.

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+			if err := reg.DeclareDerivedUnit(mType, du.Name, du.Symbol, du.Prefixes, du.ExponentMinusOne+1, du.TargetSymbol, du.NewUnitInTargets); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+var systemDerivedUnits = derivedUnitDefs{
+	"Concentration": {
```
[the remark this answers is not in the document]
msackman-synthace: @chr1sgrant I'd like to understand that last point better.

In my head, the system should work by internally always converting everything to the simplest possible unit, and then compatibility checking is trivial. Then, when you want to render a value as a string, you simply specify at that point in what units you want to render the value - i.e. you give the user the ability to specify how they want to see the value.

So when you write:
> The users have also professed a dislike of the system auto converting units so preferable if we can avoid that

I would like to better understand that.

target/mixer/marshal_plate_test.go
```diff
@@ -44,7 +44,7 @@ func TestMarshalPlateCSV(t *testing.T) {
 				`
 pcrplate_with_cooler,Input_plate_1,LiquidType,Vol,Vol Unit,Conc,Conc Unit
 A1,water,water,50,ul,0,g/l
-A4,tea,water,50,ul,10,mM/l
```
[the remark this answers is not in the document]
haydnKing: `mM` is still fully supported, it's now an alias for `mMol/l` which is the correct SI form

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+		{
+			Name:             "grams per mililitre",
```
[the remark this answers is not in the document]
haydnKing: fully agree this would be a nicer way of doing things, but we'd need a little more machinery to specify that "concentration" = (<any valid mass or molar amount> "/" <any valid volume>) or "X" or "v/v" or "w/v", so I figured it would be best to keep things simple to start with, particularly since we don't support "g.l^-1"

antha/anthalib/wunit/conversion.go
```diff
@@ -25,134 +25,73 @@ package wunit
...
-		fmt.Println("starting conc SI ", stockConc.SIValue(), " and target mass SI: ", targetmass.SIValue())
-
-	} else if stockConc.Unit().BaseSIUnit() == "kg/l" && targetmass.Unit().BaseSIUnit() == "kg" {
-		v = NewVolume(float64((targetmass.SIValue()/stockConc.SIValue())*1000000), "ul")
-		fmt.Println("starting conc SI ", stockConc.SIValue(), " and target mass SI: ", targetmass.SIValue())
-
-	} else if stockConc.Unit().BaseSIUnit() == "g/l" && targetmass.Unit().BaseSIUnit() == "g" {
-		v = NewVolume(float64((targetmass.SIValue()/stockConc.SIValue())*1000000), "ul")
-		fmt.Println("starting conc SI ", stockConc.SIValue(), " and target mass SI: ", targetmass.SIValue())
+	if gramsPerULitre, err := GetGlobalUnitRegistry().GetUnit("g/ul"); err != nil {
```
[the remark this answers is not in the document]
haydnKing: hmm, I'm happy to change this, but in general I don't think we should expect measurements to be in any particular units and should always use ConvertToString if we really want them to be in something specific

antha/anthalib/wunit/sort.go
```diff
@@ -104,22 +104,7 @@ func (cs concentrationSet) Swap(i, j int) {
...
-	if iSIValue == 0.0 {
-		return iSIValue < jSIValue
-	}
-
-	if jSIValue == 0.0 {
-		return iSIValue < jSIValue
-	}
-
-	return false
+	return cs[i].SIValue() < cs[j].SIValue()
```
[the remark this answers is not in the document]
haydnKing: I believe it is - `SortConcentrations` first checks that each concentration has the same base SI unit (it raises an error if there's a mixture of kg/l, Mol/l, X, for example), they SIValue() converts the value into the SI unit

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+			Symbol:           "m^2",
+			Prefixes:         allPrefixes,
+			ExponentMinusOne: 1,
+		},
+	},
+	"Temperature": {
+		{
+			Name:   "celsius",
+			Symbol: "℃",
+		},
```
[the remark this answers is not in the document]
haydnKing: nope, there are a bunch of aliases for this so it should function as before - "C", "˚C", and "°C" are accepted, see systemunits.go +482

antha/anthalib/wunit/systemunits.go
```diff
@@ -0,0 +1,520 @@
...
+			NewUnitInTargets: 3600.0,
+		},
+		{
+			Name:             "days",
+			Symbol:           "days",
+			TargetSymbol:     "s",
+			NewUnitInTargets: 86400.0,
+		},
+	},
+	"Angle": {
```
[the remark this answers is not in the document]
haydnKing: no, but we support can support it now, e.g.
```
reg := wunit.GetGlobalUnitRegistry()
ang, _ := reg.NewMeasurement(5.0, "degrees")
fmt.Println(ang)
inRad, _ := reg.NewMeasurement(ang.ConvertToString("rad"), "rad")
fmt.Println(inRad)
```
leads to
```
5 °
0.08726646259971647 rad
```

antha/anthalib/wunit/wunit.go
```diff
@@ -174,12 +170,19 @@ func (cm *ConcreteMeasurement) ConvertTo(p PrefixedUnit) float64 {
 	return cm.Unit().ConvertTo(p) * cm.RawValue()
 }
 
+// ConvertToString return the measurement to the units specified by symbol.
+// panic()s if symbol is not a known symbol
 func (cm *ConcreteMeasurement) ConvertToString(s string) float64 {
```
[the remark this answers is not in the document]
haydnKing: Yeah, agreed this is highly confusing but I think something that element writers are currently used to so I was reluctant to change it in this PR. Definitely something to change when we more completely change the wunit interface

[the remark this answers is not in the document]
msackman-synthace: Ok, can we at least introduce an alias then and mark ConvertToString as deprecated in comments and point them to the new method?

antha/anthalib/wunit/siprefix.go
```diff
@@ -26,20 +26,54 @@ import (
...
+}
+
+var longPrefixNames = map[string]string{
+	"y": "yocto",
+	"z": "zepto",
+	"a": "atto",
+	"f": "femto",
+	"p": "pico",
+	"n": "nano",
+	"u": "micro",
```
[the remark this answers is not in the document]
haydnKing: I chickend out and added a strings.Replace(unit, "μ", "u") in UnitRegistry - a bit manual but I think μ is enough of a special case that it's fairly OK

antha/anthalib/wunit/siprefix.go
```diff
@@ -26,20 +26,54 @@ import (
...
+	"h": "hecto",
+	"k": "kilo",
+	"M": "mega",
+	"G": "giga",
+	"T": "tera",
+	"P": "peta",
+	"E": "exa",
+	"Z": "zetta",
+	"Y": "yotta",
+}
```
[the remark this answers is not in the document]
haydnKing: 1) nope, it could be generated from the longPrefixNames or prefices
2) there was - the two character "da" would have broken the previous implementation, but I think it might be OK now
3) yup, agree that would be clearer
4) yeah true, or would a unit test be more efficient

I think it's fair to say `siunit` didn't get as much attention in this refactor as the rest of the lib, so probably worth revisiting

antha/anthalib/wunit/systemunits.go
msackman-synthace: Pffft. Switch to pointers. `map[string]*Unit` and then `for _, s := range unit.Symbols { unitsBySymbol[s] = unit }`

antha/anthalib/wunit/unit.go
haydnKing: 1. :sadface:
2. Hmm, I've not expressed this well, but what I mean is that `mm^2` is actually `(mm)^2`, not `m(m^2)`, so `1 mm^2` is `10^-6 m^2`

Undoubtedly there are examples where it would be negative, though since we currently only support one prefix (by which I mean when we add `m.s^-2` we automatically add `mm.s^-2`, but `m.ms^-2` would have to be added as a derived unit), I think we're unlikely to run into that case any time soon

antha/anthalib/wunit/systemunits.go
msackman-synthace: And even if you don't switch to pointers, I don't really think there's going to be a performance hit or memory explosion...

antha/anthalib/wunit/unit.go
haydnKing: oops.
(or was this wrt. the existence of the func?)

antha/anthalib/wunit/unit.go
haydnKing: they're shorthand for the calls to UnitRegistry.DeclareX, which define a bunch of units with different prefixes

antha/anthalib/wunit/unit.go
msackman-synthace: the `fmt.Println`.

antha/anthalib/wunit/wunit.go
haydnKing: Yeah, I really wanted to consolidate these, but feared it might get out of hand

antha/anthalib/wunit/unit.go
haydnKing: Yeah, totally agreed. Personally I think we should have
`func (self *Measurement) ConvertTo(pu PrefixedUnit) (*Measurement, error)`
so we 
1. can return an error instead of panic()
2. reduce confusion of what ConvertTo is actually doing
3. reduce chances of users operating directly with unitless floats

antha/anthalib/wunit/systemunits.go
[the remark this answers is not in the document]
chr1sgrant: I would consider the user's choice of dimension when specifying the measurement value to be an implicit choice really. I certainly wouldn't have thought you'd want to add parameters for the user to choose which unit should be returned in the vast majority of cases. 

I agree it would be reasonable to do so in the element if the context would make conversion to a particular unit sensible (say based on convention...i.e. ng/ul or fM for DNA concentrations or U/ml for enzyme activity). In terms of default behaviour though, probably the most reasonable thing would be to just round to the nearest prefix which would be greater than 1?

So if specifying 1ng/ul the system can auto convert to 1* 10e9 g/ul for some internal calculations,  and then convert back to 1ng/ul rather than returning 1* 10e9 g/ul later on.

antha/anthalib/wunit/siprefix.go
```diff
@@ -62,41 +106,6 @@ func SIPrefixBySymbol(symbol string) SIPrefix {
...
-func PrefixDiv(x string, y string) string {
-	// divide x by y, what do you get?
-
-	l1 := RoundInt(math.Log10(prefices[x].Value))
-	l2 := RoundInt(math.Log10(prefices[y].Value))
-	return ReverseLookupPrefix(l1 - l2)
-}
-
 // make the prefix structure
 func MakePrefices() map[string]SIPrefix {
```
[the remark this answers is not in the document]
chr1sgrant: I Agree with this; much clearer.

antha/anthalib/wunit/unit.go
chr1sgrant: I would prefer returning an error output and handling what to do upon error in elements

antha/anthalib/wunit/wdimension.go
```diff
@@ -202,74 +182,57 @@ func DivideConcentration(v Concentration, factor float64) (newconc Concentration
...
-
-	if math.IsNaN(factor) {
-		err = errors.Errorf("NaN value found dividing concentrations %s and %s", conc1.ToString(), conc2.ToString())
-		return
+	if conc2.IsZero() {
+		return 0, errors.Errorf("while dividing concentrations %s and %s: cannot divide by zero", conc1, conc2)
 	}
-
-	return factor, nil
+	return conc1.RawValue() / conc2.ConvertTo(conc1.Unit()), nil
```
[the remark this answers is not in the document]
chr1sgrant: It may be better to return the Inf or Nan in both cases rather than converting to zero, or returning a more specific error type ErrNan, ErrInf

antha/anthalib/wunit/wdimension.go
```diff
@@ -202,74 +182,57 @@ func DivideConcentration(v Concentration, factor float64) (newconc Concentration
...
 }
 
 // AddConcentrations adds a variable number of concentrations from an original concentration.
 // An error is returned if the concentration units are incompatible.
-func AddConcentrations(concs ...Concentration) (newconc Concentration, err error) {
+func AddConcentrations(concs ...Concentration) (Concentration, error) {
 
 	if len(concs) == 0 {
-		err = errors.Errorf("Array of concentrations empty, nil value returned")
+		return NewConcentration(0.0, "g/ul"), nil
```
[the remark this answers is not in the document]
haydnKing: Hmm, yeah a bit random I agree. There is a bit of a problem here in that you can't add the return value to a concentration in Mol/l, for example :/

antha/anthalib/wunit/wunit.go
msackman-synthace: I'm happy for this future step to just be recorded as a jira ticket so that it doesn't get lost.

antha/anthalib/wunit/conversion.go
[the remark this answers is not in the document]
haydnKing: oh, actually this is converted to g/ul because the final division is (g) / (g/ul) -> (ul)

antha/anthalib/wunit/unit.go
msackman-synthace: We're going to have to cope with making breaking changes to elements in general. So given that currently we don't have any ability to track via versions dependencies between elements and anthalib, we can't hope to do anything clever. So at a minimum, we need to figure out "how do we know if making this change breaks elements?", and if it does, then I suspect we'll have to introduce the new api regardless, and mark the old as deprecated, and come back to the issue once we have proper versioning in place.

antha/anthalib/wunit/systemunits.go
haydnKing: sigh, you're right

antha/anthalib/wunit/siprefix.go
```diff
@@ -22,114 +22,54 @@
...
-		prefix := SIPrefix{string(rune), math.Pow10(exponent)}
-		pref_map[string(rune)] = prefix
-		exponent += 1
-	}
-
-	exponent = 3
-
-	pfcs = "kMGTPEZY"
+var (
+	Yocto = newPrefix("y", "yocto", 1e-24)
```
msackman-synthace: I would probably add a comment here saying something like "keep these in order from smallest to largest as we rely on Go evaluating these in declaration order - see https://golang.org/ref/spec#Package_initialization "

antha/anthalib/wunit/unit.go
```diff
@@ -99,18 +99,18 @@ func (self *Unit) BaseSISymbol() string {
 
 // ConvertTo get the conversion factor between this unit and pu.
 // This function will call panic() if pu is not compatible with this unit, see CompatibleWith
```
msackman-synthace: comment needs correcting.

antha/anthalib/wunit/unit.go
```diff
@@ -99,18 +99,18 @@ func (self *Unit) BaseSISymbol() string {
...
+		return 0.0, errors.Errorf("cannot convert units: base units for %s and %s do not match: %s != %s", self.PrefixedSymbol(), pu.PrefixedSymbol(), self.base, pu.BaseSISymbol())
 	}
 
-	return self.BaseSIConversionFactor() / pu.BaseSIConversionFactor()
+	return self.BaseSIConversionFactor() / pu.BaseSIConversionFactor(), nil
 }
 
 // CompatibleWith returns true if the units can be converted to the supplied units.
 // If this function returns false then calling ConvertTo with the same units will
 // case a panic
```
msackman-synthace: comment needs correcting 2.

antha/anthalib/wunit/siprefix.go
haydnKing: I added comment, though I don't think we currently *rely* on SIPrefix being in order, but it would be a bit daft if it wasn't

haydnKing:
I think this is ready for another attempt to merge
@chr1sgrant one side effect of this change is that units which were previously automatically converted (e.g. we used to convert "ng/ul" into "mg/l") are no longer converted automatically. I've already fixed the test bundles so that Liquid names match up, can you think of any other potential effects of this change?

msackman-synthace:
Ok, I don't like ConvertTo and ConvertToByString - the naming annoys me and gives no clue that their return types are so different. Ultimately, I think ConvertToByString will completely go away - that functionality is probably not necessary long term, but I accept now isn't the moment to address this.

To my mind, ConvertTo -> AsFloatWithUnit and then ConvertToByString -> RemeasureWithUnit is clearer.

antha/anthalib/wunit/unitfromstring_test.go
```diff
@@ -35,213 +34,194 @@ type unitFromStringTest struct {
...
+			ComponentNameOnly: "(1X)",
+		},
+	}.Run(t)
 }
 
-func TestParseVolume(t *testing.T) {
+type volTest struct {
+	VolString string
+	Volume    Volume
+	Error     bool
```
msackman-synthace: hi!

antha/anthalib/wunit/unitregistry.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+}
+
+// DeclareUnit add a unit to the registry, as well as corresponding entries for valid prefixes
+// If validPrefixes is zero length, only the base symbol will be added
+func (self *UnitRegistry) DeclareUnit(measurementType, name, baseSymbol, SISymbol string, validPrefixes []SIPrefix, exponent int) error {
+	self.mutex.Lock()
+	defer self.mutex.Unlock()
+	if len(validPrefixes) == 0 {
+		validPrefixes = []SIPrefix{None} //if no prefixes specified, only add default
```
msackman-synthace: I don't really like this. I would prefer a simpler rule, like "None is always added".

antha/anthalib/wunit/unitregistry.go
```diff
@@ -0,0 +1,227 @@
...
+// This is expected to be used when there are multiple convensions for writing a unit, for example
+//   reg.DeclareAlias("volume", "L", "l", SIPrefixes)
+// will lead to all units with "L" (e.g. "uL", "mL") being converted to "l" (e.g. "ul", "ml", etc).
+// If validPrefixes is zero length, only the base symbol will be added
+// Note there is no value scaling, for that see DeclareDerivedUnit
+func (self *UnitRegistry) DeclareAlias(measurementType, baseSymbol, baseTarget string, validPrefixes []SIPrefix) error {
+	self.mutex.Lock()
+	defer self.mutex.Unlock()
+	if len(validPrefixes) == 0 {
+		validPrefixes = []SIPrefix{None} //if no prefixes specified, only add default
```
msackman-synthace: same in here.

antha/anthalib/wunit/wdimension_test.go
```diff
@@ -0,0 +1,700 @@
...
+
+type TestFn func(*testing.T, Measurement)
+
+type NewMeasurementTest struct {
+	Value            float64
+	Unit             string
+	ExpectedSIValue  float64
+	ExpectedBaseUnit string
+	ExpectedPrefix   string
+	ExpectError      bool
```
msackman-synthace: Amazing! ... except looking at how this is used in Run, this is more ExpectPanic, not ExpectError

antha/anthalib/wunit/wdimension.go
```diff
@@ -24,40 +24,38 @@ package wunit
...
+	if value, err := GetGlobalUnitRegistry().NewMeasurement(value, unit); err != nil {
+		panic(err)
+	} else {
+		return value
+	}
+}
 
 // length
 type Length struct {
 	*ConcreteMeasurement
```
msackman-synthace: I have to say, this feels deeply dodgy to me - I don't think I've ever seen a case where you really want to store a pointer within value types. It means if you do

```
a := ZeroLength()
b := a
b.Incr(3) // let's pretend this works
```

At this point you would still have `a == b`.

I can believe that maybe there are good reasons why it's been done this way. But it looks spooky to me...
I think if this changed, you could then drop stuff like CopyVolume...

antha/anthalib/wunit/wdimension.go
```diff
@@ -156,23 +139,16 @@ func DivideVolume(v Volume, factor float64) (newvolume Volume) {
...
-		return
+	if vol2.IsZero() {
+		return 0, errors.Errorf("while dividing volume %s by %s: cannot divide by zero", vol1, vol2)
 	}
 
-	return factor, nil
+	return vol1.SIValue() / vol2.SIValue(), nil
 }
 
 func (c Concentration) Dup() Concentration {
```
msackman-synthace: There's Copy... and Dup?!! :(

antha/anthalib/wunit/wunit.go
```diff
@@ -71,26 +57,24 @@ type Measurement interface {
...
+	InUnit(p PrefixedUnit) (Measurement, error)
+	// InStringUnit wrapper for InUnit which fetches the unit from the global UnitRegistry
+	InStringUnit(symbol string) (Measurement, error)
+	// ConvertToString deprecated, please use ConvertTo or InStringUnit
 	ConvertToString(s string) float64
-	// add to this measurement
+	// AddTo add to this measurement
+	AddTo(m Measurement) error
+	// SubtractFrom subtract from this measurement
+	SubtractFrom(m Measurement) error
```
msackman-synthace: I thought we'd discussed this and felt that `IncrBy` and `DecrBy` were the way to go. Did that get vetoed?

antha/anthalib/wunit/wdimension.go
msackman-synthace: (totally happy for this to be out of scope for this PR)

antha/anthalib/wunit/README
```diff
@@ -1,47 +0,0 @@
-Units and measurements:
```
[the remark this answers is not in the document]
msackman-synthace: (having looked further, this is actually plenty, so resolving this)

antha/anthalib/wunit/unitregistry.go
haydnKing: I had this previously and something went awry with it... but I can't remember what... I'll have a look

antha/anthalib/wunit/wdimension.go
haydnKing: Yeah, I agree this is pretty spooky - I've also been confused several times that our static types don't appear to be pointers, but actually are... I think this is out of scope though, happy to open a ticket to review static types

antha/anthalib/wunit/wunit.go
haydnKing: Oh, yeah, we did agree that. My monkey brain must have forgotten :( Shouldn't be too big of a change

antha/anthalib/wunit/sort_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+import (
+	"testing"
+)
+
+// SortTest this is used to test both SortConcentrations and SortVolumes with only
+// a little code duplication
+type SortTest struct {
+	Name  string
+	Input []interface{}
```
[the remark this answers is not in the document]
haydnKing: have done this because I agree this is cleaner, but it did cost some loc with validate (some of which I think could be recouped with some advanced use of reflection, but meh), and also in the struct initialisation... probably acceptable though

antha/anthalib/wunit/wdimension.go
haydnKing: Yeah... looks like Volume.Dup and CopyVolume are used with about the same frequency, while CopyConcentration is more common the Concentration.Dup
Could fix this, but it's gonna touch a whole load of extra files :-/

msackman-synthace: Maybe mark one as deprecated in a comment then to try and force people to migrate?

antha/anthalib/wunit/sort_test.go
```diff
@@ -19,200 +19,224 @@ func (self *SortTest) unexpectedError(err error) bool {
...
-		test.RunVolumes(t)
-	}
-}
 func TestSortConcentrations(t *testing.T) {
-	SortTests{
+	ConcentrationSortTests{
 		{
-			Name: "simple test",
-			Input: []interface{}{
+			SortTest: SortTest{
```
msackman-synthace: Damn, I didn't think this was necessary - I thought you could just inline it all, but having played around, I can see you do have to do all this too. That is very annoying - slightly regretting asking you for this change now. Sorry.

antha/anthalib/wunit/wdimension.go
haydnKing: Ok, marking the Dups as deprecated, then at least it's the same form as AddVolumes etc

antha/anthalib/wunit/sort_test.go
haydnKing: np, if i'd realised it was gonna be quite so long i'd have pushed back, but I think it's a better pattern in general

PR #882, opened 2018-08-31
Calculate liquidlevel during physical simulation

haydnKing:
This fixes part (1) of ANTHA-2010, which is caused by physical simulation not taking liquid level into account during moves.

It's pretty small, but there's some awkwardness around use of wunit.Volume (see antha/anthalib/wtype/plate.go +1182) that I'm not too happy with, but don't see a better way of doing it without reopening the wunit can of worms


antha/anthalib/wtype/plate.go
```diff
@@ -1171,10 +1171,16 @@ func (self *Plate) WellCoordsToCoords(wc WellCoords, r WellReference) (Coordinat
 
 	var z float64
 	//return the bottom as a lower bound for liquid reference
-	if r == BottomReference || r == LiquidReference {
+	if r == BottomReference {
 		z = child.GetPosition().Z + child.Bottomh
 	} else if r == TopReference {
 		z = child.GetPosition().Z + child.GetSize().Z
+	} else if r == LiquidReference {
+		if volInUl, err := child.CurrentVolume().InStringUnit("ul"); err != nil {
```
msackman-synthace: It's fairly idiomatic in Go to have "Must" variants of methods (the regexp package does this for example). In those variants, they don't return an error and panic if the internal error is non nil. I wonder if it's worth adding those if this is going to be a common pattern for volume/ul combinations?

antha/anthalib/wtype/lhwell.go
```diff
@@ -603,6 +604,26 @@ func (lhw *LHWell) GetLiquidLevelModel() wutil.Func1Prm {
...
+	if f := lhw.GetLiquidLevelModel(); f == nil {
+		return 0.0
+	} else if quad, ok := f.(*wutil.Quadratic); !ok {
+		return 0.0
+	} else if quad.C > vol { //no negative or imaginary heights
+		return 0.0
+	} else if quad.A == 0 { //linear model
+		return (vol - quad.C) / quad.B
+	} else {
+		return (-quad.B + math.Sqrt(quad.B*quad.B-4.0*quad.A*(quad.C-vol))) / (2.0 * quad.A)
```
msackman-synthace: I think I'm going to need walking through the maths on this as I don't understand how this relates to the stuff in the comment...

haydnKing: should be the result of
`v = Ah^2 + Bh + C` solve for `h`
but actually could be simpler as for the models gilson support `C=0`

antha/anthalib/wtype/plate.go
haydnKing: Yes, I think that's a good idea

haydnKing: Have added [ANTHA-2057](https://synthace.atlassian.net/browse/ANTHA-2057) to cover this

antha/anthalib/wtype/lhwell.go
msackman-synthace: Gotcha, and lgtm.

msackman-synthace:
All looks good to me... but we have zero test coverage of these changes. :(

haydnKing:
ah good point, it shouldn't be hard to add one in

microArch/simulator/liquidhandling/simulator_testcode_test.go
```diff
@@ -1338,16 +1347,20 @@ func (self *SimulatorTest) run(t *testing.T) {
...
 	//check assertions
 	if self.Assertions != nil {
 		for _, a := range self.Assertions {
-			(*a)(self.Name, t, vlh)
+			(*a)(t, vlh)
 		}
 	}
 }
+
+type SimulatorTests []SimulatorTest
```
msackman-synthace: When you start doing these yourself, my work is done :1st_place_medal:

microArch/simulator/liquidhandling/simulator_testcode_test.go
```diff
@@ -201,6 +202,9 @@ func makeLHWell(p *LHWellParams) *wtype.LHWell {
 		p.bottomh,
 		p.dunit)
 	w.Crds = p.crds
+	if p.liquidLevelModel != nil {
+		w.SetLiquidLevelModel(*p.liquidLevelModel)
```
msackman-synthace: Dislike. I would make SetLLModel(nil) into a noop.

haydnKing: Hmm, ideally this would involve changing wutil.Quadratic to have pointer receivers, which would likely involve changes to the drivers :-/ could make SetLiquidLevelModel accept *wutil.Quadratic rather that wutil.Func1Param?

msackman-synthace: woah, how did I miss that?! In which case we're already dealing with a pointer. So then you just need to drop the * and the nil-test in _test, and in lhwell.go you need to cope with m being nil. Or have I managed to miss something again?

haydnKing: ah fine - somehow forgot that the fact that Quadratic has non-pointer methods isn't a problem

PR #883, opened 2018-09-04
Add copy constructors for concrete measurement types in wunit

haydnKing:
Copy constructors are required until each type embeds ConcreteUnit by value rather than pointer, but they weren't implemented for each type.

These are implemented on the package to match with `AddVolumes` etc and existing functions.

Tests have also been updated to call copy constructors during TestNew<X>

### DIFF
--- a/antha/anthalib/wunit/wdimension.go
+++ b/antha/anthalib/wunit/wdimension.go
@@ -58,6 +58,14 @@ func NewLength(v float64, unit string) Length {
 	return Length{NewTypedMeasurement("Length", v, unit)}
 }
 
+// CopyLength duplicate the Length
+func CopyLength(v Length) Length {
+	if isNil(v.ConcreteMeasurement) {
+		return Length{}
+	}
+	return NewLength(v.RawValue(), v.Unit().PrefixedSymbol())
+}
+
 // area
 type Area struct {
 	*ConcreteMeasurement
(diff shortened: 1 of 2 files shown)


haydnKing:
(secondary goal is to discover the minimal PR for which @msackman-synthace can request changes)

msackman-synthace:
This is so painful. For values you really should be able to write:

```
a := b
```

and then a should be a copy of b. Adding 200 loc to get around the fact that we have a mistake of a pointer inside a value is not something that excites me. Can you give some estimate of how painful/disruptive it's going to be to fix the underlying problem? -  i.e. we should be using (I think) ConcreteMeasurement and not *ConcreteMeasurement for embedding.

haydnKing:
Somewhat hard to predict I think - shouldn't be too hard to make the change, but it's possible we have code which relies on the strange behaviour. I have a ticket for making that change anyway, I'll spend a bit of time on it and see whether it starts to snowball

msackman-synthace:
For code to rely on it would require stuff like:

```
a := b
a.Incr(100)
```

and then use b assuming it has been incremented. Given how non-obvious that behaviour is (because neither a or b have a pointer type), if we can find and remove that code, that would be a win.

But I totally agree to see how much it starts to snowball...

haydnKing:
closing since we can live without this until the time of the glorious refactor(tm)

PR #884, opened 2018-09-04
Mars data/support truncated data files

chr1sgrant:


antha/AnthaStandardLibrary/Packages/platereader/dataset/parse/mars.go
```diff
@@ -167,16 +164,25 @@ func parseWellData(xlsxBinary []byte, sheet int, headerrows int) (welldatamap ma
...
-
-		if cellstr == "A" {
-			wellrowstart = i
-			break
+		cellstr := firstCellOfRow.String()
+		// at most double letter rows will be supported as first row
+		if _, err := secondCellOfRow.Int(); len(cellstr) < 3 && err == nil {
+			var invalid bool
+			for _, char := range cellstr {
+				if !unicode.IsLetter(char) {
```
msackman-synthace: This is confusing to me - I had made assumptions that the cellstr probably starts at "0" or "1" and may eventually work its way through letters... I'm making those assumptions because there are no comments to guide me. This line invalidates my assumptions as it says you're looking for only letters. Could you write a longer section of prose at the top of this method that explains how the file is laid out and thus what we're looking for - what the expected invariants are for the first cell of row etc?

antha/AnthaStandardLibrary/Packages/spreadsheet/spreadsheet.go
```diff
@@ -117,24 +117,24 @@ func SheetToCSV(sheet *xlsx.Sheet) (records [][]string) {
 // An error is returned if the column aor row number specified is beyond the range available in the sheet.
```
msackman-synthace: typo - `aor`

antha/AnthaStandardLibrary/Packages/spreadsheet/spreadsheet.go
```diff
@@ -117,24 +117,24 @@ func SheetToCSV(sheet *xlsx.Sheet) (records [][]string) {
...
 // Counting starts from zero. i.e. the cell at the first row  and first column would be called by
 // cell, err := GetDataFromRowCol(sheet, 0,0)
-func GetDataFromRowCol(sheet *xlsx.Sheet, col int, row int) (cell *xlsx.Cell, err error) {
-	if col >= len(sheet.Rows) || col < 0 {
-		var cols []int
+func GetDataFromRowCol(sheet *xlsx.Sheet, rowInSheet int, columnInSheet int) (cell *xlsx.Cell, err error) {
```
msackman-synthace: I would be tempted to change the type to uint64 for rowInSheet and columnInSheet - it is (I assume) nonsensical to use negative numbers.

antha/AnthaStandardLibrary/Packages/spreadsheet/spreadsheet.go
```diff
@@ -117,24 +117,24 @@ func SheetToCSV(sheet *xlsx.Sheet) (records [][]string) {
...
+	if rowInSheet >= len(sheet.Rows) || rowInSheet < 0 {
+		var rowIndices []int
 		for key := range sheet.Rows {
```
msackman-synthace: This is quite a lot of work to go to for an error message. Meaningful error messages are great, but I wonder if just a simple "available rows from 0 to %d", len(sheet.Rows) would be just as useful?

antha/AnthaStandardLibrary/Packages/spreadsheet/spreadsheet.go
```diff
@@ -117,24 +117,24 @@ func SheetToCSV(sheet *xlsx.Sheet) (records [][]string) {
...
+			rowIndices = append(rowIndices, key)
 		}
-		return nil, fmt.Errorf("column %d not found in xlsx sheet. Found these: %v", col, cols)
+		return nil, fmt.Errorf("row %d not found in xlsx sheet. Found these: %v", rowInSheet, rowIndices)
 	}
 
-	column := sheet.Rows[col]
-	if row >= len(column.Cells) || row < 0 {
-		var rows []int
+	column := sheet.Rows[rowInSheet]
```
msackman-synthace: Surely, if you have a slice of rows, and you index it by a row index, what you get back is a row - i.e. a slice of columns. Definitely not a `column`. It really reads to me like this should be named `row`...

antha/AnthaStandardLibrary/Packages/spreadsheet/spreadsheet.go
```diff
@@ -117,24 +117,24 @@ func SheetToCSV(sheet *xlsx.Sheet) (records [][]string) {
...
+	if columnInSheet >= len(column.Cells) || columnInSheet < 0 {
+		var columnIndices []int
 		for key := range column.Cells {
-			rows = append(rows, key)
+			columnIndices = append(columnIndices, key)
```
msackman-synthace: same comment as above applies - I suspect just knowing the length may be just as useful.

PR #885, opened 2018-09-06
[ANTHA-2073] Simplify QPCR support.

msackman-synthace:
I'm going to need some help figuring out how to test this best - there is 0% code coverage from both local tests and antha-tester tests. Seems like some integration tests in the latter would be useful. Any ideas?

### DIFF
--- a/target/qpcrdevice/qpcrdevice.go
+++ b/target/qpcrdevice/qpcrdevice.go
@@ -21,9 +21,8 @@ type QPCRDevice struct {
 var _ target.Device = (*QPCRDevice)(nil)
 
 // NewQPCRDevice returns a new QPCR Machine
-func NewQPCRDevice() *QPCRDevice {
-	ret := &QPCRDevice{}
-	return ret
+func New() *QPCRDevice {
+	return &QPCRDevice{}
 }
 
 // CanCompile implements a Device
--- a/target/qpcrdevice/qpcrdevice.go
+++ b/target/qpcrdevice/qpcrdevice.go
@@ -28,8 +28,9 @@ func NewQPCRDevice() *QPCRDevice {
 
 // CanCompile implements a Device
 func (a *QPCRDevice) CanCompile(req ast.Request) bool {
-	can := ast.Request{}
-	can.Selector = append(can.Selector, target.DriverSelectorV1QPCRDevice)
+	can := ast.Request{
+		Selector: []ast.NameValue{target.DriverSelectorV1QPCRDevice},
+	}
 	return can.Contains(req)
 }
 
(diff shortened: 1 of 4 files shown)


ddn0:
Thoughts:
- highest level: create a package that has a bunch of integration tests that call `execute.Run()` with constructed device targets and dummy elements that have been registered
- next highest level (probably better): above but at the level of `codegen.Compile()`

target/qpcrdevice/qpcrdevice.go
```diff
@@ -38,8 +39,11 @@ func (a *QPCRDevice) MoveCost(from target.Device) int64 {
 	return 0
 }
 
-// singleCallForInstruction generates a device call, based on a single qPCR insruction.
-func singleCallForInstruction(inst *ast.QPCRInstruction) (driver.Call, error) {
+func (dev *QPCRDevice) callForInstruction(inst *ast.QPCRInstruction) (target.Inst, error) {
```
haydnKing: somewhat strange function name since it now returns an `Inst`, no?

msackman-synthace: Indeeed. I'll rename it to `instForInstruction`... ;)

msackman-synthace:
Ok, here's what I've discovered. The bundle elements-westeros/DemoWorkflows/qPCR/qPCR_test_workflow_large_sampleSet.bundle.json does contain a qPCR_run component. This does generate the right ast instruction. However, antha-tester appears to run only with the human enabled, so the qPCR compiler never gets hit. So we get out:

```
== Round  5:
    * [QPCR] QPCR request, definition 2017-11-28 DoE3_1 48hr run.edt, barcode abcd
```

If in `auto.go` I hack in the qpcrdevice, then this does get called and appears to work:

```
== Round  5:
    * [prm] Ensure that the experiment file 2017-11-28 DoE3_1 48hr run.edt is configured, then put the plate (abcd) into qPCR device. Check that the driver software is running. Once ready, accept to start the qPCR analysis.
== Round  6:
    * [run] Perform qPCR Analysis
```

So I am happy enough the code can be made to get hit.

However, I'm concerned that antha-tester is apparently working with only the human device enabled - surely this would prevent vast amounts of code from being hit? I don't understand really how any of the robot instructions for the LH stuff is being hit - surely we should be defaulting back to having a human pipette everything in this case?

haydnKing:
hmm, I think antha-tester manually creates a PM driver when it runs the bundle, see: https://github.com/Synthace/antha-runner/blob/master/cmd/antha-tester/main.go#L72
Perhaps the qPCR device could also be added?

msackman-synthace:
Yeah, so I've made that change in the linked PR in -runner: https://github.com/Synthace/antha-runner/pull/340/files#diff-d694522c2f9df13d3cc1fd5249167281R104

Just rerunning the whole -tester suite to see what's getting hit. I wonder if we're accidentally selecting the human in preference to the qpcrdevice in some cases?

ddn0:
Could just force no human for antha-tester

msackman-synthace:
Ok, so because I don't want this PR to grow beyond its scope, I'm going to merge this and address the other issues around testing in a separate PR. The fact is we do have a suitable test bundle and it does hit this code ... if a bunch of other issues are fixed.

PR #886, opened 2018-09-06
[Antha-1978] Add structs to allow reporting of head movement behaviour

haydnKing:
We need to allow instruction generators (e.g. PipetMaxDriver) to better report the movement behaviour of head assemblies within liquidhandlers, specifically so that:
1. we know the range of allowable velocities that can be requested (this will allow us to catch more issues at simulation time)
2. we can accurately calculate how long movement actions will take in order to improve our timing estimates

This change achieves these with MovementBehaviour, which defines the movement in each of the X, Y, and Z directions, as well as the allowable parameter ranges. It also defines the order in which movements are carried out - e.g. for gilson X and Y are moved together, the Z is moved separately - and also some actions which can occur before or after the movement, such as moving to the safety height.

Currently we only support linear acceleration and deceleration as a movement profile -- I have no evidence that this is actually the case for gilson though it seems like a likely first order estimate -- but it should be easy to add new movement profiles along with associated serialisation and deserialisation.

This change introduces just under 450 non test loc, though will hopefully lead to simplifying some tests in `microArch/simulator/liquidhandling` and the complete removal of `microArch/scheduler/liquidhandling/simulator`

### DIFF
--- a/antha/anthalib/wtype/movementprofile.go
+++ b/antha/anthalib/wtype/movementprofile.go
@@ -1,18 +1,61 @@
 package wtype
 
 import (
+	"fmt"
 	"github.com/pkg/errors"
 	"math"
+	"strings"
 
 	"github.com/antha-lang/antha/antha/anthalib/wunit"
 )
 
 // MovementBehaviour describe movement behaviour in three dimensions
 type MovementBehaviour struct {
-	Profiles   map[Dimension]LinearMovementProfile
+	Profiles   []LinearMovementProfile
 	Order      [][]Dimension //Order in which movement is carried out - dimensions in the same list are carried out simultaneously
 	BeforeMove []MovementAction
 	AfterMove  []MovementAction
+}
+
+// NewMovementBehaviour builds a new movement behaviour
+func NewMovementBehaviour(xProfile, yProfile, zProfile LinearMovementProfile, order [][]Dimension, beforeActions, afterActions []MovementAction) (*MovementBehaviour, error) {
+	if xProfile == nil || yProfile == nil || zProfile == nil {
+		return nil, errors.New("cannot use nil movement profile")
+	} else if err := assertOrderValid(order); err != nil {
+		return nil, err
+	} else {
+		return &MovementBehaviour{
+			Profiles:   []LinearMovementProfile{xProfile, yProfile, zProfile},
+			Order:      order,
+			BeforeMove: beforeActions,
+			AfterMove:  afterActions,
+		}, nil
+	}
+}
+
+// assertOrderValid each dimension should feature exactly once
+func assertOrderValid(order [][]Dimension) error {
+	errs := []string{}
+	seen := map[Dimension]int{}
+	for _, group := range order {
+		for _, dim := range group {
+			seen[dim] = seen[dim] + 1
+		}
+	}
+
+	for _, dim := range Dimensions {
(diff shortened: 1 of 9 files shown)


antha/anthalib/wtype/movementprofile.go
```diff
@@ -0,0 +1,381 @@
...
+			Order:      order,
+			BeforeMove: beforeActions,
+			AfterMove:  afterActions,
+		}, nil
+	}
+}
+
+// assertOrderValid each dimension must feature exactly once
+func assertOrderValid(order [][]Dimension) error {
+	errs := []string{}
```
msackman-synthace: there's now `utils.ErrorSlice`

antha/anthalib/wtype/movementprofile.go
```diff
@@ -0,0 +1,381 @@
...
+			Profiles:   MovementProfile{xProfile, yProfile, zProfile},
+			Order:      order,
+			BeforeMove: beforeActions,
+			AfterMove:  afterActions,
+		}, nil
+	}
+}
+
+// assertOrderValid each dimension must feature exactly once
+func assertOrderValid(order [][]Dimension) error {
```
msackman-synthace: (minor) In my head (possibly addled by a few years of java abuse), assert is something that blows up if it fails. This feels more like `check` rather than `assert`

antha/anthalib/wtype/movementprofile.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	return ret
+}
+
+// MovementAction an action that is carried out before or after a move operation
+type MovementAction interface {
+	// Duration calculate the time taken by the action and the final position of the head assembly,
+	// given the current location of the head assembly and movement behaviour
+	Duration(location Coordinates, behaviour *MovementBehaviour) (wunit.Time, Coordinates)
```
msackman-synthace: it's not clear to me whether `location` is the starting position of this action, or the desired ending position.

antha/anthalib/wtype/movementprofile.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	*self = make(MovementActions, 0, len(actions))
+	for _, action := range actions {
+		m := map[string]*json.RawMessage{}
+		if err := json.Unmarshal(*action, &m); err != nil {
+			return errors.WithMessage(err, "unmarshalling MovementAction")
+		}
+
+		var t movementActionType
```
msackman-synthace: This is ... not ideal. Who has come up with this use of json and can we change it?

antha/anthalib/wtype/movementprofile.go
```diff
@@ -0,0 +1,381 @@
...
+
+// GenericAction an action which happens in constant time at the begining or end of a move
+// for example locking or unlocking a head
+type GenericAction struct {
+	TimeTaken wunit.Time
+}
+
+// NewGenericAction create a new generic action, asserting that the time taken is positive
+func NewGenericAction(time wunit.Time) (*GenericAction, error) {
+	if !(time.IsZero() || time.IsPositive()) {
```
msackman-synthace: there's no time.IsNegative()?

haydnKing:
yeah... this got more complicated than necessary for the time being. Will keep the branch around in case it becomes useful, but will close for now

PR #887, opened 2018-09-06
rm unused/unmaintained packages from AnthaStandardLibrary/Packages

haydnKing:


m-i-sadowski:
This code is clearly anomalous as part of antha-lang since it's not really part of the language, rather it falls into the bucket of client code and would be better living alongside the elements that consume it (at least some of which exist, perhaps on branches) 

I would definitely be in favour of rationalising the stuff in AnthaStandardLibrary in favour of maintaining a good separation between what should change at the pace of its client elements and what should be included as part of the language itself. That would mean migrating most (probably not all) of the code in here out to relevant other packages, most likely as part of element dependencies. 

I think as it stands then removing at least a good part of this will break some element code somewhere out there so it's not appropriate to delete it wholesale without migrating it out.

PR #888, opened 2018-09-07
[ANTHA-2076] File Semantics

msackman-synthace:


msackman-synthace:
Oh yes, sure enough, with this in, a full antha-tester run actually runs the qpcr bundles and hits the qpcr driver now.

execute/executeutil/unmarshal.go
```diff
@@ -1,73 +1,88 @@
...
-		if err := unmarshal(opt.BundleData, &bundle); err != nil {
-			return nil, err
+	containers := make([]Bundle, len(paths))
+	for idx, path := range paths {
+		if bites, err := ioutil.ReadFile(path); err != nil {
+			errs = append(errs, fmt.Errorf("Error when reading file %s: %v", path, err))
+		} else {
+			container := &containers[idx]
+			if err := json.Unmarshal(bites, container); err != nil {
+				errs = append(errs, fmt.Errorf("Error when parsing content of %s: %v", path, err))
```
ddn0: style was to have error messages begin lower case; if we have changed that, please ignore.

execute/executeutil/unmarshal.go
```diff
@@ -1,73 +1,88 @@
...
-	var bundle Bundle
-	var expected workflowtest.TestOpt
+	bundles := make(map[string]*Bundle, len(paths))
+	params := make(map[string]*execute.RawParams, len(paths))
+	workflows := make(map[string]*workflow.Desc, len(paths))
 
-	if len(opt.BundleData) != 0 {
-		if err := unmarshal(opt.BundleData, &bundle); err != nil {
-			return nil, err
+	containers := make([]Bundle, len(paths))
```
ddn0: technically you could leave new `Bundle`s to be stack allocated within the loop body itself. Here you intend this as a performance optimization by preallocating? If so, I guess keep, but I wanted to be sure why.

execute/executeutil/unmarshal.go
```diff
@@ -1,73 +1,88 @@
...
-		param.Config = bundle.Config
-		param.Parameters = bundle.Parameters
-		expected = bundle.TestOpt
-	} else {
-		if err := unmarshal(opt.WorkflowData, &desc); err != nil {
+	}
+	return bundles, workflows, params, errs
+}
+
+func UnmarshalSingle(bundlePath, workflowPath, paramsPath string) (*Bundle, error) {
```
ddn0: preference to godoc all public functions

microArch/driver/liquidhandling/compareinstructions.go
```diff
@@ -30,24 +30,14 @@ func mergeMovs(ris []RobotInstruction) []RobotInstruction {
...
-func (es errorSlice) Error() string {
-	strs := make([]string, len(es))
-	for i, err := range es {
-		strs[i] = err.Error()
-	}
-	return strings.Join(strs, "\n")
-}
-
-func CompareInstructionSets(setA, setB []RobotInstruction, comparators ...RobotInstructionComparatorFunc) errorSlice {
+func CompareInstructionSets(setA, setB []RobotInstruction, comparators ...RobotInstructionComparatorFunc) utils.ErrorSlice {
```
ddn0: godoc

execute/executeutil/test.go
```diff
@@ -110,82 +83,75 @@ func FindTestInputs(basePath string) ([]*TestInput, error) {
...
 		}
 
-		bundle, err := Unmarshal(UnmarshalOpt{
-			BundleData:   bdata,
-			ParamsData:   pdata,
-			WorkflowData: wdata,
-		})
-		if err != nil {
-			return nil, fmt.Errorf("error parsing %q: %s", strings.Join(input.Paths(), ","), err)
+		if len(errs) > 0 {
```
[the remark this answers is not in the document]
msackman-synthace: I don't think so - the issue is that we're now looking at *all* json files, which means I'm pretty sure we're going to start picking up files that legit don't have a bundle/workflow/params in them. I don't yet see a way of distinguishing between fatal and non-fatal errors in these cases - maybe you have some ideas?

execute/executeutil/unmarshal.go
```diff
@@ -1,73 +1,88 @@
...
 	execute.RawParams
 	workflowtest.TestOpt
 }
 
-// Unmarshal parses parameters and workflow.
-func Unmarshal(opt UnmarshalOpt) (*Bundle, error) {
-	if len(opt.BundleData) != 0 && (len(opt.ParamsData) != 0 || len(opt.WorkflowData) != 0) {
-		return nil, errBundleWithParams
-	}
+// Attempts to read and parse all the file paths supplied,
```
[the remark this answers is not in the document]
msackman-synthace: Oh, I'm happy to obey that if you want - I know it is "go style". Personally I think those rules are ... questionable - but I do not want to rock the boat - happy to keep to the standard if that is something we've been doing in the past.

execute/executeutil/unmarshal.go
msackman-synthace: I probably shouldn't have written it this way - apologies. Yes, it's a mechanism to reduce calls to malloc, but this is totally non perf critical code. So it probably has no value here other than to confuse the reader - I'll change.

microArch/driver/liquidhandling/compareinstructions.go
ddn0: If you touch the line, you need to bring it up to spec ;)

execute/executeutil/unmarshal.go
ddn0: that was what I've been working under, but I'm not sure how much code has been written under either assumption. For me consistency is first. Actual decision is less important.

execute/executeutil/unmarshal.go
```diff
@@ -1,73 +1,88 @@
...
+	bundles := make(map[string]*Bundle, len(paths))
+	params := make(map[string]*execute.RawParams, len(paths))
+	workflows := make(map[string]*workflow.Desc, len(paths))
 
-	if len(opt.BundleData) != 0 {
-		if err := unmarshal(opt.BundleData, &bundle); err != nil {
-			return nil, err
+	containers := make([]Bundle, len(paths))
+	for idx, path := range paths {
+		if bites, err := ioutil.ReadFile(path); err != nil {
```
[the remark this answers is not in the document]
ddn0: then it's ok with me. (this is why I tend to use `bs`. `bites` sounds like we are vampires.

execute/executeutil/unmarshal.go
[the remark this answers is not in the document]
ddn0: Yeah, I personally prefer your style (from my old javadoc days), but I think for overall codebase consistency, I'd prefer gdoc without exceptions rather than cherry-picking. Easier to explain to new starters. Longer discussion welcome (outside this PR ;) ).

execute/executeutil/test.go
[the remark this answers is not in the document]
ddn0: Good point. Ignore me then.

PR #892, opened 2018-09-12
[ANTHA-2100] Remove awaitInst and associated parts.

msackman-synthace:
Local tests still pass. Just waiting for antha-tester to finish.


msackman-synthace:
antha-tester also still passes as before

PR #894, opened 2018-09-13
Turns out SName isn't used either.

msackman-synthace:
local tests pass. Just waiting on antha-tester...


msackman-synthace:
Yup, they pass too.

PR #899, opened 2018-09-17
Unify LHInstruction.Add{Product,Result} into .AddOutput.

msackman-synthace:
Not sure what to do about AddComponent: I'm happy to change it to AddInput, but there are 3 elements that I can find that use it (Transformation, Setup_Input_Plate, Transformation_Multi). Just fearful of changing a public API that may be used by 3rd party elements - how do we normally handle this case?

NB: this PR is entirely find and replace, with the exception of lhinstruction.go itself.

Local tests pass. Just running through antha-tester now...


haydnKing:
this LGTM assuming antha-runner passes, though probably better for sid to have a look too

PR #900, opened 2018-09-17
[Antha-2010] Fix Liquid Level Follow SuckInstruction generation

haydnKing:
when aspirating with LLF, the head moves downwards to track the liquid level. Unlike when dispensing, when aspirating it is therefore possible to collide with the bottom of the well.

This PR implements LLF aspiration in a way which prevents this. First, the minimum liquid level for which LLF can be used is calculated as `ASPZOFFSET + LLFBELOWSURFACE`. SuckInstructions which don't go below this level are carried out entirely using LLF and ones where the starting well volume is already below this value aspirate from well bottom as normal without LLF. Otherwise, the SuckInstruction generates two Aspirate instructions - one using LLF until the minimum level is reached, and the next from the well bottom without LLF to aspirate the remaining volume.

This PR depends on: https://github.com/antha-lang/antha/pull/891 
and https://github.com/Synthace/PipetMaxDriver/pull/52 and https://github.com/Synthace/PipetMaxDriver/pull/53


antha/anthalib/wtype/lhwell.go
```diff
@@ -624,6 +624,19 @@ func (lhw *LHWell) GetLiquidLevel(volume wunit.Volume) float64 {
 	}
 }
 
+// GetLiquidVolume estimate the volume of liquid for a given liquid level.
+// performs the inverse of GetLiquidLevel
```
msackman-synthace: Minor, but "inverse" is potentially confusing - insofar as one might think the inverse of a `get` is a `set`. So you may just want to spell out "height -> volume, rather than volume -> height"

antha/anthalib/wunit/wdimension.go
```diff
@@ -462,6 +462,50 @@ func NewVelocity(v float64, unit string) Velocity {
 	return Velocity{NewTypedMeasurement("Velocity", v, unit)}
 }
 
+// Velocity3D struct composed of lengths in three axes
+type Velocity3D struct {
+	X, Y, Z Velocity
```
msackman-synthace: totally minor, just an alternative design to think about:

`type Velocity3D [3]Velocity`

and then you index by the axis (need to double check you wouldn't have to cast... yup, all good https://play.golang.org/p/TfFmkmmWZPF)

antha/anthalib/wunit/wdimension.go
```diff
@@ -462,6 +462,50 @@ func NewVelocity(v float64, unit string) Velocity {
...
+	case XAxis:
+		self.X = v
+	case YAxis:
+		self.Y = v
+	case ZAxis:
+		self.Z = v
+	default:
+		panic("unknown axis")
+	}
+}
```
msackman-synthace: yeah, these 2 methods might go away with the above design.

inventory/testinventory/platelibrary.go
```diff
@@ -4748,7 +4748,9 @@ var plateBytes = []byte(`
   "WellXStart": 14.53,
   "WellYStart": 11.24,
   "WellZStart": 3.7,
-  "Extra": {}
+  "Extra": {
+   "ll_model": "{\"Quadratic\":\"\",\"A\":0,\"B\":25,\"C\":0}"
```
msackman-synthace: Yikes really? We really have to be string -> string in here, and not string -> json.RawMessage?

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -1746,6 +1744,103 @@ func (ins *SuckInstruction) GetParameter(name InstructionParameter) interface{}
...
+	ret := make(map[string]wunit.Volume, ins.Multi)
+	initial := ins.getInitialVolumesByWell()
+	removed := ins.getRemovedVolumesByWell()
+	for _, well := range ins.WellFrom {
+		ret[well] = wunit.SubtractVolumes(initial[well], removed[well])
+	}
+	return ret
+}
+
+func (ins *SuckInstruction) getAspirate(volumes []wunit.Volume, useLLF []bool) RobotInstruction {
```
msackman-synthace: wow, we legit have a use case where we would want llf on some heads but not all?

msackman-synthace: (actually, I'm not sure whether llf is applying to a head, or a well?)

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -1746,6 +1744,103 @@ func (ins *SuckInstruction) GetParameter(name InstructionParameter) interface{}
...
+	aspins.Overstroke = ins.Overstroke
+	aspins.What = ins.What
+	aspins.Plt = ins.FPlateType
+
+	for _, vol := range volumes {
+		aspins.Volume = append(aspins.Volume, wunit.CopyVolume(vol))
+	}
+	aspins.LLF = append(aspins.LLF, useLLF...)
+	for len(aspins.LLF) < aspins.Multi {
+		aspins.LLF = append(aspins.LLF, false)
```
msackman-synthace: how come you append to LLF to make sure it's at least as long as Multi (though it could also be longer), but you don't treat Volume the same way?

microArch/driver/liquidhandling/compositerobotinstruction.go
msackman-synthace: I thought it was in general best practice to return the ground type, not the interface type?

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -2672,9 +2604,13 @@ type SetDriveSpeedInstruction struct {
 	Speed float64
 }
 
-func NewSetDriveSpeedInstruction() *SetDriveSpeedInstruction {
+// NewSetDriveSpeedInstruction set the speed at which the head will move
+// drive should be "X", "Y", or "Z", and speed is the speed in mm/s
+func NewSetDriveSpeedInstruction(drive string, speed float64) *SetDriveSpeedInstruction {
```
msackman-synthace: can't we use axis here?

microArch/driver/liquidhandling/robotinstruction.go
```diff
@@ -383,9 +383,9 @@ func (sori *SetOfRobotInstructions) UnmarshalJSON(b []byte) error {
 		case "MIX":
 			ins = NewMixInstruction()
 		case "SPS":
-			ins = NewSetPipetteSpeedInstruction()
+			ins = NewSetPipetteSpeedInstruction(0, 0, 0.)
```
msackman-synthace: did you miss a 0 on the end?

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -1484,9 +1465,33 @@ func (self *VirtualLiquidHandler) SetPipetteSpeed(head, channel int, rate float6
 	return ret
 }
 
-//SetDriveSpeed - used
+// SetDriveSpeed sets the speed at which the head will move.
+// Drive should be one of "X", "Y", "Z", rate is expressed in mm/s.
+// XXX Bug: it is currently not possible to select which head assembly should be affected
+// currently assume we're talking about group zero
 func (self *VirtualLiquidHandler) SetDriveSpeed(drive string, rate float64) driver.CommandStatus {
```
msackman-synthace: same on this one - can we use axis here too?

antha/anthalib/wunit/wdimension.go
haydnKing: Hmm, I had this at some point, but something made me change it... I'll have a look and see if it can come back

inventory/testinventory/platelibrary.go
haydnKing: Yeah, the model serialisation is a bit bonkers (see `LHWell.SetLiquidLevelModel`)... I'd like to tidy it up, but I fear getting trapped in "helpers" again

microArch/driver/liquidhandling/compositerobotinstruction.go
haydnKing: It's applying to a channel (also referred to as a cone (and also as a "probe" by trilution)) within an adaptor within a head... anyway, I don't think we should ever support a mixture of LLF and non-LLF operations in a single command, even in more advanced liquidhandlers I'd guess such a command actually gets broken down into two independent commands because of how the servos/head articulation is likely to work. Thing is changing it will require changing the liquidhandler interface

microArch/driver/liquidhandling/robotinstruction.go
haydnKing: two `int`s and a `float64`, non? or is it go style to include the trailing 0?

inventory/testinventory/platelibrary.go
msackman-synthace: Yeah, I've just spent 15 mins digging through this. I don't see why we can't simplify this, (maybe there are other services that use this data?) but I agree this is out of the scope of this ticket.

microArch/driver/liquidhandling/robotinstruction.go
msackman-synthace: Yeah, it's just 2 lines further down, you've used `0.0`. Consistency ftw (personally, I reckon 0.0 is more readable than 0.)

microArch/driver/liquidhandling/compositerobotinstruction.go
msackman-synthace: Ok, so I think I would certainly benefit from some more code comments explaining what the difference is between eg a channel and an adaptor - it's not obvious to me.

On a slightly different note, given that we have both the Suck instruction and the Aspirate instruction, I'm struggling to understand a) what the difference is between them; and b) why these changes only apply to Suck.

microArch/driver/liquidhandling/robotinstruction.go
haydnKing: Doh!

antha/anthalib/wunit/wdimension.go
haydnKing: ah yeah, so it could be switched, but they you have
`Velocity3D{someVelocity, someOtherVelocity, yetAnotherVelocity}`
instead of 
```
Velocity3D{
X: thisIsTheXVelocity,
Y: thisIsTheYVelocity,
Z: thisIsTheZVelocity,
}
```
which is clearer I think

msackman-synthace: Totally agree that's clearer, but consider this: https://play.golang.org/p/sEr3xLalIKg
Yes, you can actually do indices in a literal array declaration. See the examples at the bottom of https://golang.org/ref/spec#Composite_literals

However, as I said at the start - this is minor - I really don't have strong feelings in either direction.

msackman-synthace: Ahh, I found the relevant text:

> For array and slice literals the following rules apply:
>
> * Each element has an associated integer index marking its position in the array.
> * An element with a key uses the key as its index. The key must be a non-negative constant representable by a value of type int; and if it is typed it must be of integer type.
> * An element without a key uses the previous element's index plus one. If the first element has no key, its index is zero.

haydnKing:
I've just discovered a bug in this that's causing some errant moves... I'll write some tests of SuckInstruction.Generate

haydnKing:
Ok, I'm gonna park this one pending a response from gilson regarding the bizarre trilution "error" message... personally I'm of the opinion that we should nix LLF entirely if it's going to cause unpredictable (though seemingly deterministic) errors that only occur at device runtime

PR #901, opened 2018-09-25
[Antha-2111] independent multichannel

haydnKing:
This change is to fix support for generating multichannel instructions with different volume in each channel.

testing of the main planning function, `liquidhandler.Plan` was unified and extended to cover independent multichannel, as well as several other cases which were not previously tested for.


haydnKing:
I believe this is now OK, but I'll double check tomorrow AM that all tests have been superseded and add more tests where appropriate

microArch/driver/liquidhandling/convertinstructions.go
```diff
@@ -40,46 +40,26 @@ import (
...
-	tfrs, err := convertInstructions(inssIn, robot, carryvol, channelprms, multi, legacyVolume)
-
-	if err != nil {
-		return []*TransferInstruction{}, robot, err
+func ConvertInstructions(ctx context.Context, inssIn LHIVector, robot *LHProperties, carryvol wunit.Volume, channelprms *wtype.LHChannelParameter, multi int, legacyVolume bool, policy *wtype.LHPolicyRuleSet) ([]*TransferInstruction, error) {
+	// we call convertInstructions twice because
+	// 1) calling convertInstructions with multi = 8 when there are no actuall multichannel instructions causes
+	//    undesirable source volume selection, see tests "TestExecutionPlanning/single_channel_well_use", and
+	//    "TestExecutionPlanning/single_channel_auto_allocation"
+	// 2) convertInstructions makes changes to robot, meaning that it must be called exactly once with the the copy of robot passed to the function
```
msackman-synthace: wait, you start by suggesting you're going to justify why you're calling it twice, and then you say here it must be called exactly once... help me out?! :)

microArch/driver/liquidhandling/convertinstructions.go
```diff
@@ -40,46 +40,26 @@ import (
...
+		return convertInstructions(inssIn, robot, carryvol, channelprms, multi, legacyVolume)
+	} else {
+		return convertInstructions(inssIn, robot, carryvol, channelprms, 1, legacyVolume)
 	}
-
-	return tfrs, robot, nil
 }
 
-func hasMCB(ctx context.Context, tfrs []*TransferInstruction, rbt *LHProperties, policy *wtype.LHPolicyRuleSet) (bool, error) {
+func hasMultiChannelBlock(ctx context.Context, tfrs []*TransferInstruction, rbt *LHProperties, policy *wtype.LHPolicyRuleSet) (bool, error) {
```
msackman-synthace: +1 for expansion of TLAs

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -469,30 +469,29 @@ func (ins *TransferInstruction) Generate(ctx context.Context, policy *wtype.LHPo
...
-			// if we can't do it, we can't do it
-			if maxvol.IsZero() {
-				continue
+				// if we can't do it, we can't do it
+				if maxvol.IsZero() {
+					continue
+				}
+				for i := range vols {
+					vols[i] = wunit.CopyVolume(maxvol)
+				}
```
msackman-synthace: I'd invert this - now it's so small, it may look better to lift the for loop into the if and invert the test - avoiding having to use continue. Up to you though - it's a personal taste thing to be sure.

microArch/driver/liquidhandling/transferparams.go
```diff
@@ -76,24 +76,29 @@ func (mtp MultiTransferParams) RemoveInitialBlanks() MultiTransferParams {
...
 }
 
-func (mtp MultiTransferParams) RemoveVolume(vol wunit.Volume) {
-	for _, t := range mtp.Transfers {
-		if !t.Zero() {
-			t.Volume.Subtract(vol)
+// RemoveVolumes reduce the volume of the ith transfer by vols[i]
+func (mtp MultiTransferParams) RemoveVolumes(vols []wunit.Volume) {
+	for i, t := range mtp.Transfers {
+		if !t.Zero() && i < len(vols) {
```
msackman-synthace: hang on, `i` is only going to increase. So it seems to me you should be doing an `if i >= len(vols) { break }` before this block?

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+}
+
+func (test *PlanningTest) checkPositionConsistency(t *testing.T) {
+	for pos := range test.Liquidhandler.Properties.PosLookup {
+
+		id1, ok1 := test.Liquidhandler.Properties.PosLookup[pos]
+		id2, ok2 := test.Liquidhandler.FinalProperties.PosLookup[pos]
+
+		if ok1 && !ok2 || ok2 && !ok1 {
```
msackman-synthace: `ok1 != ok2` ?

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -0,0 +1,442 @@
...
+		switch p1.(type) {
+		case *wtype.Plate:
+			pp1 := p1.(*wtype.Plate)
+			pp2 := p2.(*wtype.Plate)
+			if pp1.Type != pp2.Type {
+				t.Fatal(fmt.Sprintf("Plates at %s not same type: %s %s", pos, pp1.Type, pp2.Type))
+			}
+			it := wtype.NewAddressIterator(pp1, wtype.ColumnWise, wtype.TopToBottom, wtype.LeftToRight, false)
+
+			for {
```
msackman-synthace: I think you should be able to do a:

`for it := wtype.NewAddressIterator(...); it.Valid(); it.Next() {`

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -0,0 +1,442 @@
...
+				t.Fatal(fmt.Sprintf("Tipwaste changed type: %s %s", tw1.Type, tw2.Type))
+			}
+		}
+
+	}
+
+}
+
+func (test *PlanningTest) expected(err error) bool {
+	if (err == nil) && test.ErrorString == "" {
```
msackman-synthace: As you're doing a test for a prefix, I'd be tempted to rename the test.ErrorString field to test.ErrorPrefix or such.

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -0,0 +1,442 @@
...
+		return true
+	} else if err != nil {
+		return strings.HasPrefix(err.Error(), test.ErrorString)
```
msackman-synthace: bug bug bug. If test.ErrorString is the empty string, "", then this will always pass. So you want `err != nil && test.ErrorString != ""`

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -0,0 +1,442 @@
...
+				continue
+			} else {
+				for address, well := range plate.Wellcoords {
+					if !well.IsEmpty() {
+						got[address] = well.Contents().CName
+					}
+				}
+			}
+
+			if !reflect.DeepEqual(expected[plateNum], got) {
```
msackman-synthace: if you lift this if block into the else above, you can then drop the two `continue`s

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -0,0 +1,442 @@
...
+			return nil, errors.New("nil input plate in request")
+		} else {
+			for address, well := range plate.Wellcoords {
+				if !well.IsEmpty() {
+					got[address] = well.CurrentVolume().MustInStringUnit("ul").RawValue()
+				}
+			}
+		}
+
+		ret = append(ret, got)
```
msackman-synthace: again, I'd lift this into the else given the control flow here.

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -0,0 +1,442 @@
...
+		}
+	}
+
+	return true
+}
+
+// AssertInitialInputVolumes check that the input layout is as expected
+// expected is a map of well location (in A1 format) to liquid to volume in ul
+// tol is the maximum difference before an error is raised
+func AssertInitialInputVolumes(tol float64, expected ...map[string]float64) Assertion {
```
msackman-synthace: Naming... Assert is a verb, which means I expect this function to do the asserting. But it doesn't - it return a function that'll do it later. So to me, I think naming changing these from AssertFooBar to FooBarAssertion makes sense.

antha/anthalib/wtype/lhwell.go
```diff
@@ -465,6 +465,12 @@ func (w *LHWell) Clean() {
 	w.Extra = newExtra
 }
 
+func (w *LHWell) updateComponentLocation() {
+	if w.WContents != nil {
+		w.WContents.Loc = fmt.Sprintf("%s:%s", IDOf(w.Plate), w.Crds.FormatA1())
```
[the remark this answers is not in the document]
m-i-sadowski: It's partly historical - the extra effort required for maintaining these object relationships post serialize / deserialize ended up being more than seemed reasonable at the time

Perhaps a slightly better reason is that the location can be either where something *is* or where it is *going to be*, although this is only slightly better since I think we really should be representing this directly.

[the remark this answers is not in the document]
haydnKing: personally I think that would make a great deal more sense, and I'd be keen to make that change since the current method is difficult to trace, but I'm also unsure if there's a particular reason for the string representation... @m-i-sadowski ?

microArch/driver/liquidhandling/convertinstructions.go
haydnKing: it needs to be called with `robot` once (because it makes changes that need to be propagated back), but the first call uses a copy of `robot`.
Other options include 
1) returning a new robot (which was what was meant to be happening before, but on closer inspection the returned copy has no way of propagating far enough up the call stack without a significant refacator)
2) changing the call signature to `robot **LHProperties`, but I figured this would get messy

antha/anthalib/wtype/lhwell.go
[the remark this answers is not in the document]
m-i-sadowski: To summarise the brief conversation we just had I'm not 100% sure that pointers to wells are the right way to go here. Once the functions this field performs (i.e. both *current* and *future* locations of a component) are separated out and we have decided whether or not we still want something like a sample tracker to exist we can revisit. 

In the short term I agree a string representation is not correct, however. A PlateLocation type already exists in wtype for this purpose, I'd be happy to see a refactor into that as a separate PR.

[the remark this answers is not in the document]
msackman-synthace: All sounds good to me :)

microArch/scheduler/liquidhandling/planning_test.go
haydnKing: good spot!

microArch/scheduler/liquidhandling/planning_test.go
haydnKing: oh 💩

haydnKing:
Ok, I've sorted through the tests - there was indeed one test which had slipped - so thing we're ready for another run at this

microArch/scheduler/liquidhandling/planning_test.go
```diff
@@ -168,10 +156,10 @@ func (test *PlanningTest) checkPositionConsistency(t *testing.T) {
 }
 
 func (test *PlanningTest) expected(err error) bool {
-	if (err == nil) && test.ErrorString == "" {
+	if err != nil && test.ErrorPrefix != "" {
+		return strings.HasPrefix(err.Error(), test.ErrorPrefix)
+	} else if err == nil && test.ErrorPrefix == "" {
```
msackman-synthace: ohhh, I think we can now flatten this further - would this work:

```
        if err != nil && test.ErrorPrefix != "" {
		return strings.HasPrefix(err.Error(), test.ErrorPrefix)
	} else {
                return err == nil && test.ErrorPrefix == ""
        }
```

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -471,15 +471,14 @@ func (ins *TransferInstruction) Generate(ctx context.Context, policy *wtype.LHPo
...
-				maxvol := vols.MaxMultiTransferVolume(prms.MinPossibleVolume())
-
-				// if we can't do it, we can't do it
-				if maxvol.IsZero() {
+				if maxvol := vols.MaxMultiTransferVolume(prms.MinPossibleVolume()); !maxvol.IsZero() {
+					for i := range vols {
+						vols[i] = wunit.CopyVolume(maxvol)
+					}
+				} else {
+					// we can't transfer any volumes with the non-independent head so move on to the next parallelset
```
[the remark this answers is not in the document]
haydnKing: no worries - I think the scoping is a bit clearer now, which is why I went for it

microArch/scheduler/liquidhandling/planning_test.go
haydnKing: aw yeah... surprised the linter didn't catch that actually, it usually picks up on `if (...) {return true} return false`

PR #903, opened 2018-09-28
All absorbance data

chr1sgrant:
This change adds a more general method to the AbsorbanceData definition which returns all absorbance data in a generic format with all absorbance data returned for all wells.

### DIFF
--- a/antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
+++ b/antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
@@ -123,6 +123,29 @@ func (data MarsData) TimeCourse(wellname string, exWavelength int, emWavelength 
 	return
 }
 
+// AllAbsorbanceData returns all absorbance readings for a specified well.
+// An error is returned if the well is not found.
+func (data MarsData) AllAbsorbanceData(wellName string) (readings []wtype.Absorbance, err error) {
+
+	wellData, found := data.Dataforeachwell[wellName]
+
+	if !found {
+		return readings, fmt.Errorf("No data found for well %s", wellName)
+	}
+
+	for _, measurement := range wellData.Data.Readings[0] {
+		readings = append(readings,
+			wtype.Absorbance{
+				Wavelength:   float64(measurement.RWavelength),
+				Reading:      measurement.Reading,
+				WellLocation: wtype.MakeWellCoordsA1(wellName),
+				Annotations:  measurement.ReadingType,
+			})
+	}
+
+	return readings, nil
+}
+
 // AbsScanData returns all wavelengths and readings for a specified well.
 func (data MarsData) AbsScanData(well string) (wavelengths []int, Readings []float64) {
 	wavelengths = make([]int, 0)
@@ -257,8 +280,9 @@ func (data MarsData) Absorbance(well string, wavelength int, options ...interfac
 		result, err := data.ReadingsAsAverage(well, platereader.EMWAVELENGTH, wavelength, fmt.Sprint(options[0]))
 		if err == nil {
 			return wtype.Absorbance{
-				Reading:    result,
-				Wavelength: float64(wavelength),
+				Reading:     result,
+				Wavelength:  float64(wavelength),
+				Annotations: fmt.Sprint(options[0]),
 			}, nil
 		} else {
 			errs = append(errs, err.Error())
--- a/antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
+++ b/antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
@@ -126,20 +126,19 @@ func (data MarsData) TimeCourse(wellname string, exWavelength int, emWavelength 
 // AllAbsorbanceData returns all absorbance readings using the well location as key.
 func (data MarsData) AllAbsorbanceData() (readings map[string][]wtype.Absorbance, err error) {
 
-	readings = make(map[string][]wtype.Absorbance)
+	readings = make(map[string][]wtype.Absorbance, len(data.Dataforeachwell))
 
 	for wellName, wellData := range data.Dataforeachwell {
 
-		var wellReadings []wtype.Absorbance
+		var wellReadings = make([]wtype.Absorbance, len(wellData.Data.Readings[0]))
 
-		for _, measurement := range wellData.Data.Readings[0] {
-			wellReadings = append(wellReadings,
-				wtype.Absorbance{
-					Wavelength:   float64(measurement.RWavelength),
-					Reading:      measurement.Reading,
-					WellLocation: wtype.MakeWellCoordsA1(wellName),
-					Annotations:  []string{measurement.ReadingType},
-				})
+		for readingIndex, measurement := range wellData.Data.Readings[0] {
+			wellReadings[readingIndex] = wtype.Absorbance{
+				Wavelength:   float64(measurement.RWavelength),
+				Reading:      measurement.Reading,
+				WellLocation: wtype.MakeWellCoordsA1(wellName),
+				Annotations:  []string{measurement.ReadingType},
+			}
 		}
 
 		readings[wellName] = wellReadings
--- a/antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
+++ b/antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
@@ -139,7 +139,7 @@ func (data MarsData) AllAbsorbanceData(wellName string) (readings []wtype.Absorb
 				Wavelength:   float64(measurement.RWavelength),
 				Reading:      measurement.Reading,
 				WellLocation: wtype.MakeWellCoordsA1(wellName),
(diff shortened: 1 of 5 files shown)


antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
```diff
@@ -123,6 +123,31 @@ func (data MarsData) TimeCourse(wellname string, exWavelength int, emWavelength
...
 }
 
+// AllAbsorbanceData returns all absorbance readings using the well location as key.
+func (data MarsData) AllAbsorbanceData() (readings map[string][]wtype.Absorbance, err error) {
+
+	readings = make(map[string][]wtype.Absorbance)
+
+	for wellName, wellData := range data.Dataforeachwell {
+
+		var wellReadings []wtype.Absorbance
```
msackman-synthace: For speed, I would definitely use make with a capacity here - you know the size upfront here (len(wellData.Data.Readings[0]))

antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
```diff
@@ -123,6 +123,31 @@ func (data MarsData) TimeCourse(wellname string, exWavelength int, emWavelength
 	return
 }
 
+// AllAbsorbanceData returns all absorbance readings using the well location as key.
+func (data MarsData) AllAbsorbanceData() (readings map[string][]wtype.Absorbance, err error) {
+
+	readings = make(map[string][]wtype.Absorbance)
```
msackman-synthace: You also know the needed size here (len(data.Dataforeachwell)) so I would use that too as a capacity argument when making this map.

antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
```diff
@@ -257,8 +282,9 @@ func (data MarsData) Absorbance(well string, wavelength int, options ...interfac
 		result, err := data.ReadingsAsAverage(well, platereader.EMWAVELENGTH, wavelength, fmt.Sprint(options[0]))
 		if err == nil {
 			return wtype.Absorbance{
-				Reading:    result,
-				Wavelength: float64(wavelength),
+				Reading:     result,
+				Wavelength:  float64(wavelength),
+				Annotations: []string{fmt.Sprint(options[0])},
```
msackman-synthace: Hmm, the comment for this method really suggests that options should be a string. Is there any need for it to be interface{} ?

antha/AnthaStandardLibrary/Packages/platereader/dataset/spectramax.go
```diff
@@ -113,73 +113,89 @@ type Well struct {
...
 	}
 	return 0.0, fmt.Errorf("No reading found for wavelength %d: found: %+v", wavelength, readings)
 }
 
-// GetDataByWell returns all readings for that well.
-func (s SpectraMaxData) GetDataByWell(wellName string) (readings []WavelengthReading, err error) {
+// AllAbsorbanceData returns all readings for using the well name as key.
+func (s SpectraMaxData) AllAbsorbanceData() (readings map[string][]wtype.Absorbance, err error) {
+
+	readings = make(map[string][]wtype.Absorbance)
```
msackman-synthace: Same here - you can provide a capacity which will help with performance.

antha/AnthaStandardLibrary/Packages/platereader/dataset/spectramax.go
```diff
@@ -113,73 +113,89 @@ type Well struct {
...
-			break
-		}
-	}
 
-	if !wellFound {
-		return readings, fmt.Errorf("No readings found for well %s: found: %+v", wellName, s.Experiment[0].PlateSections[0].Wavelengths[0])
-	}
+	for _, w := range wells {
+
+		var wellReadings []wtype.Absorbance
```
msackman-synthace: And here too, wellReadings can be created with the perfect capacity upfront too.

antha/AnthaStandardLibrary/Packages/platereader/dataset/spectramax.go
```diff
@@ -113,73 +113,89 @@ type Well struct {
...
+}
 
-	} else {
-		return readings, fmt.Errorf("Only Spectramax data in scan format is currently supported. Please run Absorbance reading as scan")
+func (s SpectraMaxData) dataForWell(wellName string) ([]wtype.Absorbance, error) {
+
+	allWellData, err := s.AllAbsorbanceData()
+
+	if err != nil {
+		return []wtype.Absorbance{}, err
```
msackman-synthace: Typically, if there's an error, it's more normal to return nil for a slice than an empty slice.

antha/AnthaStandardLibrary/Packages/platereader/dataset/spectramax.go
```diff
@@ -113,73 +113,89 @@ type Well struct {
...
 	}
-	return
+
+	wellData, found := allWellData[wellName]
+
+	if !found {
+		return []wtype.Absorbance{}, fmt.Errorf("No data found for well %s", wellName)
```
msackman-synthace: Same on this one.

antha/anthalib/wtype/absorbance.go
```diff
@@ -31,8 +31,13 @@ type Absorbance struct {
...
 	Wavelength   float64
 	Pathlength   wunit.Length
-	Status       []string
 	Reader       string
+	// Annotations is a field to add custom user labels
+	Annotations []string
+}
+
+func (a Absorbance) WavelengthAsInt() int {
+	return int(a.Wavelength)
```
msackman-synthace: Err, this is a bit concerning really - obviously going from a float to an int risks throwing away data. If this method is needed, I feel it would be good if it was well documented with warnings.

antha/AnthaStandardLibrary/Packages/platereader/dataset/mars.go
```diff
@@ -126,20 +126,19 @@ func (data MarsData) TimeCourse(wellname string, exWavelength int, emWavelength
 // AllAbsorbanceData returns all absorbance readings using the well location as key.
 func (data MarsData) AllAbsorbanceData() (readings map[string][]wtype.Absorbance, err error) {
 
-	readings = make(map[string][]wtype.Absorbance)
+	readings = make(map[string][]wtype.Absorbance, len(data.Dataforeachwell))
 
 	for wellName, wellData := range data.Dataforeachwell {
 
-		var wellReadings []wtype.Absorbance
+		var wellReadings = make([]wtype.Absorbance, len(wellData.Data.Readings[0]))
```
msackman-synthace: Minor stylistic: I tend to avoid using `var` whenever possible - this case (and lower down at line 140) you can use `:=`.

antha/anthalib/wtype/absorbance.go
```diff
@@ -36,6 +36,16 @@ type Absorbance struct {
...
+// WavelengthAsInt will return the Wavelength field as an int.
+// Whilst it is possible that the wavelength used may be a decimal,
+// Wavelength would typically be expected to be in the form of an integer of the wavelength in nm.
+// In some platereader data sets this is stored as a float so this method is
+// intended to take the safest representation, as a float, and return the more
+// common representation, as an int, for parsers where it is known that the wavelength
+// is stored as an int.
+// This method would therefore not be safe to use for situations
+// where the wavelength may in fact be represented by a decimal.
+//
```
msackman-synthace: Excellent comment, thank you!

Now, the problem is that a float64 can be a much bigger number than an int. The range of a float64 is ±2.23×10^−308 to ±1.80×10^308. Whereas an int (assuming it's actually an int64) can only go up to 2^63. Now if you're super confident that we'll never see a wavelength in nm > math.MaxInt64 then I would recommend that you do something like this:

```
func (...) WavelengthAsInt() int {
  if a.Wavelength > float64(math.MaxInt64) {
    panic("...some suitable error message...")
  }
  return int(a.Wavelength)
}
```

FWIW, I'd also prefer to see uint64 rather than just int - a negative wavelength seems non-sensical, and int is always a bit confusing as the width depends on the cpu you're running on. However, I can totally see that if we've not standardised on uint64 elsewhere, then changing this may cause awkward problems elsewhere.

chr1sgrant: Ok I see what you mean but in the context of an Absorbance reading the working range would be maximum 100 - 1000nm. 

Other spectroscopy (X ray, IR) may be outside of this range but that would no longer be Absorbance. 
The more likely potential issue to be mindful of would be the potential for a reading at say 595.5 becoming truncated to 595. 

I've actually refactored this to use the wunit.Length in another PR which I will make a PR for shortly anyway, with that in mind the best course of action may be to refactor the name of this method with that in mind...?

```go
func (a Abs) WavelengthToNearestNM()int
```

msackman-synthace: Yup, makes sense to me - if it's going to be changed to use Length consistently soon then I think just renaming the method as you suggest will be totally fine.

PR #904, opened 2018-09-28
[Antha-2113] % w/v volume fix

haydnKing:
fixes JSONification of concentrations with units which contain strings, e.g. "% w/v"

### DIFF
--- a/antha/anthalib/wunit/parse.go
+++ b/antha/anthalib/wunit/parse.go
@@ -1,0 +1,47 @@
+// wunit/unitfromstring.go: Part of the Antha language
+// Copyright (C) 2014 the Antha authors. All rights reserved.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+//
+// For more information relating to the software or licensing issues please
+// contact license@antha-lang.org or write to the Antha team c/o
+// Synthace Ltd. The London Bioscience Innovation Centre
+// 2 Royal College St, London NW1 0NH UK
+
+package wunit
+
+import (
+	"github.com/pkg/errors"
+	"strconv"
+)
+
+// extractFloat extract the longest valid float from the left hand side of
+// the string and return it and the remainder of the string
+// returns an error if no float is found
+func extractFloat(s string) (float64, string, error) {
+	var longest int
+	var ret float64
+	for i := range s {
+		if f, err := strconv.ParseFloat(s[:i], 64); err == nil {
+			ret = f
+			longest = i
+		}
+	}
+
+	if longest == 0 {
+		return 0.0, s, errors.Errorf("unable to extract float from %q", s)
+	}
+	return ret, s[longest:], nil
+}
--- a/antha/anthalib/wunit/parse.go
+++ b/antha/anthalib/wunit/parse.go
@@ -23,14 +23,13 @@
 package wunit
 
 import (
-	"github.com/pkg/errors"
 	"strconv"
 )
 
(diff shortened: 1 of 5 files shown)


antha/anthalib/wunit/serialize_test.go
```diff
@@ -24,11 +24,94 @@ package wunit
...
+			t.Error(err)
+		} else if !reflect.DeepEqual(input, output) {
+			t.Errorf("before and output don't match:\ne: %v\ng: %v", input, output)
+		}
+	})
+}
+
+func TestVolumeSerialization(t *testing.T) {
+	for _, unit := range GetGlobalUnitRegistry().ListValidUnitsForType("Volume") {
+		for _, value := range []float64{-5.91209102398, 0.0, 1.123131231231e58} {
```
msackman-synthace: So. You're going to hate me, but what about +Inf, -Inf, NaN ?

antha/anthalib/wunit/parse.go
```diff
@@ -0,0 +1,47 @@
...
+	"strconv"
+)
+
+// extractFloat extract the longest valid float from the left hand side of
+// the string and return it and the remainder of the string
+// returns an error if no float is found
+func extractFloat(s string) (float64, string, error) {
+	var longest int
+	var ret float64
+	for i := range s {
```
msackman-synthace: Exciting features of Go, unicode, and `range`: by doing this, `i` will increase by whole runes - i.e. it may not go up one by one, which is perfect for this use case. :+1:

haydnKing: oh nice, I didn't actually realise that... how novel that the basic string is more than []char

msackman-synthace: It's pretty nuts - everyone goes on about how basically string == []byte, but most of the time you actually want to treat it as []rune. If you index a string by a number, you may jump into the middle of a rune, which is a pretty common gotcha. But if you first convert to []rune, then you're fine.

antha/anthalib/wunit/serialize.go
```diff
@@ -42,20 +40,19 @@ func marshal(x stringer) ([]byte, error) {
...
-func unmarshal(b []byte) (value float64, unit string, err error) {
+func unmarshal(b []byte) (float64, string, error) {
 	var s *string
-	if err = json.Unmarshal(b, &s); err != nil {
-		return
-	} else if s == nil {
-		return
+	if err := json.Unmarshal(b, &s); err != nil {
+		return 0.0, "", err
+	} else if s == nil || *s == "" {
```
msackman-synthace: you could combine these first two branches: `; err != nil || s == nil || *s == "" { return 0.0, "", err }`

antha/anthalib/wunit/serialize.go
```diff
@@ -42,20 +40,19 @@ func marshal(x stringer) ([]byte, error) {
...
+		return 0.0, "", nil
 	}
-	if _, err = fmt.Fscanf(strings.NewReader(*s), "%e%s", &value, &unit); err != nil && err == io.EOF {
-		unit = ""
-		if _, err = fmt.Fscanf(strings.NewReader(*s), "%e", &value); err != nil {
-			return
-		}
+
+	if v, u := extractFloat(*s); len(u) == len(*s) {
```
msackman-synthace: I'd give a more meaningful name to `u` - `tail` or something like that perhaps. Also, this method should have a comment at the very least explaining that the 2nd return value is not the whole string...

antha/anthalib/wunit/serialize_test.go
```diff
@@ -25,10 +25,102 @@ package wunit
...
+func TestMassSerialization(t *testing.T) {
+	for _, unit := range GetGlobalUnitRegistry().ListValidUnitsForType("Mass") {
+		for _, value := range []float64{-5.91209102398, 0.0, 1.123131231231e58} {
+			massSerializationTest(t, fmt.Sprintf("%g %s", value, unit), NewMass(value, unit))
+			massSerializationTest(t, fmt.Sprintf("%g%s", value, unit), NewMass(value, unit))
+		}
+	}
+}
+
+func timeSerializationTest(t *testing.T, input string, expected Time) {
```
msackman-synthace: wait wait, you've basically duplicated the same thing 3 times here. json.Unmarshal works on an interface anyway because it's doing reflection internally. And measurementsEqual just requires a Measurement. So I reckon you may be able to collapse the inner func massively - I think you could be able to wind up with something like this (dup myType for Mass/Concentration/Time):

```
func myTypeSerializationTest(t *testing.T, input string, expected MyType) {
  t.Run(input, func(t *testing.T) {
    var output MyType
    runSerializationIdentity(t, input, expected, &output)
  }))
}

func runSerializationIdentity(t *testing.T, input string, expected, output Measurement) {
...
}
```

PR #906, opened 2018-10-04
Refactor/absorbance

chr1sgrant:


antha/AnthaStandardLibrary/Packages/platereader/Platereader.go
```diff
@@ -87,12 +75,16 @@ func EstimatePathLength(plate *wtype.Plate, volume wunit.Volume) (pathlength wun
...
 func PathlengthCorrect(pathlength wunit.Length, reading wtype.Absorbance) (pathlengthcorrected wtype.Absorbance) {
-
-	referencepathlength := wunit.NewLength(10, "mm")
-
-	pathlengthcorrected.Reading = reading.Reading * referencepathlength.RawValue() / pathlength.RawValue()
-	return
+	pathlengthcorrected = reading.Dup()
+	err := pathlengthcorrected.PathLengthCorrect(pathlength)
+	if err != nil {
+		panic(err)
```
msackman-synthace: What are the rules in AnthaStdLib for using panic vs the Error intrinsic?

antha/AnthaStandardLibrary/Packages/platereader/Platereader.go
```diff
@@ -108,36 +100,16 @@ shifts in chemical equilibria as a function of concentration
...
-	return
-}
-
-//example
-
-/*
-func OD(Platetype wtype.LHPLate,wellvolume wtype.Volume,reading wtype.Absorbance) (od wtype.Absorbance){
-volumetopathlengthconversionfactor := 0.0533//WellCrosssectionalArea
-OD = (Blankcorrected_absorbance * 10/(total_volume*volumetopathlengthconversionfactor)// 0.0533 could be written as function of labware and liquid volume (or measureed height)
+	return wunit.NewConcentration(concfloat, "Mol/l"), nil
```
msackman-synthace: Probably best to be consistent as to whether litre should be `l` or `L`.

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,165 @@
...
+	"fmt"
+	"math"
+	"strings"
+
 	"github.com/antha-lang/antha/antha/anthalib/wunit"
+	"github.com/pkg/errors"
 )
 
+// CorrectionType is a label given to describe the nature of the correction of an absorbance reading.
+type CorrectionType string
```
msackman-synthace: If corrections are being limited to absorbances, then I think absorbance should be in its name - `AbsorbanceCorrectionType`. It's quite odd to see `Type` in the name of a `type`, though reading the comments in the consts below, it does read ok to me. I wonder though whether there's a suitable synonym - `AbsorbanceCorrectionMethod` maybe? I'm not sure if that's actually better...

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,165 @@
...
+	return nil
 }
 
-func (sample *Absorbance) PathlengthCorrect(pathlength wunit.Length) {
+func toNM(l wunit.Length) float64 {
+	return l.SIValue() * wunit.Nano.Value
+}
+
+// Dup creates a duplicate of the absorbance reading, with exact equality for all values.
+func (sample *Absorbance) Dup() Absorbance {
```
msackman-synthace: I would suggest that Dup should return the same type as it takes - in this case `*Absorbance`.

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,165 @@
...
-// where the wavelength may in fact be represented by a decimal.
+// where the wavelenWavelengthToNearestNm be represented by a decimal.
 //
-func (a Absorbance) WavelengthAsInt() int {
-	return int(a.Wavelength)
+func (a Absorbance) WavelengthToNearestNm() int {
+	if uint64(toNM(a.Wavelength)) > math.MaxUint64 {
+		panic(errors.Errorf("the value for wavelength %v cannot be safely converted to an integer value", a.Wavelength))
+	}
+	return int(toNM(a.Wavelength))
```
msackman-synthace: I like that you're doing the range test against Uint64. But for that to be consistent, you ideally should be returning a uint64 too. I can understand though if we're using `int` generally within elements, this could cause problems. So maybe:
1. Test against MaxInt64, not MaxUint64
2. add a comment basically saying "yes, I know the comparison with MaxInt64 is not the same as int, but for now it's better than nothing"

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,165 @@
...
-	WellLocation WellCoords
-	Reading      float64
-	Wavelength   float64
-	Pathlength   wunit.Length
-	Reader       string
+	WellLocation WellCoords             `json:"WellLocation"`
+	Reading      float64                `json:"Reading"`
+	Wavelength   wunit.Length           `json:"Wavelength"`
+	Pathlength   wunit.Length           `json:"Pathlength"`
+	Corrections  []AbsorbanceCorrection `json:"Corrections"`
```
msackman-synthace: Would you ever have > 1 correction of the same type? If not, I wonder whether this would be better as `map[CorrectionType]AbsorbanceCorrection` ?

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,165 @@
...
 }
 
-func (sample *Absorbance) BlankCorrect(blank Absorbance) {
-	if sample.Wavelength == blank.Wavelength &&
-		sample.Pathlength == blank.Pathlength &&
-		sample.Reader == blank.Reader {
-		sample.Reading = sample.Reading - blank.Reading
+// IsPathLengthCorrected returns true if the absorbance reading has been pathlength corrected
+func (a Absorbance) IsPathLengthCorrected() bool {
+	for _, correction := range a.Corrections {
```
msackman-synthace: if it were possible to change the type of Corrections, then you wouldn't need the iteration here...

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,165 @@
...
+		Reader:       sample.Reader,
+		Source:       sample.Source,
+		Annotations:  append([]string{}, sample.Annotations...),
+	}
+}
+
+// BlankCorrect subtracts the blank reading from the sample absorbance.
+// If the blank sample is not equivalent to the sample, based on wavelength and pathlength, an error is returned.
+func (sample *Absorbance) BlankCorrect(blanks ...Absorbance) error {
+	var errs []string
```
msackman-synthace: in `antha/utils` there's an errorSlice type which you may find nicer.

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,165 @@
...
+	for _, blank := range blanks {
+		if sample.Wavelength.EqualToRounded(blank.Wavelength, 9); sample.Pathlength.EqualToRounded(blank.Pathlength, 4) && sample.Reader == blank.Reader {
```
msackman-synthace: I don't know what these magic numbers, 9 and 4, mean. Could you bring them out to consts?

antha/anthalib/wtype/absorbance.go
chr1sgrant: refactored this to a map as suggested

antha/anthalib/wtype/absorbance.go
chr1sgrant: I could not find this

antha/AnthaStandardLibrary/Packages/platereader/Platereader.go
```diff
@@ -24,34 +24,22 @@
...
-	abs.Wavelength = wavelength
-	// add calculation to work out pathlength from volume and well geometry abs.Pathlength
-
-	return abs
-}
-
-func Blankcorrect(blank wtype.Absorbance, sample wtype.Absorbance) (blankcorrected wtype.Absorbance, err error) {
+// BlankCorrect subtracts the blank reading from the sample absorbance.
+// If the blank sample is not equivalent to the sample, based on wavelength and pathlength, an error is returned.
+func Blankcorrect(sample *wtype.Absorbance, blank *wtype.Absorbance) (blankcorrected *wtype.Absorbance, err error) {
```
msackman-synthace: casing - did you mean BlankCorrect (as you've written in the comment)?

antha/AnthaStandardLibrary/Packages/platereader/Platereader.go
```diff
@@ -87,12 +75,16 @@ func EstimatePathLength(plate *wtype.Plate, volume wunit.Volume) (pathlength wun
...
-func PathlengthCorrect(pathlength wunit.Length, reading wtype.Absorbance) (pathlengthcorrected wtype.Absorbance) {
-
-	referencepathlength := wunit.NewLength(10, "mm")
-
-	pathlengthcorrected.Reading = reading.Reading * referencepathlength.RawValue() / pathlength.RawValue()
-	return
+// PathLengthCorrect normalises an absorbance reading
+// to a standard reference pathlength of 1cm.
+// 1cm is the pathlength used to normalise absorbance readings to OD.
+func PathlengthCorrect(pathlength wunit.Length, reading *wtype.Absorbance) (pathlengthcorrected *wtype.Absorbance) {
```
msackman-synthace: similarly PathlengthCorrect or PathLengthCorrect?

antha/AnthaStandardLibrary/Packages/platereader/Platereader.go
msackman-synthace: It's odd that blankcorrect returns an error but this one panics. Should they not behave the same?

antha/anthalib/wtype/absorbance.go
msackman-synthace: Totally agree with that idea for naming. However, I fully understand if you feel that is out of scope for this PR.

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,176 @@
...
 // is stored as an int.
 // This method would therefore not be safe to use for situations
-// where the wavelength may in fact be represented by a decimal.
+// where the wavelenWavelengthToNearestNm be represented by a decimal.
 //
-func (a Absorbance) WavelengthAsInt() int {
-	return int(a.Wavelength)
+func (a Absorbance) WavelengthToNearestNm() int {
+	// necessary to use int64 here in order to conveniently evaluate size.
+	if int64(toNM(a.Wavelength)) > math.MaxInt64 {
```
msackman-synthace: ahh, so the maximum value you can fit into an int64 is math.MaxInt64. So by definition, this test can never fail because your cast will shrink a larger value from the float64. I think you just need to remove the cast - I think that sholud still compile because the math.MaxInt64 I believe are just untyped consts. If anything you actually want to promote that to the wider float64 type. So if it won't compile with no casts, try `... > float64(math.MaxInt64)`

antha/anthalib/wtype/absorbance.go
```diff
@@ -23,53 +23,176 @@
...
+		}
+	}
+	return false
+}
 
-		sample.Annotations = append(sample.Annotations, "Blank Corrected")
+// AddSource associates the Absorbance sample to the liquid which it corresponds to.
+// This is the actual sample that was measured rather than the original parent
+// sample which may have been diluted.
+func (a *Absorbance) AddSource(l *Liquid) error {
```
msackman-synthace: If it's called `Add` I would expect it's adding to a set/bag/list/map. To me, this would be better named as `SetSource` because it's destructive - it unsets the previous source.

antha/anthalib/wtype/absorbance.go
msackman-synthace: https://github.com/antha-lang/antha/blob/master/utils/errors.go

antha/anthalib/wtype/absorbance.go
chr1sgrant: Actually the current behaviour is that if a Source is already set then an error is returned and the liquid is not updated; this is probably overly strict; there should probably be a means of overriding an existing  source

PR #911, opened 2018-10-11
support reading concentration of a liquid with a bracketed name

chr1sgrant:
Code adapted to handle edge case encountered in the wild.


antha/anthalib/wunit/unitfromstring.go
```diff
@@ -56,7 +56,7 @@ func ParseConcentration(s string) (bool, Concentration, string) {
 	if l, r := strings.Index(s, "("), strings.LastIndex(s, ")"); l >= 0 && l < r {
 		value, unit := extractFloat(strings.TrimSpace(s[l+1 : r]))
 		if trimUnit := strings.TrimSpace(unit); !reg.ValidUnitForType("Concentration", trimUnit) {
-			return false, NewConcentration(0, "g/l"), s
+			// do not return yet, continue to try next liquid with concentration format
 		} else if componentName := strings.TrimSpace(s[:l] + s[r+1:]); componentName != "" {
 			return true, NewConcentration(value, trimUnit), componentName
 		} else {
```
haydnKing: I think this would be a cleaner way of achieving the same thing
```
// try and extract a valid unit
if trimUnit := strings.TrimSpace(unit); reg.ValidUnitForType("Concentration", trimUnit) {
    if componentName := strings.TrimSpace(s[:l] + s[r+1:]); componentName != "" {
        return true, NewConcentration(value, trimUnit), componentName
    } else {
        return true, NewConcentration(value, trimUnit), s
    }
}
```

msackman-synthace:
This sort of thing worries me. Just because it's seen in the wild doesn't mean we have to support it.

IMO, we should be setting simple and clear rules for what we do support and what we don't for importing external files. I fear that if we continue down the current path then the behaviour will become highly complex, possibly inconsistent, and quite possibly indescribable. The probability that we extract the _wrong_ number from such a title will increase, leading to surprises down the line. Hence why I think we should have very simple rules so that if we are rejecting an external file, it is clear why and it is clear what the user needs to change in order to get the file accepted. Trying to support everything we come across seems to me to be a recipe for disaster.

chr1sgrant:
I see what you mean and agree but in this particular case this structure is actually very common in chemical names so I think we should support it e.g. (S)-Styrene, (Phenyl) Salicylate, Boc-Phe(4-NH2)-OH

The convention can be simple if we stipulate that the unit must be a separate field to the rest of the compound name  and either at the back or the front:

6M Glucose 
(D)Glucose 6M
(D)Glucose (6M)

antha/anthalib/wunit/unitfromstring.go
```diff
@@ -52,15 +52,27 @@ func ParseConcentration(s string) (bool, Concentration, string) {
...
+
+	// unit location indicated by parentheses before component Name- e.g. "(6M) Glucose"
+	if l, r := strings.Index(s, "("), strings.Index(s, ")"); l >= 0 && l < r && strings.HasPrefix(s, "(") {
+		value, unit := extractFloat(strings.TrimSpace(s[l+1 : r]))
+		if trimUnit := strings.TrimSpace(unit); reg.ValidUnitForType("Concentration", trimUnit) {
+			if componentName := strings.TrimSpace(s[:l] + s[r+1:]); componentName != "" {
+				return true, NewConcentration(value, trimUnit), componentName
+			} else {
+				return true, NewConcentration(value, trimUnit), s
+			}
```
haydnKing: hmm, i'm not particularly happy to add two new special cases particularly since the brackets are optional... which makes we wonder if we should support brackets around the measurement at all? If we only support the measurement being at the start or the end then brackets aren't adding any information we don't have already so surely they're superfluous? Sorry to be a pain, but given the vast number of times this function is called during simulation we should be really cautions about adding to it.

If we really need to, I'd rather do something like: 
```go
func dropLeadingBracket(s string) string {
  if closingIndex := strings.Index(s, ")"); closingIndex >= 0 && strings.HasPrefix(s, "(")  {
     return s[1:closingIndex] + s[closingIndex+1:]
  }
  return s
}
```
then use `SplitValueAndUnit(dropLeadingBrackets(s))` below

chr1sgrant: OK, we can remove the support for the unit in brackets before the liquid name, then it's just one case. My preference of all of these formats to use is the liquid name followed by unit in brackets. Glucose (5g/L)

antha/anthalib/wunit/unitfromstring.go
```diff
@@ -52,15 +52,15 @@ func ParseConcentration(s string) (bool, Concentration, string) {
 
 	reg := GetGlobalUnitRegistry()
 
-	// unit location indicated by parentheses - e.g. "Glucose (6M)"
-	if l, r := strings.Index(s, "("), strings.LastIndex(s, ")"); l >= 0 && l < r {
+	// unit location indicated by parentheses after component Name - e.g. "Glucose (6M)"
+	if l, r := strings.LastIndex(s, "("), strings.LastIndex(s, ")"); l >= 0 && l < r && strings.HasSuffix(s, ")") {
```
msackman-synthace: This is fine, but it could be a fraction clearer: you are actually enforcing that the last char of the string s is ")". So... you could just do that:

```go
if l, r := strings.LastIndex(s, "("), len(s)-1; l >= 0 && s[r] == ")" {
```

which is a bit shorter and more obvious (at least to me)

PR #915, opened 2018-10-19
Fix/missing volumes litres

m-i-sadowski:
This PR fixes an issue where some logic meant to prevent trivial volumes being requested (by rounding very small remainders down to zero) ignored the units of the volumes and therefore considered any volume with an absolute value of <0.0001 as trivial *regardless of unit*. 

This caused an issue when volumes were originally specified in larger-than-normal units, such as litres.

I have added a check to take units into account and some tests to show intended behaviour and test for regression. I've also run this against antha-tester with no problems.

### DIFF
--- a/antha/anthalib/wtype/componentvector.go
+++ b/antha/anthalib/wtype/componentvector.go
@@ -124,17 +124,17 @@ func (cv ComponentVector) getLocTok(x int) []string {
 	return ret
 }
 
-func (cv ComponentVector) Equal(cv2 ComponentVector) bool {
-	if len(cv) != len(cv2) {
+func (cv1 ComponentVector) Equal(cv2 ComponentVector) bool {
+	if len(cv1) != len(cv2) {
 		return false
 	}
 
-	for i := 0; i < len(cv); i++ {
-		if cv[i] != nil && cv2[i] != nil {
-			if !cv[i].EqualTypeVolumeID(cv2[i]) {
+	for i := 0; i < len(cv1); i++ {
+		if cv1[i] != nil && cv2[i] != nil {
+			if !cv1[i].EqualTypeVolumeID(cv2[i]) {
 				return false
 			}
-		} else if !(cv[i] == nil && cv2[i] == nil) {
+		} else if !(cv1[i] == nil && cv2[i] == nil) {
 			return false
 		}
 	}
--- a/antha/anthalib/wtype/componentvector.go
+++ b/antha/anthalib/wtype/componentvector.go
@@ -123,3 +123,21 @@ func (cv ComponentVector) getLocTok(x int) []string {
 
 	return ret
 }
+
+func (cv ComponentVector) Equal(cv2 ComponentVector) bool {
+	if len(cv) != len(cv2) {
+		return false
+	}
+
+	for i := 0; i < len(cv); i++ {
+		if cv[i] != nil && cv2[i] != nil {
+			if !cv[i].EqualTypeVolumeID(cv2[i]) {
+				return false
+			}
+		} else if !(cv[i] == nil && cv2[i] == nil) {
+			return false
+		}
+	}
+
+	return true
+}
--- a/antha/anthalib/wtype/liquid.go
+++ b/antha/anthalib/wtype/liquid.go
@@ -1134,14 +1134,14 @@ func (cmp Liquid) FullyQualifiedName() string {
 	}
 }
 
-func (l *Liquid) EqualTypeVolume(l2 *Liquid) bool {
-	return l.CName == l2.CName && l.Volume().EqualTo(l2.Volume())
+func (l1 *Liquid) EqualTypeVolume(l2 *Liquid) bool {
+	return l1.CName == l2.CName && l1.Volume().EqualTo(l2.Volume())
 }
 
-func (l *Liquid) EqualTypeVolumeID(l2 *Liquid) bool {
-	if !l.EqualTypeVolume(l2) {
+func (l1 *Liquid) EqualTypeVolumeID(l2 *Liquid) bool {
+	if !l1.EqualTypeVolume(l2) {
 		return false
 	}
 
-	return l.ID == l2.ID
+	return l1.ID == l2.ID
 }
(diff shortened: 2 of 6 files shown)


microArch/driver/liquidhandling/getComponents.go
```diff
@@ -369,7 +369,10 @@ func updateDests(dst wtype.ComponentVector, match wtype.Match) wtype.ComponentVe
 	for i := 0; i < len(match.M); i++ {
 		if match.M[i] != -1 {
 			dst[i].Vol -= match.Vols[i].ConvertToString(dst[i].Vunit)
```
m-i-sadowski: For aid of reading: this bit's actually the change, the rest is test code and stuff in support of that

antha/anthalib/wtype/componentvector.go
```diff
@@ -123,3 +123,21 @@ func (cv ComponentVector) getLocTok(x int) []string {
...
 }
+
+func (cv ComponentVector) Equal(cv2 ComponentVector) bool {
+	if len(cv) != len(cv2) {
+		return false
+	}
+
+	for i := 0; i < len(cv); i++ {
+		if cv[i] != nil && cv2[i] != nil {
+			if !cv[i].EqualTypeVolumeID(cv2[i]) {
```
msackman-synthace: not `return cv[i].EqualTypeVolumeID(cv2[i])` ?

antha/anthalib/wtype/componentvector.go
```diff
@@ -123,3 +123,21 @@ func (cv ComponentVector) getLocTok(x int) []string {
 
 	return ret
 }
+
+func (cv ComponentVector) Equal(cv2 ComponentVector) bool {
```
msackman-synthace: (cosmetic) it may be slightly clearer if you either call these `a` and `b`, or `cv1` and `cv2`. Having `cv` and `cv2` slightly threw me.

antha/anthalib/wtype/liquid.go
```diff
@@ -1133,3 +1133,15 @@ func (cmp Liquid) FullyQualifiedName() string {
 		return cmp.Kind()
 	}
 }
+
+func (l *Liquid) EqualTypeVolume(l2 *Liquid) bool {
+	return l.CName == l2.CName && l.Volume().EqualTo(l2.Volume())
+}
+
+func (l *Liquid) EqualTypeVolumeID(l2 *Liquid) bool {
```
msackman-synthace: similarly on these two, it might be fractionally neater if the var names were more balanced. But, clearly, cosmetic concern.

microArch/driver/liquidhandling/getComponents_test.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func TestUpdateDests(t *testing.T) {
+	// func updateDests(dst wtype.ComponentVector, match wtype.Match) wtype.ComponentVector
+
+	testCases := []testCase{
+		{
+			Name:     "Regression1 - Don't lose relatively small transfers",
+			Cmps:     wtype.ComponentVector{&wtype.Liquid{Vol: 0.01, Vunit: "l"}},
+			Match:    wtype.Match{M: []int{0}, Vols: []wunit.Volume{wunit.NewVolume(0.00999, "l")}},
```
msackman-synthace: we support big units too - El should would I think. It's important that tests are amusing ;)

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -369,7 +369,8 @@ func updateDests(dst wtype.ComponentVector, match wtype.Match) wtype.ComponentVe
...
-			if dst[i].Vol < 0.0001 {
+
+			if dst[i].Volume().MustInStringUnit("ul").RawValue() < 0.0001 {
```
msackman-synthace:
```bash
> grep -r 0\\.00 | grep .go:|grep -v AnthaStandardLibrary|grep -v test|grep -v .pb.
microArch/driver/liquidhandling/getComponents.go:                       if checkVol > 0.0001 && checkVol < mpv.ConvertToString(dst[i].Vunit) {
microArch/driver/liquidhandling/getComponents.go:                       if dst[i].Vol < 0.0001 {
antha/anthalib/wtype/makelhpolicy.go:   IntoEmpty                 = numericCondition{Class: "WELLTOVOLUME", Range: conditionRange{Lower: 0.0, Upper: 0.009}}
antha/anthalib/wtype/liquid.go: if lhc == nil || lhc.Type == "" || lhc.CName == "" || lhc.Vol < 0.0000001 {
antha/anthalib/wtype/liquid.go: if cmp2.Vol == 0.00 && cmp2.Tvol > 0.00 {
antha/anthalib/wtype/liquid.go: } else if cmp.Tvol > 0.00 {
antha/anthalib/wtype/lhwell.go: tolerance := wunit.NewVolume(0.000001, "ul")
antha/anthalib/wunit/wunit.go:  if isNil(cm) || math.Abs(cm.Mvalue) < 0.00000000001 {
```

Is there any chance we could unify a bunch of these to a named const somewhere?

m-i-sadowski: There actually is (was?) one in wtype... can we consider that a separate issue though? While I agree it's related it is scope creep compared to the initial purpose of this change.

msackman-synthace: I don't think this is the first PR I've reviewed which has added a 0.00....1 magic number in. I'm happy for that to be another PR, provided it's a PR that gets done very soon! :)

haydnKing: it's also complicated by the fact that the unit that you choose to compare in is important, so there's always gonna be at least one magic value popping up until we are able to abandon floats altogether

m-i-sadowski: Yes I totally agree - I think in a lot of cases like this when the actual device capabilities are known it should be possible to have a locally set value which is relative to machine tolerances as well as a global Planck-volume type thing which sets the absolute lower limit for other cases.

haydnKing: i'm looking forward to doing plank-volume scale liquid handling

PR #916, opened 2018-10-19
Fix issue with splitting mixed instruction layers

m-i-sadowski:
This PR fixes a bug whereby instruction nodes which contain prompt and split instructions caused a runtime error by extending the existing method for splitting up mixed  nodes to include this case. 

I've run this with the test bundles and it checks out fine.

### DIFF
--- a/microArch/scheduler/liquidhandling/ichain.go
+++ b/microArch/scheduler/liquidhandling/ichain.go
@@ -182,47 +182,103 @@ func (it *IChain) getOrderedLHInstructions(acc []*wtype.LHInstruction) []*wtype.
 	}
 }
 
-func (it *IChain) SplitMixedNodes() {
-	if it.hasMixAndSplitOnly() {
-		it.splitMixedNode()
+func (it *IChain) SplitMixedNodes() *IChain {
+	var ret *IChain
+
+	if len(it.getInstructionTypes()) != 1 {
+		ret = it.splitMixedNode()
 	}
 
-	// stop if we reach the end
-	if it.Child == nil {
-		return
+	if it.Child != nil {
+		// carry on
+		it.Child.SplitMixedNodes()
 	}
 
-	// carry on
-	it.Child.SplitMixedNodes()
+	if ret != nil {
+		return ret
+	}
+
+	return it
 }
 
-func (it *IChain) splitMixedNode() {
-	// put mixes first, then splits
+func (it *IChain) FindEnd() *IChain {
+	if it.Child == nil {
+		return it
+	}
+
+	return it.Child.FindEnd()
+}
+
+func (it *IChain) splitMixedNode() *IChain {
+	// put mixes first, then splits, then prompts
 
 	mixValues := make([]*wtype.LHInstruction, 0, len(it.Values))
 	splitValues := make([]*wtype.LHInstruction, 0, len(it.Values))
+	promptValues := make([]*wtype.LHInstruction, 0, len(it.Values))
 
 	for _, v := range it.Values {
 		if v.Type == wtype.LHIMIX {
 			mixValues = append(mixValues, v)
 		} else if v.Type == wtype.LHISPL {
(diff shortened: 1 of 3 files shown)


microArch/scheduler/liquidhandling/ichain.go
```diff
@@ -182,47 +182,103 @@ func (it *IChain) getOrderedLHInstructions(acc []*wtype.LHInstruction) []*wtype.
...
+		ch.Values = v
+
+		if ch.Parent != nil {
+			ch.Parent.Child = ch
+		}
+
+		ch = NewIChain(ch)
+	}
+
+	return top
```
msackman-synthace: Hmm. If you call makeNew... with no args, we would get back a non-nil IChain. It feels to me that both no args, and only-empty-list-args should return nil? So I think that would look maybe something like this?

```go
var top, cur *IChain
for _, v := range vals {
  if len(v) == 0 {
    continue
  }
  cur = NewIChain(cur)
  if top == nil {
    top = cur
  }
  cur.Values = v
}
return top

File: microArch/scheduler/liquidhandling/ichain.go [hunk 2]
```diff
@@ -182,47 +182,103 @@ func (it *IChain) getOrderedLHInstructions(acc []*wtype.LHInstruction) []*wtype.
...
+
+	for _, v := range vals {
+		if len(v) == 0 {
+			continue
+		}
+
+		ch.Values = v
+
+		if ch.Parent != nil {
+			ch.Parent.Child = ch
```
msackman-synthace: I really don't like this. I dislike that half the stitching is done in the NewIChain constructor (i.e. setting the child's parent pointer) and the other half (setting the parent's child pointer) you have to remember to do in the call site. Can we not do both in the constructor?

microArch/scheduler/liquidhandling/ichain.go
```diff
@@ -182,47 +182,103 @@ func (it *IChain) getOrderedLHInstructions(acc []*wtype.LHInstruction) []*wtype.
...
+		}
+
+		ch = NewIChain(ch)
+	}
+
+	return top
+}
+
+// remove the current node and replace with newch
+// return newch as the new head if necessary
```
msackman-synthace: It's Friday, but still, it's taken me 5 mins of staring at this code to convince myself it's right, and to decipher what "if necessary" means. Could you expand that comment a little?

Also, there's one call site for this method, and it appears to not care very much about whether it gets back a nil or not. I wonder whether it might be more intuitive if this method always returns the head of the chain?

m-i-sadowski:
Thanks for the helpful suggestions! I've revised and re-tested with antha-tester; all still well

microArch/scheduler/liquidhandling/ichain.go
```diff
@@ -182,47 +182,103 @@ func (it *IChain) getOrderedLHInstructions(acc []*wtype.LHInstruction) []*wtype.
...
+func (it *IChain) SplitMixedNodes() *IChain {
+	var ret *IChain
+
+	if len(it.getInstructionTypes()) != 1 {
+		ret = it.splitMixedNode()
+	}
+
+	if it.Child != nil {
+		// carry on
+		it.Child.SplitMixedNodes()
```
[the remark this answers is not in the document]
m-i-sadowski: makes sense, I've done this and it seems fine

microArch/scheduler/liquidhandling/ichain.go
m-i-sadowski: great, also done

msackman-synthace:
Sorry, no way github will let me annotate the relevant file: because of the improvement new ichain.NewIChain, in executionhelpers.go line 77, that line can now be deleted.

I can't find any other cases where NewIChain is being called and we were manually stitching parents and children together.

msackman-synthace:
Ok, tests still pass. Merging.

PR #918, opened 2018-10-24
List deps

ddn0:
# Why
Help visualize dependencies between elements

# Example of use
```bash
antha list elementDependencies \
  --outputPackage repos.antha.com/elements \
  --byPath \
  --pathMatch 'Elements/([^/]+).*' \
  --pathReplace '$1' \
  Elements/
```

# Test Plan
- [x] compare generated files before and after this change


antha/compile/element.go
```diff
@@ -139,23 +139,53 @@ func manglePackageName(pkg string) string {
...
+	HasSteps      bool
+	HasValidation bool
+	HasSetup      bool
+	HasAnalysis   bool
+	LineMap       map[int]int
+	DependsOn     []string
+}
+
+// Element is a preprocessing pass from antha element file to go file
+type Element struct {
```
msackman-synthace: This does not feel like a great name...

antha/compile/element.go
```diff
@@ -1279,71 +1326,73 @@ var (
...
-	type Param struct {
-		Name     string
-		BareName string
-		Desc     string
-		Kind     string
-		BareKind string
+
+	info := p.Info()
+
+	funcs := map[string]interface{}{
```
msackman-synthace: Hmm. It's pretty minor, but I kinda preferred the old approach of doing the quoting once before building the struct.

cmd/antha/cmd/list_element_dependencies.go
```diff
@@ -0,0 +1,156 @@
...
+		if err := filepath.Walk(path, func(path string, f os.FileInfo, err error) error {
+			if err != nil {
+				return err
+			}
+
+			if f.IsDir() {
+				return nil
+			}
+
+			if !isElementFile(f.Name()) {
```
msackman-synthace: you could save so many loc by joining up these ifs! ;)

ddn0:
Relative to `go list`,
- (👍) slightly more special behavior for the `model` and `element` sub-packages (by merging them with their root) which are kind of implementation details of the current transpilation process and no dependency on `go`
- (👍) also element paths are not yet encoded in the import paths so `file blah/blah/element.an` => `import basepath/element` so we wouldn't be able to go back from the go files to their original element definitions without another map
- (👎) `go list` will give us automatic support for modules (useful at some unspecified future state) and build flags (probably not useful).

Any approach will basically boil down to parsing for `import` statements.

antha/compile/element.go
ddn0: Is the original `Antha` better for the pass?

I think the other `Antha` to `Element` changes seem more clear for the actual file level data, but happy to revert this aspect.

msackman-synthace: Ahh, damn this internet. I meant to delete that comment. That one I decided was aces :)

antha/compile/element.go
msackman-synthace: Yeah, I guessed that's why you made the change. Hey ho.

antha/compile/element.go
```diff
@@ -1279,71 +1326,73 @@ var (
...
-		LineMap       map[int]int
-	}
-
-	elementPath := normalizePath(p.elementPath)
+	return template.Must(template.New("").Funcs(funcs).Parse(tmpl)).Execute(out, info)
+}
 
-	tv := TVars{
+func (p *Element) Info() *ElementInfo {
+	info := &ElementInfo{
```
msackman-synthace: To be honest, I don't really see what is gained by having ElementInfo separate from Element - to me, it seems to make more sense to just make a bunch of the fields in Element public and maybe add some helper methods (eg things like `HasSteps`) could be methods.

antha/compile/elementfiles.go
```diff
@@ -0,0 +1,41 @@
...
+func NewElementFiles() *ElementFiles {
+	return &ElementFiles{}
+}
+
+// Files returns the files in the collection
+func (f *ElementFiles) Files() []*ElementFile {
+	return f.files
+}
+
+func (f *ElementFiles) addFile(name string, data []byte) {
```
msackman-synthace: pffft. Why not just make the `files` field public and then you don't have to do arguably-unnecessary wrapping.

cmd/antha/cmd/compile.go
```diff
@@ -86,15 +86,17 @@ func runCompile(cmd *cobra.Command, args []string) error {
...
 			}
 
-			if !isAnthaFile(f.Name()) {
+			if !isElementFile(f.Name()) {
 				return nil
 			}
 
 			// Collect errors processing errors
-			if err := processFile(root, path, outdir); err != nil {
+			elem, err := processFile(root, path)
```
msackman-synthace: Why did you move this out of the if-initializer? It was perfectly comfy in there! ;)

cmd/antha/cmd/list_element_dependencies.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	}
+	buf.WriteString("}\n") // nolint
+
+	io.Copy(os.Stdout, &buf) // nolint
+
+	return nil
+}
+
+func init() {
```
msackman-synthace: I _really_ want to ban any PR that adds more init() funcs...

cmd/antha/cmd/list_element_dependencies.go
```diff
@@ -0,0 +1,156 @@
...
+		var err error
+		pat, err = regexp.Compile(match)
+		if err != nil {
+			return err
+		}
+	}
+
+	var convertTo map[string]string
+	if viper.GetBool("byPath") {
+		convertTo = make(map[string]string)
```
msackman-synthace: meh, you gain approx zilch by gating this `make`. I'd just do the `convertTo := make(...)` above.

cmd/antha/cmd/list_element_dependencies.go
```diff
@@ -0,0 +1,156 @@
...
+
+		return pat.ReplaceAllString(v, replace)
+	}
+
+	type Pair struct {
+		Src string
+		Dst string
+	}
+
+	edges := make(map[Pair]bool)
```
msackman-synthace: I would totes do `map[src][]dst`

cmd/antha/cmd/list_element_dependencies.go
```diff
@@ -0,0 +1,156 @@
...
+	io.Copy(os.Stdout, &buf) // nolint
+
+	return nil
+}
+
+func init() {
+	c := listElementDependencies
+	listCmd.AddCommand(c)
+	flags := c.Flags()
+	flags.String("outputPackage", "", "base package name for generated files")
```
msackman-synthace: I don't understand why this flag is necessary, nor what it does. For example:

```bash
> antha list elementDependencies --outputPackage xxx --byPath --pathMatch 'Elements/([^/]+).*' --pathReplace '$1'   Elements/ | sort > /tmp/xxx
> antha list elementDependencies --outputPackage repos.antha.com/elements --byPath --pathMatch 'Elements/([^/]+).*' --pathReplace '$1'   Elements/ | sort > /tmp/yyy
> diff -u /tmp/xxx /tmp/yyy
--- /tmp/xxx    2018-10-25 12:48:24.742013437 +0100
+++ /tmp/yyy    2018-10-25 12:48:30.698024137 +0100
@@ -1,14 +1,18 @@
 }
 digraph {
 "repos.antha.com/antha-ninja/elements-westeros/AnthaAcademy"
+"repos.antha.com/antha-ninja/elements-westeros/AnthaAcademy" -> "repos.antha.com/antha-ninja/elements-westeros/MVL"
 "repos.antha.com/antha-ninja/elements-westeros/DOE"
 "repos.antha.com/antha-ninja/elements-westeros/DOE" -> "repos.antha.com/antha-ninja/elements-westeros/MVL"
+"repos.antha.com/antha-ninja/elements-westeros/DOE" -> "repos.antha.com/antha-ninja/elements-westeros/New-MVL"
 "repos.antha.com/antha-ninja/elements-westeros/Molecular_Biology"
 "repos.antha.com/antha-ninja/elements-westeros/Molecular_Biology" -> "repos.antha.com/antha-ninja/elements-westeros/Molecular_Biology_Starter"
 "repos.antha.com/antha-ninja/elements-westeros/Molecular_Biology_Starter"
 "repos.antha.com/antha-ninja/elements-westeros/Molecular_Biology_Starter" -> "repos.antha.com/antha-ninja/elements-westeros/MVL"
+"repos.antha.com/antha-ninja/elements-westeros/Molecular_Biology_Starter" -> "repos.antha.com/antha-ninja/elements-westeros/New-MVL"
 "repos.antha.com/antha-ninja/elements-westeros/MVL"
 "repos.antha.com/antha-ninja/elements-westeros/New-MVL"
+"repos.antha.com/antha-ninja/elements-westeros/New-MVL" -> "repos.antha.com/antha-ninja/elements-westeros/MVL"
 "repos.antha.com/antha-ninja/elements-westeros/PCR"
 "repos.antha.com/antha-ninja/elements-westeros/QPCR"
 "repos.antha.com/antha-ninja/elements-westeros/QPCR" -> "repos.antha.com/antha-ninja/elements-westeros/MVL"
@@ -18,4 +22,5 @@
 "repos.antha.com/antha-ninja/elements-westeros/Testing" -> "repos.antha.com/antha-ninja/elements-westeros/Molecular_Biology_Starter"
 "repos.antha.com/antha-ninja/elements-westeros/Testing" -> "repos.antha.com/antha-ninja/elements-westeros/MVL"
 "repos.antha.com/antha-ninja/elements-westeros/Testing" -> "repos.antha.com/antha-ninja/elements-westeros/PCR"
+"repos.antha.com/antha-ninja/elements-westeros/Testing" -> "repos.antha.com/antha-ninja/elements-westeros/Utility"
 "repos.antha.com/antha-ninja/elements-westeros/Utility"
```


```

File: cmd/antha/cmd/list_element_dependencies.go [hunk 11]
```diff
@@ -0,0 +1,156 @@
...
+	flags.Bool("byPath", false, "if set, show dependencies on file paths instead of import paths")
+	flags.String("pathMatch", "", "regex substitution to apply to paths, e.g., prefix(\\w+)")
+	flags.String("pathReplace", "", "regex substitution to apply to paths, e.g., $1")
```
msackman-synthace: I'm not really a fan of these flags - can't we achieve the same thing just piping through to `sed` ?

cmd/antha/cmd/list_element_dependencies.go
ddn0: This is to get easy deduping :D

cmd/antha/cmd/list_element_dependencies.go
ddn0: `--byPath` means emit paths to elements instead import paths of elements. It's necessary because elements are organized by file path not import path at the moment (and helpful right now because reorganization doesn't require changing the contents of files).

So, the effect of  `--outputPackage` is obscured `--byPath` such that any unique prefix will result in the same output.

cmd/antha/cmd/list_element_dependencies.go
ddn0: The difference is that the command will dedupe edges for you while sed doesn't. 

I do agree that the whole compilation process is a little jury-rigged at the moment.

cmd/antha/cmd/list_element_dependencies.go
ddn0: `--outputPackage` changes the package of generated elements but some elements themselves directly import `repos.antha.com/elements/XXX` which is not affected by `--outputPackage`. 

Summary: when `--outputPackage` doesn't match the one used it `antha compile` funny things will happen.

Happy to take suggestions on how to make this easier to understand, but it's kinda approaching the larger goal of just redoing the entire compilation flow.

msackman-synthace: I would just write a decent README explaining all of this because it's not obvious to me even after reading the code, and the help text doesn't fill in all the gaps.

antha/compile/element.go
ddn0: Not ready to solve this in this PR since there is a difference between how the pass wants data and the form that outsiders might want this data.

cmd/antha/cmd/list_element_dependencies.go
```diff
@@ -18,7 +18,19 @@ import (
 var listElementDependencies = &cobra.Command{
 	Use:   "elementDependencies <files or directories>",
 	Short: "List antha element dependencies",
-	RunE:  runListElementDependencies,
+	Long: `List antha element dependencies
+
+To generate correct results, the value of outputPackage should match
+that used in "antha compile".
+
+By default, this command will show dependencies between import paths.  If you
```
msackman-synthace: I thought you wanted to show dependencies between elements. I don't know how an import path can depend on another import path. I guess what you mean is that this command outputs the graph of dependencies between elements. By default elements are named by their import path, however by using the --byPath option, elements are instead named by their path within your filesystem.

Sorry, I'm in a grumpy mood. But I do think this text could be clearer.

ddn0: Your copy is much clearer. I'll adopt.

msackman-synthace:
That is now some good docs. Thank you. :)

msackman-synthace:
I have to say, overall, I am struggling with this PR.

This PR:
* adds ~200 loc
* adds no tests
* has no existing tests that provide coverage of the bulk of the modified code
* adds new structs that are substantially similar to existing structs

I'm basically struggling to convince myself that this is a net improvement to the code base.

PR #919, opened 2018-10-29
Refactor/instruction graph move

m-i-sadowski:
The new layout code will be a client of these types, so this move is a simple way to avoid the import cycle which would otherwise result from importing them from scheduler/liquidhandling, which will obviously need to refer to the layout code as well. 

I think this code clearly belongs here anyway:  much of it depends on wtype and almost none of it depends on anything else, including anything where it was previously located.


m-i-sadowski:
Please squash these on merge BTW

PR #920, opened 2018-11-06
Refactor/replace glpk

m-i-sadowski:
This PR replaces calls to GLPK's MIP solver with the pure-go implementation of vanilla LP provided by gonum where necessary. 

There were previously two places in which LP was used: 
1. determining how many wells of what type to be used in allocating input components
2. defining stock solution concentrations

2 was deprecated some time back so this is just a case of removal. In the case of 1. the situation is a bit more complicated as this mechanism is used widely. I have not found an integer linear programming implementation which seems to have the right license for us so despite the inherently integral nature of the problem I've reverted to using standard LP and rounding up at the end. 

This works OK in most cases but in the course of adding tests I've discovered some edges which ideally should behave differently. I can however confirm that all the current bundles pass with this implementation so I don't think these cases are critical. 

There is time pressure on this so I have focused on confirming that this works OK with our test bundles. We have a little time to see if we can fix the issues I've uncovered however if we run out of time we should simply skip those tests and get this in. 

In the medium term this should all be deprecated by other changes I have in progress (but not quite ready yet) so it's not worthwhile spending a lot of time fixing this.


microArch/scheduler/liquidhandling/liquidhandling_test.go
```diff
@@ -82,6 +81,7 @@ func GetTroughForTest() *wtype.Plate {
 	return plate
 }
 
+/*
```
haydnKing: would prefer either a comment as to why it's skipped or delete

and eww, `rand` in tests

microArch/scheduler/liquidhandling/solution_setup.go
```diff
@@ -155,14 +155,10 @@ func solution_setup(request *LHRequest, prms *liquidhandling.LHProperties) (map[
 	maxrequired := make(map[string]wunit.Concentration, len(mconcs))
 
 	//TODO this needs to be migrated elsewhere
-	var vmin wunit.Volume = wunit.NewVolume(1.0, "ul")
+	//var vmin wunit.Volume = wunit.NewVolume(1.0, "ul")
```
haydnKing: rm

microArch/scheduler/liquidhandling/solution_setup.go
```diff
@@ -205,7 +201,9 @@ func solution_setup(request *LHRequest, prms *liquidhandling.LHProperties) (map[
 		return nil, nil, fmt.Errorf("min unit %s not equal to max unit %s ", minUnit, maxUnit)
 	}
 
-	stockconcs := choose_stock_concentrations(minSIRequired, maxSIRequired, Smax, vmin.RawValue(), hshTVol)
+	//stockconcs := choose_stock_concentrations(minSIRequired, maxSIRequired, Smax, vmin.RawValue(), hshTVol)
```
haydnKing: feel like the commented out reference to choose_stock_concentrations may cause confusion -- either rm or a line of explanation?

microArch/scheduler/liquidhandling/input_plate_linear.go
```diff
@@ -51,15 +51,24 @@ func choose_plate_assignments(component_volumes map[string]wunit.Volume, plate_t
...
 	//
+	//
 
 	// defense
+	//
 
 	ppt := make([]*wtype.Plate, 0, len(plate_types))
 	h := make(map[string]bool, len(plate_types))
 
+	fmt.Println("Autoallocate plates available:")
```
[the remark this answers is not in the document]
m-i-sadowski: I actually want this particular one to remain as it's been a gotcha in the past for some users.

microArch/scheduler/liquidhandling/solution_setup.go
m-i-sadowski: This whole bit is a weird mashup of Chris's revisions and my original, now entirely deprecated, code. 

I think we should tidy this up but it's not our business here.

PR #923, opened 2018-11-12
ANTHA-2272: optimise for multichannel when requesting next free well in 384 well plate

chr1sgrant:
https://synthace.atlassian.net/browse/ANTHA-2272


antha/AnthaStandardLibrary/Packages/search/plates.go
```diff
@@ -55,6 +57,26 @@ func NextFreeWell(plate *wtype.Plate, avoidWells []string, preferredWells []stri
 
 	allWellPositions := plate.AllWellPositions(byRow)
 
+	// if the plate has greater than 8 rows and the number of rows is divisible by 8,
+	// for example, a 384 well plate,
+	// we'll skip every other row in order to optimise multichanneling.
+	if plate.WellsX() > 8 && plate.WellsX()%8 == 0 && !byRow {
```
haydnKing: it seems to me we're only really dealing with the case where `plate.WellsX() == 16` -  I don't think the code below would behave well for a 1536-well plate for example. We should either only do this for `plate.WellsX() == 16`, or also support at least `plate.WellsX() == 32` (aka. 1536-well)

chr1sgrant: Yes agreed; we'll have to choose whether to just handle 384 well plates or adapt for 1536 and make a bit more clever: will probably go for the latter

chr1sgrant: I've updated to clarify that it's just the 384 well case which it handles at present which is the most important case since 1536 plates are really used.

x-anderson:
@cosmonautspiff has flagged up that this PR has resulted in failures of tests related to detecting the NextFreeWell in a 384-well plate, in the current version of master on elements-westeros.

```
--- FAIL: TestSkipWell (0.34s)
    platePreferences_test.go:343: For:  greinerPlateSkipByColumnColumn 
         expected:  E1 
         got:  G1 
```

We should verify if this needs re-visiting, or if these tests should be updates to reflect this new machanism.

PR #924, opened 2018-11-13
[ANTHA-2277] Correct TextFile.

msackman-synthace:


antha/AnthaStandardLibrary/Packages/export/export.go
```diff
@@ -35,7 +35,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/antha-lang/antha/antha/AnthaStandardLibrary/Packages/AnthaPath"
+	anthapath "github.com/antha-lang/antha/antha/AnthaStandardLibrary/Packages/AnthaPath"
```
msackman-synthace: This is here because in the body of this file we refer to the import already as `anthapath`. This ... I had no idea imports weren't case sensitive. This is mainly insane. Anyway, goimports doesn't understand that imports are case insensitive, so will remove this import line unless we add the alias. If you remove it, the file doesn't build.

haydnKing: #CamelCaseProblems

PR #926, opened 2018-11-19
Antha-2292: only attempt all part combinations for <= than 5 part assemblies

chr1sgrant:
Currently the more comprehensive assembly validation function (`FindAllAssemblyProducts`) tests all part order combinations; this is thorough and more powerful at detecting potential mis assemblies but is very computationally expensive hence there exists a threshold number of parts above which the simpler assembly validation will occur which will only test the validity of the assembly in the part order specified. 

The threshold is currently set to 7 part assemblies (6 parts + vector) which is 7! (5040) part order combinations. In practice this is too computationally expensive to run on multiple assemblies; this PR therefore reduces the threshold to 5 part assemblies (4 parts + vector, 5! = 120) which we know simulates fine with the more comprehensive assembly validation.

### DIFF
--- a/antha/AnthaStandardLibrary/Packages/enzymes/Ligation.go
+++ b/antha/AnthaStandardLibrary/Packages/enzymes/Ligation.go
@@ -506,7 +506,7 @@ func Assemblysimulator(assemblyparameters Assemblyparameters) (s string, success
 	var failedAssemblies []DigestedFragment
 	var plasmidProducts []wtype.DNASequence
 
-	if len(assemblyparameters.Partsinorder) > 6 {
+	if len(assemblyparameters.Partsinorder) > 4 {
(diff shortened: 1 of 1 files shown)


chr1sgrant:
I've now updated the function description to clarify behaviour; also pulled out the threshold as a constant for clarity.

cosmonautspiff:
Is it right to have the threshold specified as a constant? Or should it be passed as an argument (maybe with `thresholdForOnlyTestingPartsInSpecifiedOrder = 0` meaning no threshold)?

I could see `FindAllAssemblyProducts` being used for simulation far before getting to any liquid handling. And in that case, the user might be willing to wait for the simulation to take place.

chr1sgrant:
Agreed, but in that case there is also the standalone function `FindAllAssemblyProducts` which can be used directly

PR #927, opened 2018-12-04
make search for sub component non case sensitive

chr1sgrant:
When a user uses modern software they don't expect case sensitivity and issues with white space when performing searches. In the workflow editor the elements essentially represent the user interface so we've had issues where unfortunately this expectation is not met, which leads to errors and a poor user experience. This PR fixes one source of this.


antha/anthalib/wtype/subcomponents.go
```diff
@@ -278,10 +278,10 @@ func (c ComponentList) GetByName(component string) (conc wunit.Concentration, er
 
 	component = NormaliseName(component)
 
-	conc, found := c.Components[component]
-
-	if found {
-		return conc, nil
+	for key, conc := range c.Components {
+		if equalFold(key, component) {
```
haydnKing: looks like `equalFold` just directly calls `strings.EqualFold` - is there a good reason not to remove `equalFold` and just call the `strings` version directly? It would definitely be less complex that way

antha/anthalib/wtype/subcomponents.go
```diff
@@ -278,10 +278,10 @@ func (c ComponentList) GetByName(component string) (conc wunit.Concentration, er
 
 	component = NormaliseName(component)
 
-	conc, found := c.Components[component]
-
-	if found {
-		return conc, nil
+	for key, conc := range c.Components {
```
haydnKing: I really don't like that we're iterating all the way through the dictionary just to find the matching component... can't we just always store e.g. the lower case version? If previous experience of `strings` is anything to go by I'd guess that `EqualFold` is quite expensive

chr1sgrant: Yes but then it would modify the sub component name from how the user has specified it, which I would not want to do. If they specify the sub component as pUC19 they should receive that back. ..

antha/anthalib/wtype/subcomponents.go
chr1sgrant: it also trims white space so as It's used several times it prevents code duplication and ensures the comparisons are consistent with each other

antha/anthalib/wtype/subcomponents.go
chr1sgrant: Yes I agree we should certainly be consistent at least, so either be case sensitive or handling equalfold cases as duplicates for all sub components, I would favour doing the latter.

chr1sgrant:
Ahh, just suffered case sensitivity problems and it reminded me of this PR; what do we need to do?

An overhaul of the subcomponents field (and subsequent protobuf change) may be good to do soon in order to have a DisplayName field as you suggested but that might be better for another day (i wanted to modify the field to add Sub component Volumes too)

haydnKing:
Hmm, so I can see how this might help in some cases as a temporary fix, but it looks like it might also introduce other problems, like the puc/pUC issue above...
I also can't seem to find any unit testing around setting/getting subcomponents, which is a bit of a worry as it's not all that clear to me exactly how they're meant to work. I think at the very least we should have some testing in place to show that this is actually fixing the problem at hand without breaking normal behaviour

chr1sgrant:
Hmm yes, we've got the more complex tests for mixing (https://github.com/antha-lang/antha/blob/master/antha/anthalib/mixer/subcomponents_test.go) sub components but not getting and setting. I'll add a test

antha/anthalib/wtype/subcomponents_test.go
```diff
@@ -0,0 +1,68 @@
...
+	mediaMixture := newTestComponent("LB",
+		LTWater,
+		9999,
+		wunit.NewConcentration(1, "X"),
+		wunit.NewVolume(2000.0, "ul"),
+		ComponentList{})
+
+	err := mediaMixture.AddSubComponents(someComponents)
+
+	if err != nil {
```
haydnKing: I think `if err := mediaMixture.AddSubComponents(someComponents); err != nil { ... }` is a lot clearer and it avoids polluting the scope - i.e. if you always do this you never have to worry about whether it's `err =` or `err :=`

antha/anthalib/wtype/subcomponents_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+		return c
+	}
+	someComponents := ComponentList{Components: map[string]wunit.Concentration{
+		"glycerol": wunit.NewConcentration(0.25, "g/l"),
+		"IPTG":     wunit.NewConcentration(0.25, "mM/l"),
+		"water":    wunit.NewConcentration(0.25, "v/v"),
+		"LB":       wunit.NewConcentration(0.25, "X"),
+	},
+	}
```
haydnKing: [nit picking] is this correctly formatted? seems odd there'd be two `}`s at the same level on indentation?

chr1sgrant: it's because the Components field of ComponentList is declared on the same line:

```
ComponentList{Components: map[string]wunit.Concentration{
```
guess gofmt does not pick up on that: fixed that in the latest commit now.

antha/anthalib/wtype/subcomponents_test.go
chr1sgrant: ok, made that change

PR #928, opened 2018-12-13
[WIP] a few minor changes

msackman-synthace:
This is getting exciting people!


msackman-synthace:
Ok, this is ready to have a look at. Basically:

* the stuff in cmd/antha is broken - don't try to use it - we need to have a discussion about the various subcommands - what needs keeping, what doesn't, what can't be etc.
* instead, there's cmd/composer
* to use: `cd cmd/composer; go build && rm -rf /tmp/foobar/ && ./composer -config ./sampleConfig.json -workflow ./sampleWorkflow.json`
* then `cd /tmp/foobar ; GOPATH=$GOPATH:/tmp/foobar go run main.go` (and no, it won't actually work just yet!)

The interesting things to look at are:
* the generated code /tmp/foobar/main.go and dig into there for the elements themselves
* changes to antha/compile/antha.go
* everything under laboratory/
* everything under composer/

m-i-sadowski:
For the benefit of others, you may need to run:

go get -u gopkg.in/src-d/go-git.v4/...

before cmd/composer will build. I'm documenting this because they've decided to be awkward and not let us just go get their package like any other

msackman-synthace:
Sorry, yes, I'd forgotten about that. We need to get this repo moved over to go mod asap...

m-i-sadowski:
Feature request / gotcha 1: 

element names are referred to in this style (correct me if I'm wrong): 
    $PREFIX/branch-name/path/to/element/in/repo

where PREFIX is as defined in the config. However branch-name cannot currently include slashes, which they almost always do round here. So we need to support this

a workaround currently is to use the alternative 

    $PREFIX/commit-hash/path/to/element/in/repo

nb some inline documentation in the example config/workflow to cover this point would be a good idea I think!

cmd/composer/sampleWorkflow2.json
```diff
@@ -0,0 +1,118 @@
...
+                "ParameterName": "PlateNameUsed"
+            },
+            "Target": {
+                "ElementInstance": "aliquot_D",
+                "ParameterName": "NewPlateName"
+            }
+        }
+    ],
+
+    "ElementParameters": {
```
jacksongabbard: Why wouldn't you want to hang the parameters for the elements off the elements themselves?

```
    "ElementInstances": {
        "aliquot_A": {
            "ElementTypeName": "Aliquot_Liquid",
            "Metadata": {
                "x": 140,
                "y": 553
            },
            "Parameters": {
                "ByRow": false,
                "LiquidToAliquot": "water",
                "OutPlate": "SRWFB96_riser18",
                "Replicates": 8,
                "VolumePerAliquot": "50ul"
            }
        }
        ...
    }
```

This would seem to reduce the duplication of information within the bundle and make it more straightforward to understand.

msackman-synthace: No particular reason that I can think of. Happy to change.

msackman-synthace:
re changes to wytpe.File
As discussed with @haydnKing and @m-i-sadowski there is a problem in that a wtype.File which is a Param to an element can be modified, where it really should be readonly. So what I've done is that wtype.File is really much closer to a path now, or at the least a pointer to a file. There is now a FileManager as part of laboratory, which allows you to load from a wtype.File, or write a new file (thus returning a wtype.File). We do sensible levels of caching, which can of course we tweaked later on.

Point is that if you want to write a file out, you cannot use an existing wtype.File, you have to create a new one, plus, reading a file, whether it's an input to the workflow, or a Data output from a connected preceding element always returns a copy of the data, thus preventing modifying previous elemens. So you now write element code that looks like this:

```go
  contentA, _ := lab.FileManager.ReadAll(ParamFileA)
  contentB, _ := lab.FileManager.ReadAll(ParamFileB)
  if res, err := lab.FileManager.WriteString(element, string(contentA)+"\n"+string(contentB)); err != nil {
    lab.Error(err)
    return
  } else {
    DataFile = res
  }
```

Finally, once each element is complete, we serialize its state to disk before triggering any subsequent elements.

msackman-synthace:
NB: todo: Currently, for .an files, I test every import to see if it matches. But this will go wrong when .go imports .go. So we need to just copy over the whole of the repo, and the import testing is *only* for controlling the work list for the translation of .an to .go.

Which then raises interesting questions about how to cope with multiple imports of the same thing from different revisions - really we work with go mod here to do this properly.

msackman-synthace:
Discovered that until we switch to go.mod for antha and elements, we just cannot make "multiple imports of the same repo but at different revisions" work correctly due to the impossibility of rewriting repo-internal imports.
So, rather than try to tackle switching to go.mod at this point too, I'm just going to add additional limitations to the workflow so that importing the same repo multiple times is forbidden.

msackman-synthace:
Right, how to run some stuff:

Currently, the new command is still called `composer`. So:

```bash

cd cmd/composer
go build
```

Now, composer has 4 flags at the moment:

-outdir=/path/to/dir the directory in which to write stuff. If not given, will create a random under /tmp
-run=false (bool, default true) if set to false, will stop before actually running the workflow
-keep (bool, default false) if set to true, will not delete (tmp) sources before running the workflow
-linkedDrivers (bool, default false) if set, changes build tags when building the workflow assuming GOPATH has access to sources to instruction plugins and antha-runner and thus mixer configs do not have to have a `"Connection"` configuration parameter

So for example you can do:

```bash
./composer -outdir /tmp/foobar -linkedDrivers -keep sampleRepositories.json sampleWorkflow2.json sampleMixers.json
```

(Note this will error due to too many mixers, so edit sampleMixers.json to make sure there is only one mixer defined).

General structure of interesting changes:

* /workflow - the definition of the workflow file, with code to do json (un)marshalling, merging, and validation. Deliberately has as little mutation logic as possible: as  much as possible this should be read-only representation of the json
* /composer - this is the code that takes a workflow and prepares for compilation: generation of the workflow main, and the necessary calls to transpilation, plus checking out of all the necessary code. It also does prep work for the instructnion plugins as necessary. Finally we do the compilation of the generated code
* /laboratory - used at "element run time" i.e. once we've compiled the workflow and are actually running it. Created by the generated main, and then populated by the used elements. Is a labBuilder before the elements are started, then is a laboratory within elements (and thus large parts of wtype too) and finally labEffects once passed through to the planner.
* /inventory - simplified and made more consistent than before.

Also, have a look at the generated main, and also have a look at the transpiled elements (aliquot on "matthew" branch is the only one that actually works right now) - the generated code should appear much simpler and more obvious.

msackman-synthace:
Just FTR, documenting the current indir/outdir directories:

```
inDir/
  data/
    # calls to lab.FileManager.ReadAll() must read from in here (workflow runtime)
  workflow/
    *.json # treated as workflow fragments

outDir/ (for composer)
  src/
    # where we check out the element repositories, and do the transpilation of elements
  workflow/
    main.go # the generated main.go
    data/
      workflow.json # the merged and validated workflow
  logs.txt # logs from the composer only
  bin/
    jobId # the compiled binary for the whole workflow
    drivers/
      driverId # binary instruction plugin (iff go:// or file:// Connection string in driver config)

outDir/ (for workflow runtime)
  data/
    # calls to lab.FileManager.Write* write in here
  elements/
    # each element gets a elemName.json file in here which is the serialization of the element after run
  logs.txt # logs from the workflow runtime
  errors.txt # if any element gives an error (or panic) it will go in here
  devices/
    task-id/
      device-specific-files # files generated for the device as part of the task
```

inventory/components/components.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"github.com/antha-lang/antha/laboratory/effects/id"
+)
+
+const (
+	// WaterType is the component type of water
+	WaterType = "water"
+)
+
+type ErrComponentNotExist struct {
```
m-i-sadowski: Documentary comments on all public things please (cc... all other public things)

inventory/inventory.go
```diff
@@ -1,84 +1,39 @@
...
-
-// NewComponent returns a new component of the given type
-func NewComponent(ctx context.Context, typ string) (*wtype.Liquid, error) {
-	return fromContext(ctx).NewComponent(ctx, typ)
-}
-
-// NewPlate returns a new plate of the given type
-func NewPlate(ctx context.Context, typ string) (*wtype.Plate, error) {
-	return fromContext(ctx).NewPlate(ctx, typ)
+type Inventory struct {
```
m-i-sadowski: The reuse of Inventory between this and its subpackages creates the expectation that they will follow the same interface or be otherwise similar, it's a little confusing that they aren't.

I'd suggest either this name or the ones below should change to reflect the slight difference. This could be InventorySet? Inventories? Quartermaster? InventoryManager?

inventory/tipboxes/consts.go
```diff
@@ -0,0 +1,8 @@
+package tipboxes
```
m-i-sadowski: it's a bit odd to me that this package has a consts.go and none of the others do... I would expect most if not all of them should have something similar

inventory/inventory.go
```diff
@@ -1,84 +1,39 @@
...
+	Components *components.Inventory
+	TipWastes  *tipwastes.Inventory
+	TipBoxes   *tipboxes.Inventory
+	PlateTypes *plates.Inventory
```
m-i-sadowski: Why is this "type of thing" while everything else is "thing"?

I suggest they should *all* be "type of thing"

inventory/inventory.go
msackman-synthace: Hmm, I think I would prefer to just flatten the whole thing into effects - I'll need to check but I doubt there's a whole lotta reason to have these grouped externally to labEffects.

(somewhat tangentially: In any situation where you have several of these imported in the same file, you would always have to prefix these by package name anyway so it should always be clear what inventory you're referring to. Repeating the package name in methods or type names is generally considered not good go style, i.e. generally `foo.New()` and `bar.New()` is preferred to `foo.NewFoo()` and `bar.NewBar()`)

inventory/inventory.go
msackman-synthace: Hah, yeah, I'd love to do that. Look at the diff for how painful PlateTypeName became. We totally need to make that happen but:
1. there's a question of whether that's really a great idea given it impacts everything right the way out to elements themselves
2. making it not horrendously painful may well require a much redesigned system for interacting with any of these labwares, not based off strings

I think at this point if you don't want to suffer the inconsistency for a month or so, I would prefer to undo the PlateTypeName change and plateTypes as a whole rather than push out the same change to all the other labware.

workflow/config.go
```diff
@@ -0,0 +1,212 @@
+package workflow
```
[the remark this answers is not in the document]
msackman-synthace: Right, let me have another go at trying to answer this.

I have chosen to try to make everything which is accessible from the workflow.Workflow object self contained and "leaf" like - i.e. importing workflow/ does not end up importing say, target/. Why?

The Workflow struct, and everything reachable from it, is directly concerned with just a few things: (de)serialisation, merging, and high-level validation. That is all.

A mixer config struct, as you will see from the target/mixer code is concerned with much more: beyond validation, also state machines around connecting to the instruction plugin, managing the lifecycle of the LHProperties object, plus stuff like flattening out default options etc.

So I have tried to keep to one-struct = one-purpose, or at least a minimum set of related purposes. The actual mixer doesn't give a damn how json serialisation happens, if at all. Similarly, a reader (human, eg ui dev) of the workflow struct doesn't care how plate type definitions are interpreted by each mixer.

Additionally, workflow (de)serialisation, merging, and validation is something that happens in a lot of places: basically all the new commands do it (composer, elements, migrate) in some cases several times. They need to be confident that there's no complex life-cycle stuff going on where "oh it's not safe to serialise at this point in time because you've got a connection open to Gilson IP and so that value in that field is temporarily invalid" etc etc. Keeping an API struct entirely separate (and essentially readonly) from business logic helps a lot with this.

Perhaps finally, one nasty aspect of the protobuf definitions in microservices is that they're spread all over the place, which makes reuse and importing very painful. Having your complete set of public apis defined just in one place makes life easier for consumers of that API IME.

So, for all those reasons, I have made a split in mixer config: the static config of the workflow struct is very deliberately separate from the dynamic state within target/mixer/

inventory/cache/plateCache/plateCache.go
```diff
@@ -1,86 +1,83 @@
...
-	"github.com/antha-lang/antha/inventory/cache"
+	"github.com/antha-lang/antha/inventory/plates"
 )
 
-type plateCache struct {
-	platesByType    map[string][]*wtype.Plate
-	platesFromCache map[string]bool
-}
+const (
+	batchSize = 8
```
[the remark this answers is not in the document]
msackman-synthace: Actually I'll remove. I was previously doing batch creation and then during test debitrotting found it broke a test. I have no idea what the intended semantics are but batchSize isn't helping at all at this point.

inventory/inventory.go
```diff
@@ -1,84 +1,39 @@
...
 
-// XXXNewPlates is a transitional call that will be removed once the planner
-// requires input plate types to be explicitly set.
-func XXXNewPlates(ctx context.Context) ([]*wtype.Plate, error) {
-	inv := GetInventory(ctx)
-	xInv, ok := inv.(hasXXXGetPlates)
-	if !ok {
-		return nil, errCannotListPlates
+func (inv *Inventory) LoadForWorkflow(wf *workflow.Workflow) {
+	// TODO: discuss this: not sure if we want to do this based off
```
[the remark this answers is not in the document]
msackman-synthace: To expand on that a bit: what is considered a test? For example, the test workflows checked into elemenst-westeros: these are tests, but I don't think they should have the test plate type inventory loaded. So is it really only "internal" tests somehow that get the test inventory? What about go tests that are committed into the elemenst-westeros repo?

inventory/components/components.go
msackman-synthace: As discussed on slack: nope. If we currently have a linter that demands that then we will turn that off. We should be documenting where necessary: where ever an api has no alternative to be non-obvious or requires a substantial investment of time and effort on behalf of the user to understand.

inventory/tipboxes/consts.go
msackman-synthace: Agreed. I'll nix the file.

inventory/inventory.go
msackman-synthace: (discussed further on slack)

inventory/inventory.go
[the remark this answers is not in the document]
msackman-synthace: This has been discussed in person; conclusion is that for now, if the plate types in the workflow is empty, then load the baked in library.

inventory/cache/plateCache/plateCache.go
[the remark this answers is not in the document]
msackman-synthace: This has been removed.

inventory/inventory.go
msackman-synthace: Discussed elsewhere - solution for now is to just rename this field to `Plates`.

instructions/highlevel.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"github.com/antha-lang/antha/antha/anthalib/wunit"
+)
+
+// Note: these are instructions at the level issued by intrinsics
+// (i.e. wrapped in CommandInst (see trace.go)), and are sometimes
+// called "high-level" instructions. These are ultimately the inputs
+// to calls to Device.Compile. Do confuse these with the Inst
+// interface in device.go - those are for the results of
```
mabn: This comment is outdated, there is no `Inst` in device.go, it seems to be in lowlevel.go now.

msackman-synthace:
On Tue, May 07, 2019 at 04:50:28AM -0700, Marcin Biegan wrote:
> This comment is outdated, there is no `Inst` in device.go, it seems to be in lowlevel.go now.

fixed, thank you.

PR #929, opened 2018-12-18
Antha 2373/test time estimate

chr1sgrant:


m-i-sadowski:
I weakly prefer the use of the Go time type over the Antha one here, however I'm not going to require it

PR #930, opened 2018-12-18
Improve driver.CommandStatus

haydnKing:
driver.CommandStatus is the struct by which drivers return the status of commands.

Previously this was quite messy as information was duplicated (e.g. do you check `status.OK` or `status.Errorcode == OK`), drivers constructed the struct directly in a myriad of different ways, and there was no stringer implementation.

This PR simplifies the struct and adds convenience methods for creating CommandStatus with the different codes (so you don't have to remember is it driver.ERR, driver.Err, driver.Error?), as well as the CommandStatus.String() and CommandStatus.Error() methods which significantly reduce boilerplate around the codebase.

As usual, there are driver changes associated with this, which should appear below.


microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -1281,12 +1270,7 @@ func (ins *BlowoutInstruction) OutputTo(lhdriver LiquidhandlingDriver) error {
 	for i := 0; i < ins.Multi; i++ {
 		bo[i] = true
 	}
-	ret := driver.Dispense(volumes, bo, ins.Head, ins.Multi, ins.Plt, ins.What, ins.LLF)
-	if !ret.OK {
-		return fmt.Errorf(" %d : %s", ret.Errorcode, ret.Msg)
-	}
-
-	return nil
+	return driver.Dispense(volumes, bo, ins.Head, ins.Multi, ins.Plt, ins.What, ins.LLF).Error()
```
m-i-sadowski: This usage looks a tiny bit odd to me since I'd expect to call .Error() on an object I already know to be an error. i.e. Go convention has err.Error() returning a string (hence can't be nil) - I think it's better to avoid any confusion over this here.

haydnKing:
I tested by running a bundle using the PipetMaxDriver with no problem, and have checked that all of the other drivers build successfully

haydnKing:
should be good now :)

m-i-sadowski:
Approved. I'll merge this and the driver changes in one fell swoop

haydnKing:
seems the logger got deleted, so I'll update accordingly

PR #934, opened 2019-01-07
remove unused JSON file

haydnKing:
Found this file while grepping for references to the now defunct logger, seems to be unused.

Couldn't figure out the purpose of this - seems like it used to have an XML counterpart which was removed a while ago?


haydnKing:
removing unused PR...

PR #935, opened 2019-01-08
Refactor user layout preferences

haydnKing:
Layout preferences determine which types of objects are to be placed in which deck location, and in which order.

There are two sources of layout preferences: the device plugin and the user. Previously, these sources are merged according to complex, device specific rules which are hardcoded into antha, meaning that devices can't be added without adding to this hardcode.

This change introduces a simple device independent method for merging the user positions with the driver positions:

- if no user preferences are provided, the plugin provided list is used
- otherwise, user preferences are accepted in order if they are also specified in the device's list, otherwise they are ignored.

This allows the device to specify which locations can possibly hold a specific class of labware, and the user to specify which locations should hold a specific class of labware. If the user specifies a position which doesn't exist, it is silently ignored, such that specifying "position_1" to a hamilton driver has no effect.

There are some changes to each of the device plugins and a small change to antha-runner, PRs for each will appear below


haydnKing:
PS. I tested this by running using the PipetMaxDriver and also by running antha-tester :)

microArch/driver/liquidhandling/layoutopt.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+// String list the addresses wrapped in inverted commas
+func (a Addresses) String() string {
+	return fmt.Sprintf("\"%s\"", strings.Join([]string(a), "\", \""))
+}
+
+// LayoutOpt describes the options available for layout of various types of objects
+// on a liquidhandler deck
+type LayoutOpt map[ObjectCategory]Addresses
```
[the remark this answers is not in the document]
msackman-synthace: Or, arguably better:

```
type PermittedObjectTypes struct {
	Tipboxes:  bool
	Inputs:    bool
	Outputs:   bool
	Tipwastes: bool
	Wastes:    bool
	Washes:    bool
}

type LayoutOpt map[Address]PermittedObjectTypes
```

I just feel that might make things super obvious and simple.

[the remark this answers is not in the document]
msackman-synthace: (merging could then just become the `||` for each field pair...)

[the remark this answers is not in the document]
haydnKing: Ah, but the ordering is important. The way this gets used is that when the compiler decides it needs to introduce a new tipbox it looks through the list of possible tipbox locations and puts one in the first empty place it can find - ditto for inputs/outputs etc.
We'll hopefully change this in the future - by actually search the space of possible deck layouts and using the physical simulator to choose between possible layouts so the user doesn't have to specify an order at all

microArch/driver/liquidhandling/layoutopt.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, address := range a {
+		ret[address] = true
+	}
+	return ret
+}
+
+// Filter returns a new address slice containing the addresses in b that also appear in a
+func (a Addresses) Filter(b Addresses) Addresses {
+	inA := a.Map()
+	ret := make(Addresses, len(b))
```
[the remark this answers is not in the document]
haydnKing: :headdesk:

microArch/driver/liquidhandling/layoutopt.go
```diff
@@ -0,0 +1,137 @@
+package liquidhandling
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"strings"
+)
+
+// ObjectCategory enumerate the different types of objects which can appear on deck
+type ObjectCategory int
```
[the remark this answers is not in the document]
haydnKing: Yeah, that's a bit better

microArch/driver/liquidhandling/layoutopt.go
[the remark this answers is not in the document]
msackman-synthace: Ok, this makes more sense then. I'm trying to think about what is most intuitive for the user to specify here. Because if you had the above struct, you could replace the bools with uints and have the uints represent weights (or priorities). So then it's a choice between:

```
Tipboxes: [position_1, position_5],
Inputs: [position_1, position_2],
...
```

vs

```
position_1: {Tipboxes: 10, Inputs: 3}
position_2: {Inputs: 2}
position_5: {Tipboxes: 9}
...
```

I think the latter is probably more flexible, but does seem harder to understand to me (here the numbers are priorities so 0 means disabled/forbidden, and the higher the number, the higher the priority).

So I think then what you have does then make sense, but I don't see any reason for the const and iota stuff - I'd just write a struct with some fields for the different types.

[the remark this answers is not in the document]
haydnKing: Hmm, yeah I think I may have over-estimated how often (whether!) we'd actually need to introduce new type of object...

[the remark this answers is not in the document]
haydnKing: agree it's simpler overall to have struct members, I'll fix it

microArch/driver/liquidhandling/layoutopt.go
```diff
@@ -0,0 +1,137 @@
...
+
+	ret := make([]ObjectCategory, 0, len(seen))
+	for cat := range seen {
+		ret = append(ret, cat)
+	}
+	return ret
+}
+
+// Merge combine the driver reported layout preferences with the users'
+func (lo LayoutOpt) Merge(user LayoutOpt) error {
```
msackman-synthace: Having reread the PR description, I think this should be renamed Intersect too. Merge is too ambiguous - merging 2 sets of user prefs would likely have quite different semantics (possibly union) to merging user prefs with device constraints.

haydnKing:
Think this is ready for another go :)

microArch/driver/liquidhandling/layoutopt.go
```diff
@@ -0,0 +1,107 @@
...
+		Tipwastes: lo.Tipwastes,
+		Wastes:    lo.Wastes,
+		Washes:    lo.Washes,
+	}
+}
+
+// ApplyUserPreferences combine the user supplied preferences with the driver plugin supplied rules
+// An error is returned if the user requests an object be placed in a location that the driver
+// has not allowed
+func (lo *LayoutOpt) ApplyUserPreferences(user *LayoutOpt) (*LayoutOpt, error) {
```
msackman-synthace: (nit picking) I would add to this comment:
> I.e. the user preferences must be a subset of the driver rules, but can specify any order of the acceptable addresses.

microArch/driver/liquidhandling/layoutopt.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+		if len(invalid) > 0 {
+			return nil, errors.New(invalid.String())
+		}
+
+		if len(user) > 0 {
+			return user.Dup(), nil
+		} else {
+			return driver.Dup(), nil
+		}
```
msackman-synthace: Ooooo, this is non obvious. As a user, how do I prevent a particular address being used for a purpose that the driver permits? Surely that's a valid use case?

microArch/driver/liquidhandling/layoutopt.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	if tipboxes, err := override(lo.Tipboxes, user.Tipboxes); err != nil {
```
msackman-synthace: You might prefer this:

```go
result := &LayoutOpt{}
errs := utils.ErrorSlice{
  override(&result.Tipboxes, lo.Tipboxes, user.Tipboxes),
  override(&result.Inputs, lo.Inputs, user.Inputs),
  override(&result.Outputs, lo.Outputs, user.Outputs),
  override(&result.Tipwastes, lo.Tipwastes, user.Tipwastes),
  override(&result.Wastes, lo.Wastes, user.Wastes),
  override(&result.Washes, lo.Washes, user.Washes),
}
return result, errs.Nub()
```

microArch/driver/liquidhandling/layoutopt.go
haydnKing: say the driver permits `["a", "b", "c"]`, then to exclude `"b"` for example you'd have to pass it `["a", "c"]`. 
Not exactly the easiest right now (since you need to know the defaults) but the place I'm headed for all this is that the device defaults are loaded into the UI then we have some sensible UX that allows the user to reorder / remove them as they wish

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -240,33 +240,27 @@ func (self *VirtualLiquidHandler) AddErrorf(format string, a ...interface{}) {
...
+	} else if err := check_prop(props.Preferences.Inputs); err != nil {
+		return errors.WithMessage(err, "in input preferences")
+	} else if err := check_prop(props.Preferences.Outputs); err != nil {
+		return errors.WithMessage(err, "in output preferences")
+	} else if err := check_prop(props.Preferences.Tipwastes); err != nil {
+		return errors.WithMessage(err, "in tipwaste preferences")
+	} else if err := check_prop(props.Preferences.Wastes); err != nil {
+		return errors.WithMessage(err, "in waste preferences")
+	} else if err := check_prop(props.Preferences.Washes); err != nil {
+		return errors.WithMessage(err, "in wash preferences")
```
msackman-synthace: you could do the same utils.ErrorSlice trick here and push the errors.WithMessage into the check_prop (you'll need to supply the string too). Would save some loc...

microArch/driver/liquidhandling/layoutopt.go
msackman-synthace: `antha-lang/utils`

microArch/driver/liquidhandling/layoutopt.go
msackman-synthace: but if I want to say "there are *no* addresses for, say, washes" I don't think I can do that?

haydnKing: Hmm, so the only reason I can think that someone might want to do that is so that any attempt to do any tip washing results in a run out of deck space error. I don't really see why someone would want to ban tip washing in this way, though I agree that if a user deliberately left Tipbox preferences blank for funsies they'd be surprised to see tipboxes being placed on the deck...

I think maybe this is something we should revisit once the front end can actually access the list of allowable addresses, since we can then remove the "empty list means default" convention. 

(I also wonder if we really use washes and wastes, but for another time, I think)

msackman-synthace: That's fine - I was wondering whether I was inventing an impossible/nonsensical case. Might be worth adding to the comment nevertheless just to be clear.

haydnKing: I'll add a summary of this discussion so future me isn't confused

microArch/driver/liquidhandling/lhproperties.go
```diff
@@ -506,7 +372,7 @@ func (lhp *LHProperties) AddTipBox(tipbox *wtype.LHTipbox) error {
 	return wtype.LHError(wtype.LH_ERR_NO_DECK_SPACE, "Trying to add tip box")
 }
 func (lhp *LHProperties) AddTipBoxTo(pos string, tipbox *wtype.LHTipbox) bool {
-	if lhp.PosLookup[pos] != "" {
+	if !lhp.IsEmpty(pos) {
```
msackman-synthace: and here we have another name...!

haydnKing:
certainly agree that it was unclear when a variable called `pos` was actually an LHPosition or a string... actually just found a case of shadowing
LHPosition is a bit of a stub at the moment. I have another ticket (ANTHA-2368) to expand it to include the size of the position, which could also encompass some more of this clarification

microArch/driver/liquidhandling/lhproperties.go
```diff
@@ -275,17 +185,16 @@ func (lhp *LHProperties) DupKeepIDs() *LHProperties {
...
 
 func (lhp *LHProperties) dup(keepIDs bool) *LHProperties {
-	lo := make(map[string]wtype.Coordinates, len(lhp.Layout))
-	for k, v := range lhp.Layout {
-		lo[k] = v
+	pos := make(map[string]*wtype.LHPosition, len(lhp.Positions))
+	for k, v := range lhp.Positions {
+		// be sure to copy the data
+		w := *v
+		pos[k] = &w
```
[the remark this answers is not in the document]
msackman-synthace: (I think long term, it's probably worth explicitly modelling every event and then having the ability to ask questions of a sequence of events, like "what is in deck position X right now?". So then the whole thing is built as an initial state and then a sequence of modifications. But that's for another day :) )

haydnKing:
This goes in, then there are PRs in each of the driver/plugin repos, mainly just find replace for the slightly changed structure of LHProperties
Once those are in there's a branch in antha-runner with a similar cosmetic change that'll need `dep ensure --update`ing which can then be merged in

haydnKing:
There's one possibly meaningful change in the pipetmax driver, which is this: https://github.com/Synthace/PipetMaxDriver/pull/59/files#diff-bec4d3b8eb97a20a20d09f34a52d24d0R103
which expanding the list of possible places where tipboxes can be placed

msackman-synthace:
> There's one possibly meaningful change in the pipetmax driver, which is this: https://github.com/Synthace/PipetMaxDriver/pull/59/files#diff-bec4d3b8eb97a20a20d09f34a52d24d0R103
> which expanding the list of possible places where tipboxes can be placed

What?! My read is it expands the list of places where Inputs and Outputs can be placed. Tipboxes stays the same, unless there's something really wrong with my eyes...?

msackman-synthace:
Ok, everything else looks ok to me, though I've not done any testing on this myself. Handing over to Sid on this one. :)

haydnKing:
> > There's one possibly meaningful change in the pipetmax driver, which is this: https://github.com/Synthace/PipetMaxDriver/pull/59/files#diff-bec4d3b8eb97a20a20d09f34a52d24d0R103
> > which expanding the list of possible places where tipboxes can be placed
> 
> What?! My read is it expands the list of places where Inputs and Outputs can be placed. Tipboxes stays the same, unless there's something really wrong with my eyes...?

oops, you're right - it's my eyes that have something wrong with them

PR #936, opened 2019-01-08
add % v/v as unit

chr1sgrant:


haydnKing:
seems fine but could you add a test for the new unit? I'm a little concerned about the space!

chr1sgrant:
Good point, I've added a test now which demonstrates it behaves as intended

PR #940, opened 2019-01-16
Driver Plugins define Tip Types

haydnKing:
Previously, Tip and Tipwaste definitions were made centrally in `testinventory`. This isn't right, because tips and Tipwastes are specific to the liquidhandler in question, and adding a new liquidhandler is always going to involve defining what tips and tipwastes it uses.

This change moves the canonical definitions of tips and tipwastes out of `inventory` and into the plugin, via the `TipFactory` object, which the planner uses to initialise tipwaste, query tip properties, and create new tipboxes as necessary.

`testinventory` still contains tipbox and tipwaste definitions, as these are required within antha for testing purposes. Many _many_ tests have thus been rewritten to use `testinventory` directly, rather than use the previous dance of putting testinventory into a context object then later retrieving it. For efficiency, only one `testinventory` is initialised per testing run, which is read only and also guarded by a mutex. Most of the actual changes here are to do with this.

This all appears to be working with the PMDriver on the command line, still working on making changes to the other drivers and testing via `antha-runner`.


haydnKing:
Need to hold off on this, getting a bunch of physical simulation and no deck space errors relating to tipboxes from `antha-tester`

haydnKing:
Ok having mulled and chatted to Sid, I think I'm going to conclude that this does indeed need splitting into two separate pieces of work - refactoring tests and moving tip definitions. Expect PRs sometime next week (i'm off Friday and Monday)

PR #941, opened 2019-01-16
Give inst.Output a concrete type as it's only used in one place.

msackman-synthace:
To do this has required moving Inst and Device from target to ast in order to avoid an import cycle.
In fact, I rather suspect that the reason ast.Command.Output was interface{} was solely to break the import cycle, which is a bit crazy.

I'm trying to keep these PRs as short as possible. So this PR doesn't actually make trace serializable, but it's a step in that direction.

Test plan:
antha> go build ./...
works.

### DIFF
--- a/target/qpcrdevice/qpcrdevice.go
+++ b/target/qpcrdevice/qpcrdevice.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+
 	"github.com/antha-lang/antha/ast"
 	"github.com/antha-lang/antha/driver"
 	framework "github.com/antha-lang/antha/driver/antha_framework_v1"
 	quantstudio "github.com/antha-lang/antha/driver/antha_quantstudio_v1"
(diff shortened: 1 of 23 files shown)


target/qpcrdevice/qpcrdevice.go
```diff
@@ -4,21 +4,22 @@ import (
...
 	"fmt"
+	"os"
+
 	"github.com/antha-lang/antha/ast"
 	"github.com/antha-lang/antha/driver"
 	framework "github.com/antha-lang/antha/driver/antha_framework_v1"
 	quantstudio "github.com/antha-lang/antha/driver/antha_quantstudio_v1"
 	"github.com/antha-lang/antha/target"
 	"github.com/golang/protobuf/proto"
-	"os"
```
haydnKing: ah! the legendary antha-os!

msackman-synthace:
yeah, I'm pretty sure they should be the same package, and the packages should be renamed, along with they types, and many of the chars within...

PR #945, opened 2019-01-21
Align mods 1

mcouch-synthace:
An addition and a minor bug fix.

1. Addition
The alignment score is now returned in a new Score field of the antha Alignment struct. A method has also been added to the Result struct to report this score.

2. Bug fix
The positions arrays in the Alignment struct record the query/template positions throughout an aligned region. The code was pulling positions from the ungapped side over a gap.  In the fix, the code freezes the position in the current sequence over the gap. Ultimately it might be better to set the position to 0 or -1 across a gap to make this clearer.


x-anderson:
Both changes look good - and being able to call the Score as per BioGo is particularly useful 👍

PR #946, opened 2019-01-21
add tip estimate to bundle tests

chr1sgrant:
Adding this because it can be possible to unexpectedly use many more tips without changing Time estimate; 

1 question is whether the order of the TipsUsed field of LHRequest ([]wtype.TipEstimate) is always consistent?

Cuurrently, this test code assumes this is the case and will fail if not.

### DIFF
--- a/workflowtest/testoutputs.go
+++ b/workflowtest/testoutputs.go
@@ -11,6 +11,7 @@ import (
 	"github.com/antha-lang/antha/execute"
 	"github.com/antha-lang/antha/microArch/driver/liquidhandling"
 	"github.com/antha-lang/antha/target"
+	"github.com/go-test/deep"
 )
 
 // A TestOpt is an option for running a test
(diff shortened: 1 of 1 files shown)


chr1sgrant:
looks like the order of TipsUsed []TipEstimate is liable to change so I would either need to sort both lists or, preferably, we refactor the TipsUsed to a map[string]TipEstimate

chr1sgrant:
Sorting is sorted now; next question:

Is the tip estimate deterministic enough to base passing only on exact matches?

chr1sgrant:
example error message returned running antha-tester:

```
tip estimate differences detected between expected results and simulation results: 
 slice[0].NTips: 29 != 150
slice[0].NTipBoxes: 1 != 2
```

chr1sgrant:
updated error message to contain more information now: 

```
Expected tip estimate: 
 [
	{
		"TipType": "D200 Tip Rack (PIPETMAX 8x200)",
		"NTips": 29,
		"NTipBoxes": 1
	}
] 
 Got 
 [
	{
		"TipType": "D200 Tip Rack (PIPETMAX 8x200)",
		"NTips": 150,
		"NTipBoxes": 2
	}
] 
 Differences detected: 
 slice[0].NTips: 29 != 150
slice[0].NTipBoxes: 1 != 2

```

m-i-sadowski:
I'll have to review this properly, seems reasonable but I'm confused by the first assertion. Can tip use really change significantly without changing time estimate?

chr1sgrant:
What do we want to do here?
My biggest question would be around non determinacy of the tip estimate results

PR #951, opened 2019-01-22
LHWell .Protect is not used.

msackman-synthace:
Context: Plates seem odd. They're in wtype and thus seem available to elements, but are sorta managed through sampletracker which is in microArch...
I was trying to figure out if we can drop all pointers to Liquids and just pass around maps from ids to objects, which would at least move us away from the weirdness of Dup with IDs (I found some elements that call Dup()...)
Which then made me think of Plates, which contain Wells, which contain Liquids... so I was looking at Wells and came across the protect stuff.

### DIFF
--- a/antha/anthalib/wtype/lhwell.go
+++ b/antha/anthalib/wtype/lhwell.go
@@ -204,35 +204,6 @@ Extra     : %v,
 	)
 }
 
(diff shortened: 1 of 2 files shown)


msackman-synthace:
∪ {loved}

PR #953, opened 2019-01-23
Add blast call with parameters

mcouch-synthace:
Provides for setting the blast search parameters.


mcouch-synthace:
This package needs some tests. Will write some and submit new.

PR #954, opened 2019-01-24
Add parametrized blast call

mcouch-synthace:
Existing functions pass hard coded parameters via a global parameters struct.  This struct is not exported.

In many situations it is necessary to pass parameters e.g. the target database, entrez filter by organism name.

This could be achieved by calling lower level code.  However there is important polling functionality  in `SimpleBlast` that needs to be copied or re-implemented if making a request via lower level code.

This PR adds a new low level function `BLAST` which allows parameters to be passed and contains the polling functionality.  All existing functions now execute via this one.


antha/AnthaStandardLibrary/Packages/sequences/blast/blast.go
```diff
@@ -266,6 +266,7 @@ func AllExactMatches(hits []blast.Hit) (exactmatches []blast.Hit, summary string
 	return
 }
 
+// MegaBlastP performs a blastn call against the nr database with Megablast default parameters
 func MegaBlastP(query string) (hits []blast.Hit, err error) {
```
chr1sgrant: Is this comment correct?

looks like the function calls blastp

antha/AnthaStandardLibrary/Packages/sequences/blast/blast.go
```diff
@@ -295,8 +297,14 @@ func MegaBlastN(query string) (hits []blast.Hit, err error) {
...
 }
 
-func SimpleBlast(query string) (o *blast.Output, err error) {
-	r, err := blast.Put(query, &putparams, tool, email)
+// SimpleBlast performs a blast call with pre-set query parameters
+func SimpleBlast(query string) (*blast.Output, error) {
+	return BLAST(query, &putparams)
+}
+
+// BLAST performs a blast call for specified query parameters
```
chr1sgrant: Worth explaining valid/some common PutParameter settings; could link to their documentation here: https://github.com/biogo/ncbi/blob/master/blast/blast.go

I was going to suggest linking to the html link there for what supported values can be used but the link is now dead unfortunately.

mcouch-synthace: Yep.
It looks as if BLAST API parameters may now be here https://ncbi.github.io/blast-cloud/dev/api.html
But I see some of the parameters supported by biogo are not (or are no longer) supported by the BLAST API itself e.g. `DB_GENETIC_CODE`, `PERC_IDENT`.
I'll dig a bit further.

antha/AnthaStandardLibrary/Packages/sequences/blast/blast.go
mcouch-synthace: Yes.  My comment is wrong.

mcouch-synthace: Looking into this a bit more, I'm unsure of the parameters being passed in both `MegaBlastP` and `MegaBlastN`.  For example, I understood that `megablast` is an alternative program to `blastn` optimised for rapid identification of highly similar matches between DNA query and DNA targets.  It doesn't seem that megablast would be applicable in a `blastp` search (protein query, protein target database).  Also, the parameter `MEGABLAST` is not one of the supported arguments according to documentation here  https://ncbi.github.io/blast-cloud/dev/api.html.  So there may be a separate piece of work to understand the intention and settings in those existing functions.  Until then it might be better to leave them uncommented per master.

antha/AnthaStandardLibrary/Packages/sequences/blast/blast.go
mcouch-synthace: Done.

mcouch-synthace:
Hey @chr1sgrant is this OK to go or is more work needed?

PR #955, opened 2019-01-24
Genbank writer

mcouch-synthace:
Adds a new method in `Packages/export` to export annotated `DNASequence`(s) in a basic Genbank format.

This is intended to support more user-friendly output from elements such as `Construct_Assembly_Multi` which currently exports only an unannotated Fasta file.  Genbank format will allow the inserted parts to be annotated and viewed in e.g. Snapgene viewer.

The test checks that the output Genbank file can be parsed back to an annotated `DNASequence` using the Genbank parser in `Packages/sequences/parse/genbank`.

### DIFF
--- a/antha/AnthaStandardLibrary/Packages/export/export.go
+++ b/antha/AnthaStandardLibrary/Packages/export/export.go
@@ -34,6 +34,8 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


mcouch-synthace:
Case of a feature spanning the rep origin needs further checking, will look into it.

PR #956, opened 2019-01-25
Genbank writer

mcouch-synthace:
Adds a new method in `Packages/export` to export annotated `DNASequence`(s) in a basic Genbank format.

This is intended to support more user-friendly output from elements such as `Construct_Assembly_Multi` which currently exports only an unannotated Fasta file. Genbank format will allow the inserted parts to be annotated and viewed in e.g. Snapgene viewer.

The test checks that the output Genbank file can be parsed back to an annotated `DNASequence` using the Genbank parser in `Packages/sequences/parse/genbank`.

The parser doesn't handle Genbank features with join locations e.g.
```
join(6999..7010,1..89)
complement(join(6999..7010,1..89))
```
(see [here](https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/sequences/parse/genbank/genbank_parser.go#L228) and [here](https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/sequences/parse/genbank/genbank_parser.go#L284)). These cases are handled by the writer but have not been tested with parser.


chr1sgrant:
In terms of the joins, at the moment it seems that if joins are present then we terminate; perhaps it's preferable to continue and return a warning at the end to explain the loss of annotation; but that may be out of scope for this PR.

antha/AnthaStandardLibrary/Packages/export/export.go
```diff
@@ -305,6 +307,152 @@ func FastaSerialfromMultipleAssemblies(dirname string, multipleassemblyparameter
...
+						ret = fmt.Sprintf("%d..%d", feat.Start(), feat.End())
+					}
+				}
+			}
+			return ret
+		},
+		"origin": func(seq wtype.DNASequence) []string {
+			bases := seq.Sequence()
+			lines := []string{}
+			for pos := 0; pos < len(bases); pos += 60 {
```
chr1sgrant: can we pull out 60 and 10 as constants so it's clearer what they mean?

chr1sgrant:
Tested with the output of Assembly design elements and works as expected; highlights the need to refactor assembly simulator code to use DNASequence type directly rather than converting to an intermediate DiggestedFragment format and losing any annotations.

antha/AnthaStandardLibrary/Packages/export/export.go
mcouch-synthace: Done

PR #957, opened 2019-01-25
Fix/volume error250119

m-i-sadowski:
This PR aims to fix a nasty bug with multichanneling, which manifested itself as some autoallocated input plates being requested with excessive volumes in some wells. 

What was actually happening is that in some cases zero volumes for certain transfers were being rewritten as nonzero volumes as a side effect of duplicating the initial TransferParameters when two sequential, but slightly different, multichannel operations occurred

Unfortunately this now results in increased time estimates for a few bundles over the permitted 10% mark. However I am pretty sure that these were actually not being correctly generated (eek!) on which basis I am going to generate an MR in gitlab to regenerate these time estimates. EDIT: difficult to say whether the instructions were correct here but since the result cache is the same before and after I suspect they were, the time variation is actually happening on master as well.

Since it's related, this PR also contains a minor cosmetic detail to prevent users seeing volumes which are (somewhat oddly) over the well maximum by small amounts (typically 0.5 - 2.0 ul) owing to issues with carry volumes. Carry volumes are only an approximation to guard against wells being underfilled so it makes no sense to display these where they have such nonsensical values. 

Apart from timing issues in 3 bundles these pass antha-tester OK but I've not added any regression tests, which might be sensible.


m-i-sadowski:
Update: it seems that time estimates are actually varying randomly and are not the result of this change.

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -409,6 +409,25 @@ func (ins *TransferInstruction) ChooseChannels(prms *LHProperties) {
 	}
 }
 
+func nonContiguous(vols VolumeSet) bool {
```
haydnKing: would prefer `VolumeSet.NonContiguous` plus a comment

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -409,6 +409,25 @@ func (ins *TransferInstruction) ChooseChannels(prms *LHProperties) {
...
+	seenNonZero := false
+	for _, v := range vols {
+		if v.IsPositive() {
+			if seenZero {
+				return true
+			}
+			seenNonZero = true
+		} else {
+			if seenNonZero {
+				seenZero = true
```
haydnKing: maybe rename `seenZero` -> `zeroAfterPositive`? a bit of a clunky name but it took me a few minutes to convince myself that [0,1,0] returns false

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -487,6 +518,13 @@ func (ins *TransferInstruction) Generate(ctx context.Context, policy *wtype.LHPo
 			ins.Transfers[set].AddTVolumes(vols)
 
 			mci.Multi = len(vols)
+
+			// the following call trims off trailing zero volume transfers (either end)
+			// this is necessary if the first transfer exhausts one or more terminal wells
+			if !mci.Prms.Independent {
+				tp = tp.RemoveInitialBlanks()
```
haydnKing: If this removes zeros from the end as well as per the comment it should really be renamed to e.g. `TrimEmpty()`

microArch/driver/liquidhandling/transferparams.go
```diff
@@ -33,7 +33,7 @@ func (tp TransferParams) ToString() string {
 }
 
 func (tp TransferParams) Zero() bool {
-	return tp.What == ""
+	return tp.What == "" || tp.Volume.IsZero()
```
haydnKing: So we have `TransferParams.Zero()` and `Volume.IsZero()`, personally I prefer `IsZero` since it's clear you're asking the question not setting to zero, but either-way would be nice to make this consistent

microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -479,6 +479,10 @@ func (this *Liquidhandler) revise_volumes(rq *LHRequest) error {
 			if well.IsAutoallocated() {
 				vol.Add(well.ResidualVolume())
 
+				if vol.GreaterThan(well.MaxVolume()) {
+					vol = well.MaxVolume()
+				}
```
haydnKing: Hmm, I get that this should be fine, but it does rather give me the heebejeebies since it would cover up the error we've just found, right?
How about
```
// don't exceed the well maximum by a trivial amount
if v, max := wunit.AddVolumes(vol, well.ResidualVolume()), well.MaxVolume(); v.GreaterThan(max) && v.Subtract(max).LessThan(well.ResidualVolume()) {
  vol = max
} else {
  vol = v
}
```

microArch/driver/liquidhandling/transferinstruction.go
m-i-sadowski: also changed seenNonZero --> seenPositive

microArch/driver/liquidhandling/transferinstruction.go
m-i-sadowski: It was actually worse than that: it removed all empties. I've duplicated the function and made a version which actually only removes initial blanks

microArch/scheduler/liquidhandling/liquidhandler.go
m-i-sadowski: I'm a tiny bit uneasy still since residual volumes can be large
However it basically does fit with the logic we're using so I've implemented it in this way, however instead of expecting the user to detect the excessive case I have explicitly returned an error if the volume exceeds the threshold, which seems like a much better choice.

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -487,6 +499,13 @@ func (ins *TransferInstruction) Generate(ctx context.Context, policy *wtype.LHPo
 			ins.Transfers[set].AddTVolumes(vols)
 
 			mci.Multi = len(vols)
+
+			// the following call trims off trailing zero volume transfers (either end)
+			// this is necessary if the first transfer exhausts one or more terminal wells
+			if !mci.Prms.Independent {
+				tp = tp.RemoveBlanks()
+			}
```
haydnKing: Ok, so we know there aren't any "internal" zeros because of the NonContiguous check above, so RemoveBlanks is appropriate... can you update the comment to include that?

antha/anthalib/wunit/wunit.go
```diff
@@ -272,6 +272,9 @@ func (cm *ConcreteMeasurement) IsZero() bool {
 
 // IsPositive true if the measurement is positive by more than a very small delta
 func (cm *ConcreteMeasurement) IsPositive() bool {
+	if cm == nil {
+		return false
+	}
```
mkonicek: There is a function `IsNegative` below which doesn't take `nil` into account.

If you make `IsNegative` take `nil` into account and return false, it could happen that `IsPositive` and `IsNegative` are both `false` at the same time, which could be confusing to users of this API. People might expect that exactly one of those is always true.

Idea: Remove the method `IsNegative` completely. People can write `!IsPositive` instead, and you are doing that in this PR already (in `componentvector`). Removing that function would remove any ambiguity.

haydnKing: Hmm, yeah might be worth looking into where IsNegative is used. Though zero is neither negative nor positive, so really it's `!IsPositive() && !IsZero()`

PR #959, opened 2019-01-29
1:  lh.Plan() refactor: Cosmetic refactoring

haydnKing:
the function `microArch/scheduler/liquidhandling/liquidhandler.Plan()` is the core part of our instruction generation, yet it's a nest of horror and a real pain point for development and on boarding. 

I've had a crack at a refactor before, but ended up biting off more than I could chew in one go, but at least learnt what the chunks of work to do should be. This is part one, where I attempt to:
- use camelCase consistently, at least for function calls
- hang functions off the objects on which they operate as much as possible
-- sadly, this meant moving some assertions into wtype since LHInstruction currently lives there
- push some complexity down
- improve some code comments

This change should not affect overall behaviour of the system in any way, just make it easier on the eyes.


antha/anthalib/wtype/lhinstruction.go
```diff
@@ -294,3 +295,131 @@ func (ins *LHInstruction) InputVolumeMap(addition wunit.Volume) map[string]wunit
...
+// AssertNoPointerReuse make certain that inputs and outputs are not shared among LHInstructions
+// as should be ensured by calling DupLiquids()
+func (insts LHInstructions) AssertNoPointerReuse() error {
+	p := map[*Liquid]*LHInstruction{}
+	errs := make(utils.ErrorSlice, 0, len(insts))
+
+	for _, ins := range insts {
+		for _, c := range append(ins.Inputs, ins.Outputs...) {
+			if ins2, ok := p[c]; ok {
+				errs = append(errs, fmt.Errorf("POINTER REUSE: Instructions %s %s for component %s %s", ins.ID, ins2.ID, c.ID, c.CName))
```
m-i-sadowski: verbose as it may be I usually prefer to see the summary view of an instruction in this sort of situation

antha/anthalib/wtype/lhinstruction.go
```diff
@@ -294,3 +295,131 @@ func (ins *LHInstruction) InputVolumeMap(addition wunit.Volume) map[string]wunit
...
+func (insts LHInstructions) AssertVolumesNonNegative() error {
+	errs := make(utils.ErrorSlice, 0, len(insts))
+
+	for _, ins := range insts {
+		if ins.Type != LHIMIX {
+			continue
+		}
+
+		for _, cmp := range ins.Inputs {
+			if cmp.Volume().LessThan(wunit.ZeroVolume()) {
```
m-i-sadowski: Somewhat contraversially there *is* an IsNegative method on ConcreteMeasurment;
it makes sense to use that here or (better, I think) use !isPositive()

antha/anthalib/wtype/lhinstruction.go
```diff
@@ -294,3 +295,131 @@ func (ins *LHInstruction) InputVolumeMap(addition wunit.Volume) map[string]wunit
...
+				errs = append(errs, LHErrorf(LH_ERR_VOL, "negative volume for component \"%s\" in instruction:\n%s", cmp.CName, ins.Summarize(1)))
```
m-i-sadowski: this kind of error reporting for instructions looks better than just returning IDs

antha/anthalib/wtype/lhinstruction.go
```diff
@@ -294,3 +295,131 @@ func (ins *LHInstruction) InputVolumeMap(addition wunit.Volume) map[string]wunit
...
+		for _, cmp := range ins.Inputs {
+			if tV := cmp.TotalVolume(); !tV.IsZero() {
+				totalVolume = tV
+			} else if v := cmp.Volume(); !v.IsZero() {
+				volumeSum.Add(v)
+			}
+		}
+
+		if len(ins.Outputs) != 1 {
+			errs = append(errs, LHErrorf(LH_ERR_DIRE, "mix instruction has %d results specified, expecting one at instruction:\n%s",
```
m-i-sadowski: generally I prefer digits to numbers written longhand like this anyway as they're quicker to read; in this case it's particularly important since this would look like 

"mix instruction has 5 results specified, expecting one at instruction:..."

microArch/scheduler/liquidhandling/count_tips_used.go
```diff
@@ -44,7 +44,7 @@ func (lh Liquidhandler) countTipsUsed(rq *LHRequest) (*LHRequest, error) {
 	}
 
 	// output to the request
-
+	ret := make([]wtype.TipEstimate, 0, len(teHash))
 	for _, te := range teHash {
 		if te.NTips == 0 {
 			// should not be possible but I don't want to generate confusion
```
m-i-sadowski: This comment confuses me ... and I think I wrote it!
Could you please clarify what's going on here? 
Answers in the form of a comment

microArch/scheduler/liquidhandling/input_plate_setup.go
```diff
@@ -78,13 +78,13 @@ func (is InputSorter) Less(i, j int) bool {
 // INPUT: 	"input_platetype", "inputs"
 //OUTPUT: 	"input_plates"      -- these each have components in wells
 //		"input_assignments" -- map with arrays of assignment strings, i.e. {tea: [plate1:A:1, plate1:A:2...] }etc.
-func input_plate_setup(ctx context.Context, request *LHRequest) (*LHRequest, error) {
+func (rq *LHRequest) inputPlateSetup(ctx context.Context) error {
 	st := sampletracker.FromContext(ctx)
 	// I think this might need moving too
```
m-i-sadowski: I don't know what this comment means either... if we're simplifying things here I think removing this is very well in scope!

microArch/scheduler/liquidhandling/lhrequest.go
```diff
@@ -367,3 +355,169 @@ func (rq *LHRequest) updateWithNewLHInstructions(sorted []*wtype.LHInstruction)
...
+
+		if len(oo) != len(rq.OutputOrder)-len(toRemove) {
+			panic(fmt.Sprintf("Dummy instruction prune failed: before %d dummies %d after %d", len(rq.OutputOrder), len(toRemove), len(oo)))
+		}
+
+		rq.OutputOrder = oo
+
+		// prune instructionChain
+		rq.InstructionChain.PruneOut(toRemove)
+	}
```
msackman-synthace: pfft, is this complexity really worth it? I feel like it may be less code to just do it in one pass and avoid the intermediate structure of `toRemove`?

antha/anthalib/wtype/lhinstruction.go
haydnKing: good point, this should (hopefully) never actually happen so make sense to spew as much guts as possible... 
it could even be a panic, but I'll leave as is for now

microArch/scheduler/liquidhandling/count_tips_used.go
haydnKing: huh, it seems like an insurance against ever reporting zero tips required of a given tip type, but since that can't happen with the current code I'm in favour of removing it as it could one day obscure a bug

microArch/scheduler/liquidhandling/lhrequest.go
haydnKing: tbh, I think it might be possible to either
- drop and forget the dummy instructions much earlier
- make it so they don't actually affect instruction generation so they don't need to be pruned
either would be less complex, but I don't particularly want to fall down that rabbit hole in this PR

haydnKing:
this is ready for another look, just ran it through antha-tester successfully

msackman-synthace:
lgtm, but I'll defer the final stamping to @m-i-sadowski

PR #960, opened 2019-01-30
2: lh.Plan() refactor: execute the ExecutionPlanner

haydnKing:
round 2, including:
- make building the instruction tree more explicit and less indirect
- call the instruction tree an ITree (cf. IChain) instead of a RobotInstructionSet
- reduce complexity by separating
  - the building of the tree root
  - the generation of the tree nodes
  - the collection of the tree leaves


microArch/driver/liquidhandling/lhpolicytestframework_test.go
```diff
@@ -63,7 +63,7 @@ type InstructionAssertion struct {
 	Values      map[InstructionParameter]interface{}
 }
 
-func (self *InstructionAssertion) Assert(t *testing.T, ris []RobotInstruction) {
+func (self *InstructionAssertion) Assert(t *testing.T, ris []TerminalRobotInstruction) {
```
m-i-sadowski: Documentation comment please. Not clear why this has been specialised to TerminalRobotInstruction - I can imagine wanting to Assert() things about any RobotInstruction but I've not read into it. At least a comment to help make sense of it would be useful.

microArch/driver/liquidhandling/itree.go
```diff
@@ -0,0 +1,201 @@
...
+	for _, child := range tree.children {
+		if err := child.Generate(ctx, lhpr, lhpm); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Len returns the number of leaves at the bottom of the tree
```
m-i-sadowski: Bit dubious about this method name - NumLeaves or something would make more sense to me

microArch/driver/liquidhandling/itree.go
```diff
@@ -0,0 +1,201 @@
...
+				return nac, err
+			} else {
+				acc = nac
+			}
+		}
+	}
+	return acc, nil
+}
+
+func (tree *ITree) ToString(level int) string {
```
m-i-sadowski: Documentation comment please. Wondering if a String() method which gives us ToString(Itreedepth) makes sense also

microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -67,22 +66,19 @@ import (
...
-	LayoutAgent      func(context.Context, *LHRequest, *liquidhandling.LHProperties) (*LHRequest, error)
-	ExecutionPlanner func(context.Context, *LHRequest, *liquidhandling.LHProperties) (*LHRequest, error)
-	PolicyManager    *LHPolicyManager
-	plateIDMap       map[string]string // which plates are before / after versions
+	Properties      *liquidhandling.LHProperties
+	FinalProperties *liquidhandling.LHProperties
+	SetupAgent      func(context.Context, *LHRequest, *liquidhandling.LHProperties) (*LHRequest, error)
+	LayoutAgent     func(context.Context, *LHRequest, *liquidhandling.LHProperties) (*LHRequest, error)
+	PolicyManager   *LHPolicyManager
+	plateIDMap      map[string]string // which plates are before / after versions
```
m-i-sadowski: Ironically this is probably the only case where I think the Strategy pattern used here really is likely to be used in the future. However since we really aren't using it now I can see that we may as well remove it. It does leave the other parts looking a little odd but to avoid scope creep I'm not going to quibble

microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -760,11 +756,23 @@ func (this *Liquidhandler) Plan(ctx context.Context, request *LHRequest) error {
 		return err
 	}
 
-	// now make instructions
-	if rq, err := this.ExecutionPlan(ctx, request); err != nil {
+	// make the instructions for executing this request by first building the ITree root, then generating the lower level instructions
```
m-i-sadowski: Sensible comments around dangerously confusing bits of the codebase?! It'll never catch on. 

Seriously though, I approve.

microArch/driver/liquidhandling/lhpolicytestframework_test.go
haydnKing: Ah yeah, this was a least effort in the moment thing... I believe Assert is only ever called against terminal instructions, but I'll investigate and comment

microArch/scheduler/liquidhandling/liquidhandler.go
haydnKing: agreed. I think we're probably more likely to have different strategies around setup and layout in the shorter term at least so they're less bad IMO

haydnKing:
Adding some testing around ITree makes a lot of sense, will look into that. I ran the first of these branches against antha-tester, but for the subsequent ones I haven't gotten round to it, though I am planning on doing so before the final merge

microArch/driver/liquidhandling/lhpolicytestframework_test.go
haydnKing: Yes, InstructionAssertion is specific to terminal instructions - it's used to test that policies lead to the right terminal instructions being generated, e.g. that "POSTMIXCYCLES" leads to a terminal Mix with 5 cycles. I've added a comment to clarify.

msackman-synthace:
A very large part of me, probably over 3/4ters of me, wants to see ichain and itree unified to one thing.

m-i-sadowski:
> A very large part of me, probably over 3/4ters of me, wants to see ichain and itree unified to one thing.

Yes... what a great idea!

Since we're refactoring I'm wondering if we can't get rid of the silly 'microarch' designation while we're at it? 

I think these moves would be good ones to make after this and its bretheren have worked their way in

msackman-synthace:
hang on folks! We can't do everything at once, and @haydnKing has already done an excellent job making 1 big pr into 4 small ones.

I think addressing microArch and naming, perhaps belongs in the same bucket as sorting out anthalib, anthastandardlibrary, and moving a bunch of this stuff out to elements world. So I'm hesitant for this to expand massively.

haydnKing:
fear not, I'm definitely not intending bringing that into the current work :)

m-i-sadowski:
It's OK, I wasn't suggesting we do it in this PR. Just discussing the idea - I brought up moving things around precisely because, as Haydn points out, we'd need to do that anyway to avoid an import cycle.

PR #963, opened 2019-01-31
4: lh.Plan() refactor: shrinkVolumes

haydnKing:
When autoallocating liquids, we can't be accurate about exactly how much we need, so we over-allocate. We only know how much we need once we've generated all the instructions, at which point we reduce the amount allocated accordingly.

Previously the function `Liquidhandler.reviseVolumes` assumed
a) Liquidhandler.Properties contains the final properties and
b) Liquidhandler.FinalProperties contains the initial properties
and then:
1) shrinks the auto allocated volumes by
   - scanning through the instructions to figure out exactly how much is really taken from each autogenerated well
   - updates the initial and final volumes accordingly
   - removes unused wells and unused plates
2) updates the IDs of plates in the final properties (tipboxes and tipwastes have their own function)
3) swaps the autogenerated components from FinalProperties to InitialProperties to account for (a) and (b)
4) separately iterates through and swaps non-autogenerated components because of (a) and (b), again, subtly updating component IDs

🤢 

this change:
- makes it so that `Liquidhandler.Properties` and `Liquidhandler.FinalProperties` only ever contain what you'd expect, meaning 3&4 aren't needed
- simplifies step (1) to keep track of which wells and plates get used, so no more IsTemporary flag
- updates plate, tipbox, tipwaste, and Liquid/Component IDs separately, in their own function

Testing:
This branch has been tested successfully using `antha-tester`, and it includes all the changes from PRs 1,2,3 in this series. 

None of these changes should be visible from user-land but hopefully make this whole thing a great deal easier to understand and iterate on.


m-i-sadowski:
"This branch has been tested successfully using antha-tester, and it includes all the changes from PRs 1,2,3 in this series."

I think I'll wait until those PRs get finished and merged before I read through all this... there's a lot of additional stuff here from what I can see.

microArch/driver/liquidhandling/lhproperties.go
```diff
@@ -502,6 +506,45 @@ func (lhp *LHProperties) RemovePlateWithID(id string) {
 	delete(lhp.Plates, addr)
 }
 
+// UpdateID update the id of the object with currID, returning the new ID
```
m-i-sadowski: Can you add some unit tests for this please?

microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -297,251 +296,187 @@ func (this *Liquidhandler) Execute(request *LHRequest) error {
...
-	// XXX -- HARD CODE 8 here
-	lastPlate := make([]string, 8)
-	lastWell := make([]string, 8)
-
-	vols := make(map[string]map[string]wunit.Volume)
-
-	rawvols := make(map[string]map[string]wunit.Volume)
+// shrinkVolumes reduce autoallocated volumes to the amount we actually need, removing
+// any unused wells or plates
+func (this *Liquidhandler) shrinkVolumes(rq *LHRequest) error {
```
m-i-sadowski: big and horrid though it is, this is another potentially error-prone area which really deserves some testing around the various cases it covers

haydnKing:
Ok, added some tests for `UpdateIDs` and (more extensively) `shrinkVolumes`, hopefully this is ready once https://github.com/antha-lang/antha/pull/961 goes in.

Found that `shrinkVolumes` would remove plates which weren't used whether they were autoallocated or not... I'm not sure if this would actually be an issue (maybe if there's a plate sat there incubating, waiting for a downstream step?), but it seemed like unexpected behaviour so I fixed it - `shrinkVolumes` will now only remove plates if they only contain autoallocated liquids

microArch/scheduler/liquidhandling/liquidhandling_test.go
```diff
@@ -1282,3 +1284,515 @@ func getTestMix(components []*wtype.Liquid, address string) *wtype.LHInstruction
...
+		if err := initial.AddPlateTo(addr, plate); err != nil {
+			t.Fatal(err)
+		}
+		inputPlateOrder = append(inputPlateOrder, plate.ID)
+	}
+
+	// set autoallocated flags
+	for addr, wells := range test.AutoAllocatedWells {
+		if plate, ok := initial.Plates[addr]; !ok {
+			t.Fatalf("can't set AutoGenerated: no plate at address \"%s\"", addr)
```
m-i-sadowski: AutoGenerated --> AutoAllocated

microArch/scheduler/liquidhandling/liquidhandling_test.go
```diff
@@ -1282,3 +1284,515 @@ func getTestMix(components []*wtype.Liquid, address string) *wtype.LHInstruction
...
+		} else {
+			for _, wellAddr := range wells {
+				if well, ok := plate.Wellcoords[wellAddr]; !ok {
+					t.Fatalf("can't set AutoGenerated: plate type %s at %s has no well %s", plate.GetType(), addr, wellAddr)
```
m-i-sadowski: AutoGenerated --> AutoAllocated

microArch/scheduler/liquidhandling/liquidhandler.go
m-i-sadowski: You fought the good fight!

PR #964, opened 2019-01-31
Support introspection of mixers

msackman-synthace:
Turns out we already have a way for doing device introspection, so extend that a little, and require that mixers also implement it.
All in all, a simplification of try.go.
Has been tested.


msackman-synthace:
So I've done https://github.com/Synthace/PipetMaxDriver/pull/60 but you'll need to tell me what other mixer driver plugin thingies I need to change.

target/auto/try.go
```diff
@@ -63,27 +68,16 @@ func (a *tryer) AddDriver(ctx context.Context, conn *grpc.ClientConn, arg interf
...
-		return nil
-	}
-
-	err = a.addLowLevelMixer(ctx, conn, arg)
-
-	if err == nil {
-		return nil
+func (a *tryer) AddMixer(ctx context.Context, conn *grpc.ClientConn, arg interface{}, subtypes []string) error {
+	switch {
+	case len(subtypes) > 0 && subtypes[0] == "GilsonPipetmax":
```
haydnKing: If you wanted to do it this way, you could have just done `drv.GetCapabilities().Mnfg == "Gilson"`
Seems a bit wrong for antha-lang to have to know the mapping from device name to API level though, so maybe `subtypes[0] == "some.fancy.string.prefix.v1.lowlevel"`?
I'd really like to be able to grep antha-lang for "gilson" and have nothing show up...

msackman-synthace: See, personally, I think the abstraction gains us exactly nothing, and I would expect to see the exact manufacturers in the code base. So I'd like to understand this a bit better.
Can't use GetCapabilities at this point because we don't have an interface for mixer without knowing whether it's high or low. At this point all we know is that it's a driver.

haydnKing: hmm, yeah I guess it doesn't really save that much keeping a really clean separation... I'm generally keen to keep the amount of random boiler plate that has to be thought about when adding more drivers to a minimum, but I doubt that'll ever happen at a fast enough rate to be a problem

msackman-synthace: I agree, but I also think the easy discovery from the antha code base of "oh ok, we clearly treat gilson as a low level driver" is valuable. Ult we'll end up with ~10 strings here max?

msackman-synthace: See what I've done to this code now. Super cool - rare that you get to rip methods off a type rather than an object instance.

msackman-synthace:
Done all the drivers. Ready!

PR #965, opened 2019-02-05
Policy Manager isn't used here.

msackman-synthace:
It's in LHRequest instead.


msackman-synthace:
sodding 3 mins to write the diff. 20 mins to get travis ci to work... grrr.

PR #970, opened 2019-02-11
Stop handling humans.

msackman-synthace:


msackman-synthace:
antha-tester seems to pass.

PR #971, opened 2019-02-13
Fix/antha 2357

m-i-sadowski:
This PR (finally!) contains fixes to two  separate bugs which were reported to violate user specified order of transfers, Antha-2357 and Antha-1809 (potentially also Antha-1810)

I've reproduced the final commit message here as a summary of what's happened and how it's been tested.

Antha 2357 / 1809 : mix ordering 

Summary:

Two bugs relating to mix ordering were found and fixed.

In one case the old structure of generating multichannel instructions
first meant that in some cases it would do some single channel
operations out of order because they would always be generated after
multichannel instructions even if they are for components which should
come later.

In the other case the behaviour of GetComponents was complex and
confusing and led to rearrangment of transfers from user order into
volume order as a side-effect of the alignment scoring used to select
sources. This works fine for multichanneling as it's suitably greedy but
works badly when a vector of single channel components is specified.

I've cleaned up the code in both places and removed the SingleChannel
instruction types so that MultiChanneling with 1 transfer is the new
SingleChanneling.

Test Plan:

- I've added regression tests for both cases which failed in the
  original case and pass post fix
- I've tested successfully with antha-tester both normally and with
  simulator volume checking merged in.
- I've verified by eye that both bundles specified in the original
  tickets now behave as intended.


microArch/driver/liquidhandling/convertinstructions.go
```diff
@@ -66,7 +66,7 @@ func hasMultiChannelBlock(ctx context.Context, tfrs []*TransferInstruction, rbt
 		}
 
 		for _, ins := range instrx {
-			if ins.Type() == MCB {
+			if ins.Type() == MCB && ins.GetParameter(MULTIGT1).(bool) {
```
haydnKing: nah, MULTIGT1 isn't a parameter! How about:
```go
if mcbi, ok := ins.(*MultiChannelBlockInstruction); ok {
	for _, m := range mcbi.Multi {
		if m > 1 {
			return true, nil
		}
	}
}
```
much clearer what's going on, and MULTIGT1 isn't necessary

microArch/driver/liquidhandling/transferblock_test.go
```diff
@@ -12,6 +11,7 @@ import (
 	"github.com/antha-lang/antha/inventory"
 	"github.com/antha-lang/antha/inventory/cache/plateCache"
 	"github.com/antha-lang/antha/inventory/testinventory"
+	"reflect"
```
haydnKing: moved?

microArch/driver/liquidhandling/transferblock_test.go
```diff
@@ -554,15 +560,15 @@ func testNegative(ctx context.Context, ris []RobotInstruction, pol *wtype.LHPoli
 		}
 
 		for _, ri := range ri2 {
-			if ri.Type() != SCB {
-				t.Errorf("Multichannel block generated without permission: %v %v %v", ri.GetParameter(LIQUIDCLASS), ri.GetParameter(WELLFROM), ri.GetParameter(WELLTO))
+			if ri.Type() == MCB && ri.GetParameter("MULTIGT1").(bool) {
+				t.Errorf("Multichannel block generated without permission: %v %v %v %v", ri.GetParameter(MULTI), ri.GetParameter(LIQUIDCLASS), ri.GetParameter(WELLFROM), ri.GetParameter(WELLTO))
```
haydnKing: as above - plus this error doesn't really make much sense now since a MCB is the only choice

microArch/driver/liquidhandling/transferblock_test.go
```diff
@@ -578,24 +584,19 @@ func generateRobotInstructions(t *testing.T, ctx context.Context, inss []*wtype.
...
 
 	//generate the low level instructions
-	tree := NewITree(tb)
-	if _, err := tree.Build(ctx, pol, rbt); err != nil {
-		t.Fatal(err)
-	} else if ris, err := tree.Leaves(); err != nil {
+
+	iTree := NewITree(tb)
+	iTree.Build(ctx, pol, rbt)
+	ris2, err := iTree.Leaves()
```
haydnKing: 😜 ok, ok

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -25,14 +25,13 @@ package liquidhandling
...
-	"strconv"
-	"strings"
-
 	"github.com/antha-lang/antha/antha/anthalib/wtype"
 	"github.com/antha-lang/antha/antha/anthalib/wunit"
 	"github.com/antha-lang/antha/antha/anthalib/wutil"
 	"github.com/antha-lang/antha/inventory/cache"
+	"sort"
+	"strconv"
+	"strings"
```
haydnKing: moved?

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -440,110 +448,97 @@ func (ins *TransferInstruction) Generate(ctx context.Context, policy *wtype.LHPo
...
 
-			// ISSUE: Duplicating here doesn't account for the possibility of one or more of these becoming zero
-			// 	  need to ensure there are no non-contiguous zeroes
+		return true
+	}
 
-			if vols.NonContiguous() && !mci.Prms.Independent {
-				//must be done single channel
-				break
+	resetBuffers := func(mB, sB []MultiTransferParams, ins *MultiChannelBlockInstruction) ([]MultiTransferParams, []MultiTransferParams) {
```
haydnKing: would `flushBuffers` be clearer?

microArch/driver/liquidhandling/transferinstruction.go
```diff
@@ -440,110 +448,97 @@ func (ins *TransferInstruction) Generate(ctx context.Context, policy *wtype.LHPo
...
+	multis, channels := ins.GetParallelSetsFor(ctx, prms, pol)
 
-			// set the from and to volumes for the relevant part of the instruction
-			ins.Transfers[set].RemoveFVolumes(vols)
-			ins.Transfers[set].AddTVolumes(vols)
+	for set, multi := range multis {
+		newSet := ins.Transfers[set].Component()
 
-			mci.Multi = len(vols)
+		if !setsSame(newSet, oldSet) {
```
haydnKing: comment? e.g. `// the next components are a different type, so flush all the buffers`

microArch/driver/liquidhandling/transferinstruction.go
haydnKing: (also, this is gonna be great on fancy liquidhandlers where we can drop some tips but not all of them)

microArch/driver/liquidhandling/transferinstruction_test.go
```diff
@@ -0,0 +1,277 @@
...
+			for _, cmpa := range cmps {
+				ret = append(ret, cmpa...)
+			}
+		}
+	}
+
+	return ret
+}
+
+// Conceived as a test of Antha-2357
```
haydnKing: can you add the description? seems likely ANTHA-2357 will be meaningless in a week or two!

microArch/driver/liquidhandling/transferinstruction_test.go
```diff
@@ -0,0 +1,277 @@
...
+	order := getComponentOrder(inss)
+
+	if !reflect.DeepEqual(order, expectedOrder) {
+		t.Errorf(fmt.Sprintf("Expected order %v got %v", expectedOrder, order))
+	}
+}
+
+// the idea here is that water should be moved before wine but
+// wine multichannels and water doesn't.
+func getTransferMulti() RobotInstruction {
```
haydnKing: since `getTransferMultiN()` is only used by `TestOrderingN(...)`, perhaps they should just be one function?

microArch/driver/liquidhandling/transferparams.go
```diff
@@ -8,24 +8,24 @@ import (
...
+	FVolume    wunit.Volume              // volume in 'from' well
+	TVolume    wunit.Volume              // volume in 'to' well
+	Channel    *wtype.LHChannelParameter // which channel to use
+	TipType    string                    // type of tip to use
+	FPlateWX   int                       // X dimension in 'from' plate
+	FPlateWY   int                       // Y dimension in 'from' plate
+	TPlateWX   int                       // X dimension in 'to' plate
+	TPlateWY   int                       // Y dimension in 'to' plate
+	Component  string                    // component type
+	Policy     wtype.LHPolicy            // policy attached to this transfer
```
haydnKing: ❤️ the comments!

microArch/scheduler/liquidhandling/liquidhandling_test.go
```diff
@@ -800,6 +800,7 @@ func TestExecutionPlanning(t *testing.T) {
 			InputPlates:  []*wtype.LHPlate{GetTroughForTest()},
 			OutputPlates: []*wtype.LHPlate{GetPlateForTest()},
 			Assertions: Assertions{
+				DebugPrintInstructions(),                   // nolint
```
haydnKing: ? this looks like my doing... but how? it can go since it just prints the instructions

microArch/driver/liquidhandling/convertinstructions.go
msackman-synthace: or bounce through the visitor impl on those types to avoid the dynamic type check. Might be a bit longer though...

m-i-sadowski: I've taken the middle way and made MultiChannelBlock have an instance method for this as I think it's clearest.

microArch/driver/liquidhandling/transferblock_test.go
m-i-sadowski: Blame git

microArch/driver/liquidhandling/transferblock_test.go
m-i-sadowski: I've amended the error and one other to be a bit more descriptive

microArch/scheduler/liquidhandling/liquidhandling_test.go
m-i-sadowski: I'll move the nolint and remove the noise

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -443,9 +232,23 @@ func (ins *MultiChannelBlockInstruction) Generate(ctx context.Context, policy *w
...
+		findPrmSet := func(sa []*wtype.LHChannelParameter) *wtype.LHChannelParameter {
+			for _, s := range sa {
+				if s != nil {
+					return s
+				}
+			}
+			return nil
+		}
+
+		prmSet := findPrmSet(ins.Prms[t])
```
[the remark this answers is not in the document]
m-i-sadowski: That is true; OTOH it's a *bit* neater than writing it all out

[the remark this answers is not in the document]
m-i-sadowski: Actually I think it's not needed at all; good spot. Suspect it's part of a *cul de sac* which just didn't get decomissioned

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -301,8 +61,9 @@ type MultiChannelBlockInstruction struct {
 	TPlateType [][]string
 	FVolume    [][]wunit.Volume
 	TVolume    [][]wunit.Volume
-	Multi      int
-	Prms       *wtype.LHChannelParameter
+	Component  [][]string
+	Prms       [][]*wtype.LHChannelParameter
```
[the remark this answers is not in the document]
m-i-sadowski: Yeah the purpose is to allow channel parameters to vary per sub-transfer here, and in theory each channel might itself have different properties - 'nil' means a channel isn't used so there might be gaps internally. *Leading* gaps, OTOH, don't make any sense since we haven't made an assignment of transfers to channels here, so I think we should be able to assert that the first channel is always used.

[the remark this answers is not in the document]
m-i-sadowski: This does need revisiting: we need to separately model *head* parameters and *channel* parameters for this to make any sense: channels can't legitimately declare a *multi* value (unless we support watering cans at some point) but obviously heads need to. So the idea of one multi per transfer is silly, and should not be tolerated.

PR #972, opened 2019-02-21
Handle unrecognised characters in complement

mcouch-synthace:
Currently `wtype.Comp` silently ignores unrecognised characters.

This leads to output that is shorter than input if there are unrecognised characters in the input.

Separately, it would be handy if some non-DNA characters for example `-` (gap) and `.` (match) were passed through.  Then gapped alignments could be reverse-complemented: a useful extension.

The following changes could be made:
1. raise error if a character is not a recognised nucleotide
2. allow specific extra non-DNA characters - an extended alphabet ([this list of IUPAC codes](http://www.bioinformatics.org/sms2/iupac.html]) allows both `-`and `.`)
3. pass through any unrecognised characters unmodified 

Option 1 may have a big impact. There is code to validate nucleotides in `wtype` but validation is not enforced and may not be used widely.  PR https://github.com/antha-lang/antha/pull/976 suggests no tests would break on CI currently.

Option 2 would have less impact but silent omissions would still be possible. This option seems to entail further development around alphabet handling and (again) use of validation.

Option 3 (implemented here) seems to be the simplest fix.

Or Option 1 with panic per #976.

### DIFF
--- a/antha/anthalib/wtype/biology.go
+++ b/antha/anthalib/wtype/biology.go
@@ -888,7 +888,11 @@ func Comp(s string) string {
 	}
 
 	for _, c := range s {
-		r += m[string(c)]
+		if rc, ok := m[string(c)]; ok {
+			r += rc
+		} else {
+			r += string(c)
+		}
 	}
 
 	return r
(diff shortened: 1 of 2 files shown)


mcouch-synthace:
@chr1sgrant this is the change in anthalib we discussed

PR #973, opened 2019-02-21
ANTHA-2518: Empty return if units is missing.

cosmonautspiff:
Instead of triggering a nil pointer error, the methods will return an empty variable.

Please check the differences and feel free to change how this was done.

Beware, that there will be **no** error anymore if someone forgets to enter units. Is that preferable? Perhaps we should just implement a UI check.

### DIFF
--- a/antha/anthalib/wunit/unit.go
+++ b/antha/anthalib/wunit/unit.go
@@ -109,21 +109,34 @@ func (self *Unit) GobDecode(b []byte) error {
 
 // Name get the full name of the unit
 func (self *Unit) Name() string {
+	if self == nil {
+		return ""
+	}
 	return self.prefix.Name + self.name
 }
 
 // getBaseSIConversionFactor factor to multiply by in order to convert to SI value including effect of prefix
 func (self *Unit) getBaseSIConversionFactor() float64 {
+  if self == nil {
+    return 0.0
+  }
 	return math.Pow(self.prefix.Value, float64(self.exponent)) * self.multiplier
 }
 
 // String a string representation of the unit name and symbol
 func (self *Unit) String() string {
+	if self == nil {
+		return ""
+	}
 	return fmt.Sprintf("%s[%s]", self.Name(), self.PrefixedSymbol())
 }
 
 // Prefix get the SI prefix of this unit, or " " if none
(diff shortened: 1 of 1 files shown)


x-anderson:
Looks good and really useful for catching these errors here, instead of elements! 👍  The Travis failing - is this due the the Errorf call without an argument?

antha/anthalib/wunit/unit.go
```diff
@@ -140,16 +153,25 @@ func (self *Unit) PrefixedSymbol() string {
...
+	if self == nil {
+		return ""
+	}
 	return self.siSymbol
 }
 
 // getConversionFactor get the conversion factor between this unit and rhs.
 func (self *Unit) getConversionFactor(rhs *Unit) (float64, error) {
+	if self == nil || rhs == nil {
+		return 0.0, errors.Errorf("cannot convert units: nil units provided")
```
x-anderson: Errorf should require an argument passed in for formatting.

If both self and rhs are nil - should a nil error be returned anyway?

antha/anthalib/wunit/unit.go
```diff
@@ -109,21 +109,34 @@ func (self *Unit) GobDecode(b []byte) error {
...
+		return ""
+	}
 	return fmt.Sprintf("%s[%s]", self.Name(), self.PrefixedSymbol())
 }
 
 // Prefix get the SI prefix of this unit, or " " if none
 func (self *Unit) Prefix() SIPrefix {
+	if self == nil {
+    var emptyPrefix SIPrefix
+		return emptyPrefix
```
x-anderson: Could save a line here and write:

```
return SIPrefix{}
```

and remove

```
var emptyPrefix SIPrefix
```

antha/anthalib/wunit/unit.go
cosmonautspiff: Used errors.New

Sorry, not sure I understand the nil error comment

antha/anthalib/wunit/unit.go
cosmonautspiff: done

cosmonautspiff:
My bad, I thought that building antha would format everything.

cosmonautspiff:
If this is fine, please merge.

PR #975, opened 2019-02-25
Align formatting

mcouch-synthace:
Provides support for formatted alignment output similar to NCBI.

Will make it possible to rewrite the loop in element `ValidateSequenceResults`
https://repos.antha.com/antha-ninja/elements-westeros/blob/master/Elements/Molecular_Biology_Starter/ValidateSequenceResults/element.an#L204
using data obtained by a new Blocks method on `Alignment`.

This will avoid exposing fields such as TemplateResult and QueryPositions and decouple further development of the align package from elements which depend on it.

`ValidateSequenceResults` appears to be the only released element referring to the `align` package.  There are at least two new elements being developed which will use it.  The Blocks method will be useful for these and bring consistency.


mcouch-synthace:
Rethink

PR #979, opened 2019-02-25
Implement MixSummary for new mix graph

haydnKing:
This code implements the code to create the appropriate JSON for the up-coming mix graph API

The idea is to inspect the deck layout and ITree and generate an easily digestible record of what happened during the lh task which can then be served to the front end on request - this will eventually replace the horrible horrible deck matrix API in microserviceland (which did all sorts of funky inspection of the report) and the horrible horrible "simulation" that happens in the front end to build the current mix-graph.

To keep things flexible, I have tried to minimise the assumptions made about the structure of the instruction tree, so that if in the future we (for example) add an extra layer of instructions somewhere we shouldn't break this code (at least too badly).

There are some initial limitations, which I've endeavoured to document. Most relate to known limitations of the physical simulator - that it doesn't use Mix to set sub-components properly and doesn't account for carry volumes. The other is that HighLevelLiquidhandlers are not supported. This should be fairly trivial to support, but I haven't done so as the only HL driver is currently not functioning.


haydnKing:
I had a couple of ideas about this
1) the deck layout part should include empty deck locations - I'll refactor it accordingly
2) the JSON for the deck layout and the actions should be stored separately. 
That way we can have two endpoints, one returning the layout (initial and final) and one returning the actions. That way the front end can get the layout and all the deck items quickly without having to download all the actions.

haydnKing:
Ok, think this is probably done for now. All that needs doing is for antha-runner to call the two summary methods on each mix task and store them away somewhere for the relevant microservice to serve later

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+			Description:    fmt.Sprintf("Plate with %dx%d wells", o.NCols(), o.NRows()),
+			Dimensions:     newCoordinates(o.GetSize()),
+			WellDimensions: newCoordinates(o.Welltype.GetSize()),
+			WellStart:      coordinates{X: o.WellXStart, Y: o.WellYStart, Z: o.WellZStart},
+			WellOffset:     coordinates{X: o.WellXOffset, Y: o.WellYOffset},
+			WellType:       wellTypes[o.Welltype.Shape().ShapeName],
+			Rows:           o.NRows(),
+			Columns:        o.NCols(),
+			Contents:       contents,
+		}
```
m-i-sadowski: Residual volume value here if we agree that it makes sense (which of course we will)

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+				if tip == nil {
+					missingTips = append(missingTips, &wellCoords{Row: rowNum, Column: colNum})
+				}
+			}
+		}
+
+		return &itemSummary{
+			ID:             o.ID,
+			Name:           o.Boxname,
+			Kind:           "tipbox",
```
m-i-sadowski: worth distinguishing boxes fromracks already? It's consistent within any currently (/ very near future-ly) supported instrument but different between instruments. Seems worth being accurate

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+				return nil, err
+			}
+		}
+	}
+
+	return &transferAction{Children: children}, nil
+}
+
+func (ta *transferAction) MarshalJSON() ([]byte, error) {
+	// alias the type so as not to invoke this function in a loop
```
m-i-sadowski: smart!

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+	From      *contentUpdate      `json:"from"`   // the source from which liquid is taken, and the new contents
+	To        []*contentUpdate    `json:"to"`     // the destination(s) in which liquid is placed, and their new contents. Multi-dispenses are represented as a slice of destinations
+	Volume    measurementSummary  `json:"volume"` // the volume of liquid transfered
+	Wasted    measurementSummary  `json:"wasted"` // the volume lost as "carry volume" during the transfer
+	Policy    string              `json:"policy"` // the liquid type for this transfer
+	PreMix    *mixSummary         `json:"pre_mix,omitempty"`
+	PostMix   *mixSummary         `json:"post_mix,omitempty"`
+	Blowout   *measurementSummary `json:"blowout,omitempty"`
+	TouchOff  bool                `json:"touchoff"`
+	TipSource tipLocation         `json:"tip_source"` // the source of tips used to carry out this transfer
```
m-i-sadowski: Hmmm - I really think we should represent tip changes explicitly. Otherwise the front end potentially needs to infer them.

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+			WellType:       roundWell,
+			Rows:           o.NRows(),
+			Columns:        o.NCols(),
+			MissingTips:    missingTips,
+		}
+	case *wtype.LHTipwaste:
+		return &itemSummary{
+			ID:             o.ID,
+			Name:           o.Name,
+			Kind:           "tipwaste",
```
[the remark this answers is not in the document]
haydnKing: atm these are just used to distinguish between "plate", "tipbox", "tipwaste"... I think I'd be in favour of adding a "type" field to store the actual database type

target/summary.go
```diff
@@ -0,0 +1,569 @@
+package target
```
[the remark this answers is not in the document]
haydnKing: All very good points. 

The idea of having a better contract here had occurred to me, it's something I'll have to work out with Martin when he's back (Tuesday I think).

I don't have a good answer as to where this code should live. Initially I wrote it in `microArch/scheduler/liquidhandling` but (a) it had to import pretty much everything from `microArch/driver/liquidhandling` and (b) `target/Mix` is the only place where all the information you need to build this summary actually comes together. Perhaps we should chat this out today?

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+type wellType string
+
+const (
+	roundWell  = "cylinder"
+	squareWell = "cuboid"
+)
+
+// wellTypes map from internal well types to the shorter list understood by the UI since
+// none of these are actually different in their meaning
+var wellTypes = map[wtype.ShapeTypeID]wellType{
```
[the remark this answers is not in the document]
haydnKing: Perhaps the thing to do is to bite the bullet and reduce the number of "ShapeTypeID"s that we have to the sane number: 2, and just use that directly?

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+	}
+}
+
+// tipLocation the source of a particular tip
+type tipLocation struct {
+	wellLocation
+	Reuses int `json:"reuses"` // record how many times the tip has been reused
+}
+
+type mixSummary struct {
```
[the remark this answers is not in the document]
haydnKing: 👍 that should go in with the transfer summary as well

target/summary.go
```diff
@@ -0,0 +1,569 @@
...
+	wtype.RectangleShape: squareWell,
+}
+
+type wellCoords struct {
+	Row    int `json:"row"`
+	Column int `json:"col"`
+}
+
+// itemSummary summarize an item on the deck and its initial contents
+type itemSummary struct {
```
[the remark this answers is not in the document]
haydnKing: 👍

target/summary.go
haydnKing: 😀

target/summary.go
haydnKing: Hmm, I think I'd prefer to keep them all the same for now since we don't actually represent the difference in the backend... if we later decide there's a good reason to differentiate then we can add the new type later

target/summary.go
haydnKing: :-/ I'm thinking you may be right. I'll iterate a bit on this.

target/summary.go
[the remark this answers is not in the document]
pcwieczkowski: +1 for having a schema defined - if it's JSON schema (or similar), we could use that to generate typescript definitions in the frontend (and maybe also golang structs?)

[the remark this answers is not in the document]
haydnKing: Yeah, just looking into go support for json-schema now: looks like there's a decent validator that we can use for unit-testing, the question I'm yet to answer is where the authoritative schema definition should actually live since both the front end and antha-lang will need access to it

target/layout.schema.json
```diff
@@ -0,0 +1,307 @@
+{
```
pcwieczkowski: Nice! this is what I could quickly generate with https://www.npmjs.com/package/json-schema-to-typescript - https://phabricator.synthace.us/P28

The generated types are not perfect but are very close to what @mkonicek has in his code. I'll check with him next week if this will be useful.

haydnKing:
Updates since last time:
- summary code moved back to `microArch/scheduler/liquidhandling/summary.go` with appropriate helpers hung off `target.Mix`
- tip loading and unloading is explicitly represented
- much more detail about the nature of lh operations (asp, dsp, mix), including
  - flow rates
  - dispense heights
  - LiquidLevelFollow behaviour
  - TouchOff
- more detail about objects on deck, including type
- structure of all generated JSON is validated using json-schema, and the same schema was used to generate typescript bindings for front end

Testing:
- Existing planning tests using the `microArch/scheduler/liquidhandling.PlannerTest` API now always check that they generate valid JSON
- These tests can now also use `AssertLayoutSummary` and `AssertActionsSummary` to check that the content of the JSON generated in these cases matches a given input file
  - this was somewhat complicated by two factors
    - Unmarshalling is hard since not everything is a concrete type
    - the ids of objects on deck change from simulation to simulation, so needed to be normalised before comparison

PR #980, opened 2019-02-25
Alignment to sequence collection

mcouch-synthace:
New method in `align` package to align a query to a collection (database) of targets.

Benchmarked against E. coli CDNAs (5494 sequences) - takes only a few seconds.

So it could be feasible to use this in conjunction with sequences files from a data repo, avoiding the BLAST API for tasks such as: primer specificity checking.


antha/AnthaStandardLibrary/Packages/sequences/align/align_db_test.go
```diff
@@ -0,0 +1,143 @@
...
+	if results[0].Score() != 15 {
+		t.Errorf("Error, unexpected top score: got %d, want %d\n", results[0].Score(), 15)
+	}
+
+}
+
+func TestDNADbBenchmark(t *testing.T) {
+
+	t.Skip("requires data file")
+
```
chr1sgrant: How large is this file? 
I might be tempted to include it

antha/AnthaStandardLibrary/Packages/sequences/align/align_db.go
```diff
@@ -0,0 +1,52 @@
...
+// DNADb aligns a query to a collection (or database) of sequences, testing both
+// forward and reverse directions. It returns the top scoring alignment results
+// found in rank order, up to a specified number.
+func DNADb(query wtype.DNASequence, templates []wtype.DNASequence, alignmentMatrix ScoringMatrix, maxResults int) ([]Result, error) {
+
+	results := make([]Result, 0)
+
+	for _, template := range templates {
+
+		//fmt.Printf("Checking %d, %s\n", i, template.Name())
```
chr1sgrant: please remove commented out print statement

antha/AnthaStandardLibrary/Packages/sequences/align/align_db.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"sort"
+)
+
+// DNADb aligns a query to a collection (or database) of sequences, testing both
+// forward and reverse directions. It returns the top scoring alignment results
+// found in rank order, up to a specified number.
+func DNADb(query wtype.DNASequence, templates []wtype.DNASequence, alignmentMatrix ScoringMatrix, maxResults int) ([]Result, error) {
```
chr1sgrant: The term Db (database) seems slightly misleading here (is this the typical term used to bioinformatics practitioners?) 

maybe DNASet or DNACollection would be clearer to understand the meaning when reading the code?

align.DNASet seems clearer than aligh.DNADb (to me anyway) and we seem to be converging around the term Set in other parts of the code base (Liquid Set) so would possibly be more consistent.

mcouch-synthace: OK, I like DNASet and will go with that. 
Yes bioinformaticians do tend refer to any collection as a database but it is overused.

antha/AnthaStandardLibrary/Packages/sequences/align/align_db.go
mcouch-synthace: OK

antha/AnthaStandardLibrary/Packages/sequences/align/align_db_test.go
mcouch-synthace: 100K lines, too big I think.

PR #981, opened 2019-02-26
Alignment methods

mcouch-synthace:
Provides new methods for accessing `align.Alignment` and some formatting methods.

Should help with coordination of work on the `align` package and the elements that depend on it.

It seems that `ValidateSequenceResults` is currently the only released element referring to the `align` package.  But there are at least two new elements being developed which will also use it.


antha/AnthaStandardLibrary/Packages/sequences/align/align_methods.go
```diff
@@ -0,0 +1,113 @@
+package align
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Start position in query
+func (a *Alignment) QStart() int {
```
chr1sgrant: Strictly speaking we should be following the guidelines for documenting go code which suggests starting the comment with the name of the exported method, type, func, constant or varaible:

e.g. 
```go
// QStart returns the start position of the query in a human friendly form (i.e. first position is returned as 1)
 func (a *Alignment) QStart() int {
}
```

chr1sgrant: https://blog.golang.org/godoc-documenting-go-code

chr1sgrant: The other question is whether to name QStart, TStart etc or the full name QueryStart, TemplateStart;

The latter is more self documenting when reading the code so I personally prefer that but not strongly.

mcouch-synthace: Yes will fix the non-standard comment.

Re QStart etc. I am a bit conflicted. I find the name Template a bit weird since it seems use case specific. But this seems a pretty minor point. I was thinking Q and T save on typing and also have the feature that, being both 1 character, they line up in all those places in the code that both Q and T have to be dealt with.

Let me know what you think, happy to go full text if you think better.

mcouch-synthace:
@chr1sgrant I have made improvements suggested

also including 58d5538 following merge #972

PR #982, opened 2019-02-26
Assembly optimisation

mcouch-synthace:
Trying to understand [`concreteOptimiser.go`](https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go)

Added some notes and tests.

Not passing currently due to either bug or test based on misunderstanding.

To look at:
1. Possible off-by-one error, assuming the split point `p` is zero-offset https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L354.  Split points look like they are zero-offset [here](https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L747) and [here](https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L104) for example.
2. Need to compare with [`makeFragmentsFromSolution`](https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L163) which generates the actual overhangs to return, based on the optimised solution.
3. Also possible off-by-one: this function used to calculate distance between split points  https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L448
4. In `Recombine` - no sort before validation, but `valid` assumes splits are ordered https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L521
5. Similarly here - https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L655
6. There is a normalisation of population fitness values in an innermost loop, possible speed up? https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L693
7. Possible of-by-one error calculating segment length in cost (small effect) https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L782
8. It seems that function `endsOK` checks for duplicates in the 5' overhang set, but not for whether any overhangs could be reverse complements of each other.

### DIFF
--- a/antha/AnthaStandardLibrary/Packages/Optimization/NOTES.md
+++ b/antha/AnthaStandardLibrary/Packages/Optimization/NOTES.md
@@ -1,0 +1,202 @@
+#### General
+
+The function `OptimizeAssembly` takes an amino acid translation of a
+query sequence (`string`), a multiple sequence alignment (MSA) of
+variant DNA sequences (`[]string{}`) and a constraints struct. The
+constraints struct contains the DNA query sequence itself (`string`).
+
+The function runs an iterative optimization algorithm (genetic
+algorithm) seeking a split of the DNA query (into segments) which:
+- satisfies the constraints
+- achieves a low value of a certain cost function
+
+`OptimizeAssembly` returns the best split found, with attached overhangs of
+the specified length. It would be easy to modify it to return a selection
+of close-to-best, alternative splits.
+
+### Mapping to AssemblyProblem
+The optimization works with an `AssemblyProblem` and the constraints.
+
+An `AssemblyProblem` consists of an amino acid sequence (`string`) and a
+slice of 'mutations' (`[][2]int`).  Each mutation (`[2]int`) consists of
+- a position (which must be a codon centre, zero-based) in the underlying DNA
+sequence at which mutations will be made
+- a number, which is the number of variant codons (distinct) at that position
+
+The mutations data is calculated from the input MSA. If there are no
+variants at a particular position, nothing is recorded in the
+mutations data. Otherwise, the total number of distinct codons at the
+position is recorded.
+
+The code expects (but does not require) that the wild type is
+present in the MSA, since this ensures that the wild type codons are
+included in the variant counts.
+
+```
+Example:
+
+MSA:
+
+agcaagggcgaggagctgttcaccggggtggtgcccatcctggtcgagctggacggcgac - wild type
+ S  K  G  E  E  L  F  T  G  V  V  P  I  L  V  E  L  D  G  D
+.......t.............c...................................... - variant 1; 2:ggc->gtc, 7:acc->ccc
+ .  .  V  .  .  .  .  P  .  .  .  .  .  .  .  .  .  .  .  .
+.....................c.t.................................... - variant 2; 7:acc->cct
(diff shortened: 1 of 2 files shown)


mcouch-synthace:
Not ready for review yet.

m-i-sadowski:
> Possible off-by-one error, assuming the split point p is zero-offset https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L354. 

Yeah looks pretty irrefutable to me

m-i-sadowski:
>Also possible off-by-one: this function used to calculate distance between split points https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L448

Yes this is also out by one

m-i-sadowski:
>In Recombine - no sort before validation, but valid assumes splits are ordered (wasted samples?) https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L521
Similarly here - https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L655

Yes to both

m-i-sadowski:
>There is a normalisation of population fitness values in an innermost loop, possible speed up? https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L693

Might be worth it, the intention was to provide general swappable test functions here which is why the scaling is done in that way. Whether it's worth changing this depends on whether that test is OK or we want to try others I guess. If the latter we should settle on one first then fix

m-i-sadowski:
>Possible off-by-one error calculating segment length in cost (small effect) https://github.com/antha-lang/antha/blob/master/antha/AnthaStandardLibrary/Packages/Optimization/concreteOptimizer.go#L782

Agreed again

m-i-sadowski:
>It seems that function endsOK checks for duplicates in the 5' overhang set, but not for whether any overhangs could be reverse complements of each other.

I'd always presumed this was getting into score function territory: I can certainly believe this is potentially bad but I've always assumed the the main consequence is that some fragments might end up getting truncated by annealing to the restriction-site-containing side of the chop, the other possibility being sterically hindered by the (usually much larger) sequences on the other side. In that case you would get failures but not misassemblies and up to a point this is tolerable.

I have to confess to being highly uncertain on this point though - if you know (or find out) better then by all means fix it.

m-i-sadowski:
This is excellent stuff - great to finally review this and fix the issues, also nice to see all these comments and explanations, which are all spot on. 

I'd suggest adding a few inline comments on the structures to clarify these points (now that they have been clarified) so that it's easy to see in the code what assumptions are made.

mcouch-synthace:
Skipping known fails, will re-enable one by one with fix.

PR #984, opened 2019-02-27
switching to Go 1.11.5

levitskovich:
Go 1.11 has a linter bug (https://github.com/golangci/golangci-lint/issues/148) which results in a fake warning message which breaks feature/datatable branch build


msackman-synthace:
Fine by me provided it builds.

PR #985, opened 2019-02-27
Explicitly record the size of positions on the deck

haydnKing:
Previously we've always assumed that deck positions are "SBS format" (i.e. 127.76x85.48 mm) since this is always the case with gilson.

Hamilton has several pieces of labware which aren't (the first of which being the tipwaste). This breaks the physical simulator, because it gets upset when you try to place an obviously non-SBS format piece of labware on a position which (it assumes) is SBS format.

This change adds a "Size" property to LHPosition, and requires the various liquidhandler drivers to specify the size of the available positions when returning the robot state in `GetCapabilities`. Once this PR is opened, I'll open PRs in each of the drivers and reference this PR so they show up below.

Testing:
I've added a test in the simulator to demonstrate that a non-sbs plate can now be added to a robot with a non-sbs shaped deck position.

### DIFF
--- a/antha/anthalib/wtype/geometry.go
+++ b/antha/anthalib/wtype/geometry.go
@@ -189,6 +189,9 @@ func (self Coordinates2D) Abs() float64 {
 	return math.Sqrt(self.X*self.X + self.Y*self.Y)
 }
 
+// SBSFootprint the size of standard SBS format plates
+var SBSFootprint = Coordinates2D{X: 127.76, Y: 85.48}
+
 //a rectangle
 type Rectangle struct {
 	lowerLeft  Coordinates2D
--- a/antha/anthalib/wtype/geometry.go
+++ b/antha/anthalib/wtype/geometry.go
@@ -24,7 +24,6 @@ package wtype
 
 import (
 	"fmt"
-	"github.com/antha-lang/antha/antha/anthalib/wunit"
 	"math"
 )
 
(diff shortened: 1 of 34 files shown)


antha/anthalib/wtype/geometry.go
```diff
@@ -27,27 +27,27 @@ import (
 	"math"
 )
 
-type Coordinates struct {
+type Coordinates3D struct {
 	X float64
 	Y float64
 	Z float64
 }
```
msackman-synthace: interesting discussion topic: should it be:

```go

type Coordinates3D struct {
  Coordinates2D
  Z float64
}
```

?

(not requesting this change... I don't even know what I think)

antha/anthalib/wtype/geometry.go
```diff
@@ -27,27 +27,27 @@ import (
...
 
 //String implements Stringer
-func (self Coordinates) String() string {
+func (self Coordinates3D) String() string {
 	return fmt.Sprintf("%.1fx%.1fx%.1f mm", self.X, self.Y, self.Z)
 }
 
-func (self Coordinates) StringXY() string {
+func (self Coordinates3D) StringXY() string {
 	return fmt.Sprintf("%.1fx%.1f mm", self.X, self.Y)
```
msackman-synthace: heh, in light of the above, this could then become `coord.Coordinates2D.String()` and thus avoid the method dupe... hmm.

antha/anthalib/wtype/geometry.go
haydnKing: Hmm, I'd argue it ideally shouldn't embed, since if you have a R3 coordinate there's an infinite number of ways you might go about projecting it into R2 .
Ideally I think there should be something like `Coordinates3D.Project(normal Coordinates3D) Coordinates2D` (possibly with a helper to project directly along the Z access) - providing `thing.Coordinates2D` doesn't seem as clear to me

PR #986, opened 2019-03-01
Improvements to behaviour of InSequences func  

chr1sgrant:
(updated description)
1. Change default behaviour from JUST checking the sequence itself for equality to checking both name and sequence for equality.
2. Add new options to IgnoreSequence, IgnoreName and  ExacctMatch
3. Add InSet and EqualFold funcs to sequences package for matching robust to orientation and rotation.

https://synthace.atlassian.net/browse/ANTHA-2617


mcouch-synthace:
Is this right?

Current behaviour:

1. Default (no options passed)
Sequence bases must match (trimmed, case sensitive), plasmid ignored, names ignored
2. Options `IgnoreCase`
Sequence bases must match (no trimming, case insensitive), plasmid must match, names ignored
3. Options `IgnoreCase, MatchName`
Sequence bases ignored, plasmid ignored, names must match (no trimming, case insensitive)
4. Options `MatchName`
Sequence bases ignored, plasmid ignored, names must match (trimmed, case sensitive)

Odd that `IgnoreCase` on its own requires plasmid to match.
Odd that `IgnoreCase` on its own or with `MatchName` switches off trimming.

But these outside scope of PR - ignoring option `IgnoreCase` and plasmid checking from here.

Surprising that specifying `MatchName` switches OFF sequence checking. The documentation for this option in `Option.go` suggests that passing `MatchName` applies an additional requirement - additional to the sequence check.

This is modified in the PR:

New behaviour:
1. Default (no options passed)
Sequence bases must match, names must match
2. Options `MatchName`
Sequence ignored, name must match
3. Options `MatchSequence`
Sequence bases must match, name ignored
4. Options `MatchName, MatchSequence`
Nothing will be matched

The default behaviour is changing. The documentation clearly reflects that the default behaviour is changing.

Perhaps the documentation could explain that passing `MatchName` will switch off sequence matching.  That would mirror the wording for `MatchSequence`. 

Would `JustMatchName` be clearer?

Is the option combination `MatchName, MatchSequence` valid?

mcouch-synthace:
What about alternative set of options `IgnoreCase, IgnoreSequence, IgnoreName`.
These could be used in any combination and their effect with respect to default (case sensitive match to both sequence and name) would be obvious.

chr1sgrant:
I like the suggestion of everything being specified as Ignore, that would be a lot clearer; I'll ggo ahead and implement those changes. Thanks

mcouch-synthace:
Yes that's now clearer for me too.

You have a comment saying plasmid is ignored [here](https://github.com/antha-lang/antha/blob/8bc8260a8ddbfba2898da5db42c09a504ec7d265/antha/AnthaStandardLibrary/Packages/search/InSequences.go#L45) however plasmid is still checked [here](https://github.com/antha-lang/antha/blob/8bc8260a8ddbfba2898da5db42c09a504ec7d265/antha/AnthaStandardLibrary/Packages/search/InSequences.go#L68).

chr1sgrant:
Ahh, good catch, I'll sort that

chr1sgrant:
could also be a case for adding an option `Exact` which just uses reflect.DeepEqual

antha/AnthaStandardLibrary/Packages/sequences/findSeq_test.go
```diff
@@ -223,6 +224,131 @@ var regionTests = []regionTest{
...
 
+func TestEqualFold(t *testing.T) {
+
+	type equalityTest struct {
+		A, B           *wtype.DNASequence
+		ExpectedResult bool
+	}
+
+	tests := []equalityTest{
+		equalityTest{
```
mcouch-synthace: Note `wtype.RevComp` which may be called in `FindAll` upper-cases the sequence before calling `Comp`.
So it might be worth considering something like
```
                equalityTest{
			A: &wtype.DNASequence{
				Nm:      "case change",
				Seq:     "GATcgtAGTGT",
				Plasmid: true,
			},
			B: &wtype.DNASequence{
				Nm:      "revcomp",
				Seq:     wtype.RevComp("GATCGTAGTGT"),
				Plasmid: true,
			},
			ExpectedResult: true,
		},
```

antha/AnthaStandardLibrary/Packages/search/Option.go
```diff
@@ -44,16 +44,22 @@ const (
...
 	// IgnoreSequence is an option which can be added to the InSequences
-	// functions to specify that matching not include mathcing on DNA Sequence.
+	// function to specify that matching not include matching on DNA Sequence.
 	// By default sequence searches check both sequence and name equality.
 	//
 	IgnoreSequence Option = "IgnoreSequence"
+
+	// ExactMatch is an option which can be added to the InSequences function
+	// to specify that all properties of the DNASequence must match to be
+	// classified as a match.
```
mcouch-synthace: Might worth adding here also that it overrides other options.

antha/AnthaStandardLibrary/Packages/sequences/findSeq_test.go
chr1sgrant: OK the FindAll function currently casts both to upper case so that case should be covered by the existing test for case sensitivity.

chr1sgrant:
One final change to renam the sequences.InSet func to sequences.InSequences in order to be consistent with the InSequences func in the search package.

mcouch-synthace:
@chr1sgrant file deleted?

chr1sgrant:
gosh, not sure what happened there; fixed that now

PR #987, opened 2019-03-01
return true if comparing equality of two nil measurements

chr1sgrant:


antha/anthalib/wunit/wunit.go
```diff
@@ -309,6 +310,9 @@ func (cm *ConcreteMeasurement) GreaterThanRounded(m Measurement, p int) bool {
 func (cm *ConcreteMeasurement) EqualToRounded(m Measurement, p int) bool {
 	// this is not equal to anything
```
haydnKing: comment appears to contradict the behaviour now?

chr1sgrant:
But surely if two values are both nil then they are equal so should return true?

That's the behaviour I expected when calling this.

PR #966, opened 2019-02-06
Unify carry vol

m-i-sadowski:
This PR 

- makes the carry volume a single variable defined in wtype
- unifies access to the carry volume via LHProperties, which is the proper place

I haven't changed its value, nor implemented the more realistic version which should be added in future. This is a simple refactor to tidy things up. 

It has been tested with antha-tester successfully against bundles currently in elements-westeros and looks fine

PR #988, opened 2019-03-05
add MixerWait intrinsic

chr1sgrant:


microArch/scheduler/liquidhandling/executionhelpers.go
```diff
@@ -155,14 +156,23 @@ func appendSensitively(iar [][]*wtype.LHInstruction, ins *wtype.LHInstruction, t
...
 
+// need to do this with wait time too
 func aggregatePromptsWithSameMessage(inss []*wtype.LHInstruction, topolGraph graph.Graph) []graph.Node {
+
 	// merge dependencies of any prompts which have a message in common
-	prMessage := make(map[string][][]*wtype.LHInstruction, len(inss))
+
+	type prompter struct {
+		Message  string
+		WaitTime wunit.Time
```
m-i-sadowski: IIRC wunit.Time uses floats for time encoding; this leads to the likelihood of this stuff creating multiple waits in sequence with the same message and (effectively) time because of rounding

I'd be inclined to round to whole seconds *at least* this early, however it's possible it should be done earlier

m-i-sadowski: separately I really think we should probably just alias wunit.Time to time.Duration internally by some means

antha/anthalib/wtype/lhinstruction.go
```diff
@@ -48,6 +48,7 @@ type LHInstruction struct {
 	PlateName        string
 	OutPlate         *Plate
 	Message          string
+	WaitTime         wunit.Time
```
m-i-sadowski: See my later comment, this might be better as something like time.Duration or just an int64 with a sensible resolution (seconds is tempting but we could go lower)

microArch/scheduler/liquidhandling/lhrequest.go
```diff
@@ -64,7 +64,8 @@ type LHRequest struct {
 	OutputOrder           []string
 	OutputIteratorFactory func(wtype.Addressable) wtype.AddressIterator `json:"-"`
 	InstructionChain      *wtype.IChain
-	TimeEstimate          float64
+	TimeEstimate          float64 // in seconds
+	CarryVolume           wunit.Volume
```
m-i-sadowski: CarryVolume was intentionally removed from LHRequest a few weeks back, please delete this (see https://github.com/antha-lang/antha/pull/966)

microArch/scheduler/liquidhandling/executionhelpers.go
chr1sgrant: ahh cool, already there in wunit/wdimension.go: 

```
func (t Time) AsDuration() time.Duration {
	// simply use the parser
	if d, err := time.ParseDuration(t.ToString()); err != nil {
		panic(err)
	} else {
		return d
	}
}

func FromDuration(t time.Duration) Time {
	return NewTime(float64(t.Seconds()), "s")
}
```

chr1sgrant: although that will set the precision to 3 dp of a second so I'll use a custom version.


I've just refactored to use time.Duration and set precision to nearest second at the earliest step (MixerWait).

m-i-sadowski:
Nice, think this looks OK now 

Can you make some brief comments about how you've tested this?

chr1sgrant:
**Testing Performed **
1. Tested by running the attached workflow and physically testing on a Gilson PipetMax robot.

The test composed of a run which pipetted some liquid from a to b and the running the mixer wait command to wait for 10 seconds before resuming with a subsequent transfer from b to c.

Furthermore the liquid policies for the two transfer steps were such that the aspirate and dispense wait were also tested to ensure that the implementation of MixerWait hadn't affected those existing wait sub commands. 

The behaviour witnessed was as expected.    

antha-ninja:
https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/731/diffs

antha run --driver go://github.com/Synthace/PipetMaxDriver/server --bundle waitTest3.json --policyFile Liquid_Policy_Template_File.xlsx

[waitTest3.json.txt](https://github.com/antha-lang/antha/files/2952025/waitTest3.json.txt)
[Liquid_Policy_Template_File.xlsx](https://github.com/antha-lang/antha/files/2952026/Liquid_Policy_Template_File.xlsx)

2. A secondary test was also performed which refactored the assembly and transformation elements to offer the option to AutoResume with the MixerWait command rather that MixerPrompt.
The behaviour of this workflow was also as expected.

https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/733

microArch/driver/liquidhandling/messageinstruction.go
```diff
@@ -41,16 +47,43 @@ func (msi *MessageInstruction) GetParameter(name InstructionParameter) interface
...
+		}
+		ret = lowLevelDriver.Wait(waitins.Time)
+
+		if !ret.Ok() {
+			return fmt.Errorf(" %d : %s", ret.ErrorCode, ret.Msg)
+		}
+
+	}
 
 	if msi.Message != wtype.MAGICBARRIERPROMPTSTRING {
```
m-i-sadowski: I think this needs some consideration - the idea is to make "wait message" the way to get a timed wait, so if the above is set we should always be outputting a message. 

If we see a time with MAGICBARRIERPROMPTSTRING it should be an error

microArch/scheduler/liquidhandling/executionhelpers.go
```diff
@@ -155,14 +157,23 @@ func appendSensitively(iar [][]*wtype.LHInstruction, ins *wtype.LHInstruction, t
 	return iar
 }
 
+// need to do this with wait time too
 func aggregatePromptsWithSameMessage(inss []*wtype.LHInstruction, topolGraph graph.Graph) []graph.Node {
```
m-i-sadowski: I think this needs a unit test

microArch/driver/liquidhandling/messageinstruction.go
```diff
@@ -41,16 +47,43 @@ func (msi *MessageInstruction) GetParameter(name InstructionParameter) interface
...
 	case MESSAGE:
 		return msi.Message
+	case WAIT:
+		return msi.WaitTime
 	default:
 		return msi.BaseRobotInstruction.GetParameter(name)
 	}
 }
 
 func (msi *MessageInstruction) OutputTo(driver LiquidhandlingDriver) error {
```
m-i-sadowski: I think this needs a unit test

chr1sgrant: hmm, I can't see an obvious way to do this and there doesn't seem to be any precedent for testing the driver commands produced from an OutputTo method of an instruction; do you have a suggestion on the best way to do this?

chr1sgrant: particularly since the OutputTo method doesn't have a returned output which can be evaluated (apart from an error).

microArch/driver/liquidhandling/messageinstruction.go
chr1sgrant: this is resolved now

microArch/driver/liquidhandling/messageinstruction.go
m-i-sadowski: Yes I realise the issue here is that we shouldn't be encoding logic into OutputTo 

What I suggest is that this becomes another instruction layer which produces the appropriate TerminalRobotInstructions via Generate

microArch/driver/liquidhandling/messageinstruction_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"github.com/antha-lang/antha/antha/anthalib/wutil/text"
+)
+
+// Some potentially confusing logic in need of testing
```
m-i-sadowski: Y E S !

PR #990, opened 2019-03-06
Switch linters to golangci-lint

haydnKing:
`gometalinter` is dead. See: https://github.com/alecthomas/gometalinter/issues/590

`golangci-lint` is the new kid on the block. See: https://github.com/golangci/golangci-lint

This PR swaps out one for the other, and fixes a bunch of self-contained linter errors spotted by the new tool. Mostly they relate to test code (which was apparently barely linted previously), but there are some more interesting cases, e.g.
- HTTP requests with arbitrary strings
- not storing the result of type-switches in a variable
- using `strings.ToLower(a) == strings.ToLower(b)` instead of `strings.EqualFold(a, b)`


antha/AnthaStandardLibrary/Packages/enzymes/Ligation.go
```diff
@@ -644,7 +644,8 @@ func MultipleAssemblies(parameters []Assemblyparameters) (s string, successfulas
 			if strings.Contains(err.Error(), "Failure Joining fragments after digestion") {
 				constructsitesstring := make([]string, 0)
 				constructsitesstring = append(constructsitesstring, output)
-				enzyme, err := lookup.RestrictionEnzyme(construct.Enzymename)
```
m-i-sadowski: I assume this was an issue with variable shadowing?

antha/AnthaStandardLibrary/Packages/igem/igem.go
```diff
@@ -143,39 +143,19 @@ func MakeFastaURL(partname string) (Urlstring string) {
 	return Urlstring
 }
 
-func MakeXMLURL(partnames []string) (Urlstring string) {
```
m-i-sadowski: Dead code?

antha/AnthaStandardLibrary/Packages/pubchem/pubchem.go
```diff
@@ -186,21 +186,13 @@ func makeOutputspec(spec string, optionalcallbackname string) (outputspec string
...
-	if operation_options != "" {
-		array = append(array, operation_options)
-	}
-	Urlstring := strings.Join(array, "/")
 	/*<input specification>/<operation specification>/[<output specification>][?<operation_options>]
 
 	http://pubchem.ncbi.nlm.nih.gov/rest/pug/compound/name/glucose/property/MolecularFormula,MolecularWeight/JSON
 	*/
-	res, err := http.Get(Urlstring)
+	res, err := http.Get(fmt.Sprintf("http://pubchem.ncbi.nlm.nih.gov/rest/pug/%s/%s/%s/%s", inputspec, operationspec, outputspec, operation_options))
```
m-i-sadowski: What is its issue here?

microArch/driver/liquidhandling/robotinstruction.go
```diff
@@ -121,54 +121,54 @@ func (name InstructionParameter) String() string {
...
-	VOLUNT                               = "VOLUNT"
-	WELL                                 = "WELL"
-	WELLFROM                             = "WELLFROM"
-	WELLFROMVOLUME                       = "WELLFROMVOLUME"
-	WELLTO                               = "WELLTO"
-	WELLTOVOLUME                         = "WELLTOVOLUME" // WELLTOVOLUME refers to the volume of liquid already present in the well location for which a sample is due to be transferred to.
-	WELLVOLUME                           = "WELLVOLUME"
-	WHAT                                 = "WHAT"
-	WHICH                                = "WHICH" // WHICH returns the Component IDs, i.e. representing the specific instance of an LHComponent not currently implemented.
+	CHANNEL         InstructionParameter = "CHANNEL"
```
m-i-sadowski: I'm not convinced this is an improvement

antha/AnthaStandardLibrary/Packages/enzymes/Ligation.go
haydnKing: yeah, later on in this function there was a shadowed err that was never checked

antha/AnthaStandardLibrary/Packages/pubchem/pubchem.go
haydnKing: the whole of the issue here is that `gosec` really doesn't like `http.Get(randomString)`... which is perhaps a bit overkill, but given that this is callable from elements I figured it was a good prompt to tidy up here a bit

microArch/driver/liquidhandling/robotinstruction.go
haydnKing: yeah, me neither to be honest... can look into whether I can specifically disable this option

antha/AnthaStandardLibrary/Packages/pubchem/pubchem.go
m-i-sadowski: If it were likely to happen often I'd suggest this is a candidate for switching off. As it is it doesn't seem terribly likely to be a huge issue

microArch/driver/liquidhandling/robotinstruction.go
m-i-sadowski: OK so as you pointed out, https://staticcheck.io/docs/checks#SA9004, proves I'm entirely wrong! 

_slinks off_

PR #991, opened 2019-03-11
turn off blowout if blowout volume is zero

chr1sgrant:
https://synthace.atlassian.net/browse/ANTHA-2585


chr1sgrant:
It looks like that will have the same effect as setting RESET_OVERRIDE  to true in a policy, which the job ticket may be problematic: https://synthace.atlassian.net/browse/ANTHA-2585

haydnKing:
I thought the goal was to replicate the behaviour of RESET_OVERRIDE == true? - the only issue is that it's not in the default set of policies? The current solution will look like a blowout of 0ul as far as the mix graph is concerned

(btw. EXTRA_ASP_VOLUME is not supported, mainly because of the issue described)

jnunoleitao:
I have tested a combination of 72 LHPs in a trilution simulation, with different combinations of `ExtraAsp`, `BlowoutVolume`, and `PostMix`. Simulation was successful.

I can confirm that if `BlowoutVolume==0` there's no "Purge" in the Trilution simulator (this was confirmed in the lab as well when Chris first pushed changes to this branch).

Also, no volume is building up in the tip when using `ExtraAsp`, which would happen when before we were using `ResetOverride` as the only way to prevent a blowout.

chr1sgrant:
@haydnKing can we merge this in since it looks like it unlocks the use of EXTRA_ASP_VOLUME in the lab?

haydnKing:
I really think generating a blowout instruction which then magically does nothing is the wrong way of implementing this. Instead we should only generate instructions when they actually have an effect, e.g.
```
if blowVolume > 0.0 {
  blo := NewBlowoutInstruction()
  ...
  ret = append(ret, blo)
}
```
**

> Also, no volume is building up in the tip when using ExtraAsp, which would happen when before we were using ResetOverride as the only way to prevent a blowout.

@jnunoleitao So what _is_ happening to the `EXTRA_ASP_VOLUME`? My guess is that it _does_ get blown out into the well by the piston reset instruction that happens after the blowout/purge, though it's also possible for it to be sucked up into the head during piston reset which would be bad.

To reiterate: `EXTRA_ASP_VOLUME` isn't supported and is likely to open a pandora's box of issues and absolutely absolutely absolutely shouldn't be being used for clients.

jnunoleitao:
Regarding your question @haydnKing , I have explicitely set in my DoE model not to have `Extra_Asp_Vol` and `BlowoutVolume`. I believe the extra volume is discarded with the tip.

I'm fine with not supporting `Extra_Asp_Vol` atm if we really can't, but right now we also don't properly support not having a blowout, apart from using Reset_Override, which is a problem because clients have access to this and it's unclear what the consequences are if `ResetOverride` is set. If `BlowoutVolume==0` there should be no blowout, which is currently not the case.

PR #992, opened 2019-03-11
Antha 2539/old to new bundles

matttgregg:
Still working through the merge logic, but can you take a look and see if this looks sane structurally?


workflow/workflow.go
```diff
@@ -53,6 +59,15 @@ func (wf *Workflow) WriteToFile(p string) error {
 	}
 }
 
+func (wf *Workflow) WriteTo(o io.WriteCloser) (int, error) {
+	defer o.Close()
+	if bs, err := json.Marshal(wf); err != nil {
```
msackman-synthace: Normally in Go, I would expect a var which is a writer of some sort to be called `w` or at least start with `w`. I'm sure you mean `o` as in `output`, but it's slightly non-idiomatic.

However, given you've got a Writer here, I doubt you need to do the serialisation in memory. I think you should be able to use json.Encoder and friends to avoid the intermediate buffer. Again, not super critical, but it might make the api simpler because you would have the mirror of the reading version - and so you just need to do pathToWriter (e.g. os.CreateFile or os.Open with suitable flags), and then can pass in either Stdout or the file handle to the same write function.

workflow/v1point2/config.go
```diff
@@ -0,0 +1,43 @@
+package v1point2
```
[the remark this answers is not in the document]
matttgregg: I prefer `v1_2` too, but whichever linting tool VSCode runs for me says `don't use an underscore in a package name` - I'm happy to take style from you rather than the tool though.

workflow/v1point2/migrater.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	return nil
+}
+
+func (m *Migrater) migrateParameters() error {
+	m.BaseWorkflow.JobId = workflow.JobId(m.OldWorkflow.Properties.Name)
+
+	if m.BaseWorkflow.Meta.Rest == nil {
+		m.BaseWorkflow.Meta.Rest = make(map[string]interface{})
```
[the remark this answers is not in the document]
matttgregg: I like that idea - I felt sad initialising it from outside.

workflow/repositories.go
```diff
@@ -177,3 +177,33 @@ func (r *Repository) ensureGitRepo() error {
...
+func (r *Repository) BuildElementTypeMap(prefix RepositoryPrefix) (*ElementTypeMap, error) {
+	seen := make(map[string]struct{})
+	etm := make(ElementTypeMap)
+
+	r.Walk(func(f *File) error {
+		if filepath.Ext(f.Name) != ".an" {
+			return nil
+		}
+
+		dir, _ := filepath.Split(f.Name)
```
msackman-synthace: I think you want filepath.Dir

workflow/repositories.go
```diff
@@ -177,3 +177,33 @@ func (r *Repository) ensureGitRepo() error {
...
 }
+
+type ElementTypeMap map[ElementTypeName]ElementType
+
+func (r *Repository) BuildElementTypeMap(prefix RepositoryPrefix) (*ElementTypeMap, error) {
+	seen := make(map[string]struct{})
+	etm := make(ElementTypeMap)
+
+	r.Walk(func(f *File) error {
+		if filepath.Ext(f.Name) != ".an" {
```
msackman-synthace: Sorry, I've been lazy here - it's clear we should have an `IsAnthaFile(path string) bool` predicate and use that consistently.

workflow/repositories.go
```diff
@@ -177,3 +177,33 @@ func (r *Repository) ensureGitRepo() error {
...
+			return nil
+		}
+
+		dir, _ := filepath.Split(f.Name)
+		ename := filepath.Base(dir)
+
+		if _, s := seen[dir]; s {
```
msackman-synthace: the bool you've named `s` I always name `found`

On the other hand, there's probably no need for the seen set given that etm is a map anyway - it's really not going to save any measurable time

workflow/repositories.go
```diff
@@ -177,3 +177,33 @@ func (r *Repository) ensureGitRepo() error {
...
+			return nil
+		}
+		seen[dir] = struct{}{}
+
+		etm[ElementTypeName(ename)] = ElementType{
+			ElementPath:      ElementPath(filepath.Clean(dir)),
```
msackman-synthace: if you use filepath.Dir, the docs do claim it gets cleaned too, so you may be able to drop the call to Clean here.

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,267 @@
...
+func (m *Migrater) migrateGlobalMixerConfig() {
+	generic := workflow.GlobalMixerConfig{}
+	generic.CustomPolicyRuleSet = m.Old.Config.CustomPolicyRuleSet
+	generic.IgnorePhysicalSimulation = m.Old.Config.IgnorePhysicalSimulation
+	generic.InputPlates = m.Old.Config.InputPlates
+	generic.OutputPlates = m.Old.Config.OutputPlates
+	generic.PrintInstructions = m.Old.Config.PrintInstructions
+	generic.UseDriverTipTracking = m.Old.Config.UseDriverTipTracking
+	m.Cur.Config.GlobalMixer = generic
+}
```
msackman-synthace: I'm afraid I really don't like this syntactically - I would write:

```go
m.Cur.Config.GlobalMixer = &workflow.GlobalMixerConfig{
  CustomPolicyRuleSet: ...,
  ...
}
```

Visually looks much less noisy to me.

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,267 @@
...
+
+func (m *Migrater) migrateLayoutPreferences() *workflow.LayoutOpt {
+	lp := workflow.LayoutOpt{}
+	lp.Inputs = m.Old.Config.DriverSpecificInputPreferences
+	lp.Outputs = m.Old.Config.DriverSpecificOutputPreferences
+	lp.Tipboxes = m.Old.Config.DriverSpecificTipPreferences
+	lp.Tipwastes = m.Old.Config.DriverSpecificTipWastePreferences
+	lp.Washes = m.Old.Config.DriverSpecificWashPreferences
+	return &lp
```
msackman-synthace: same here, I would always do

```go
return &workflow.LayoutOpt{
  Inputs: ...,
  ...
}
```

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,267 @@
...
+func updatePlateTypes(names []string) []wtype.PlateTypeName {
+	ptnames := make([]wtype.PlateTypeName, len(names))
+	for i, v := range names {
+		ptnames[i] = wtype.PlateTypeName(v)
+	}
+	return ptnames
+}
+
+func (m *Migrater) migrateGilsonConfig() *workflow.GilsonPipetMaxInstanceConfig {
+	config := workflow.GilsonPipetMaxInstanceConfig{}
```
msackman-synthace: and this one, sorry!

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+		m.Cur.Elements.Instances[name] = ei
+	}
+
+	return nil
+}
+
+type multiRepositoriesMap map[*workflow.Repository]*workflow.ElementTypeMap
+
+func uniqueElementType(mrm *multiRepositoriesMap, name workflow.ElementTypeName) (*workflow.ElementType, error) {
```
msackman-synthace: hmm, a map is internally a pointer value anyway, so there should never ever be any reason to pass around a pointer to a map. This is definitely something Go has done a very bad job at documenting - the difference between a slice (which is a raw value) and a map (which is a pointer).

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,267 @@
...
+	var et *workflow.ElementType
+
+	for _, rmap := range *mrm {
+		if v, found := (*rmap)[name]; found {
+			if et != nil {
+				return nil, errors.New("element type " + string(name) + " is found in multiple repositories")
```
msackman-synthace: Personally, I always try to avoid string concatenation. Probably for no good reason, but I guess drilled into me from SQL work and such. So I would always default to `fmt.Errorf(...)`. That way you could also use `%v` and drop the `string(...)` cast

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+
+	if et == nil {
+		return nil, errors.New("element " + string(name) + " could not be found in the supplied repositories")
+	}
+	return et, nil
+}
+
+func (m *multiRepositoriesMap) writeMultiRepositoriesMap(w io.Writer) {
```
msackman-synthace: I assume this is for debugging - it appears to be dead code?

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,267 @@
...
+			return err
+		}
+
+		repoMaps[rep] = rmap
+	}
+
+	ets := make(workflow.ElementTypes, 0, len(m.Old.Processes))
+
+	for _, v := range m.Old.Processes {
+		if _, s := seen[v.Component]; s {
```
msackman-synthace: again, `found` is more idiomatic.

cmd/migrate/main.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if err := m.Validate(); err != nil {
+		if ignoreValidation {
+			logger.Log("ValidationError", err)
+		} else {
+			logger.Fatal(err)
+		}
+	}
+
+	m.WriteToPath(toFile)
```
msackman-synthace: this can return an error so you probably want to test that.

workflow/config.go
```diff
@@ -40,6 +40,15 @@ type GilsonPipetMaxConfig struct {
 	Devices  map[DeviceInstanceID]*GilsonPipetMaxInstanceConfig `json:"Devices,omitempty"`
 }
 
+// InsertDeviceConfig inserts or updates config for a device instance
+func (c *GilsonPipetMaxConfig) InsertDeviceConfig(d DeviceInstanceID, ic *GilsonPipetMaxInstanceConfig) {
```
msackman-synthace: I don't like this. But let's come back to this before the end of this work and we'll figure out what to do here. I'd prefer a solution in which we can never accidentally end up with a nil Devices. Probably explicit constructions maybe via the NewWorkflow method...?

workflow/repositories.go
matttgregg: Yes - I'll pull it out. I think that I had an idea that I was going to do some checks (i.e. warn on same type defined in multiple places) but maybe not the right place to do that.

workflow/repositories.go
matttgregg: Looks good!

workflow/config.go
matttgregg: Yes :) Seems cleaner to initialise empty maps in the appropriate places rather than nil.

workflow/v1_2/migrater.go
matttgregg: I've left these as is due to them all being promoted fields, which prevents me doing the cleaner struct initialisation.

msackman-synthace: Oh! I see. Yes, I take your point.

cmd/migrate/main.go
```diff
@@ -0,0 +1,58 @@
...
+		flag.PrintDefaults()
+		fmt.Fprintf(flag.CommandLine.Output(), "All further args are interpreted as paths to up to date workflows to be merged and used as a basis for the upgrade.\n")
+	}
+
+	var fromFile, toFile, gilsonDevice string
+	var ignoreValidation bool
+	flag.StringVar(&toFile, "to", "", "File to write to (default: will write to stdout)")
+	flag.StringVar(&fromFile, "from", "", "File to migrate from (default: will be read from stdin)")
+	flag.StringVar(&gilsonDevice, "gilson-device", "", "A gilson device name to use. Optional. (May be taken from the merge files if not present.)")
+	flag.BoolVar(&ignoreValidation, "ignore-validation", false, "Ignore validation for the generated workflow.")
```
msackman-synthace: Interesting. I tend to dislike negatives in flag names. I think if it was me I would do "validate" with a default of true. "Ignore" isn't so bad as "not" or "no", but possibly worth thinking about.

cmd/migrate/main.go
```diff
@@ -0,0 +1,58 @@
...
+		} else {
+			logger.Fatal(err)
+		}
+	}
+
+	if err := m.MigrateAll(); err != nil {
+		logger.Fatal(err)
+	}
+
+	if err := m.Validate(); err != nil {
```
msackman-synthace: If you call it ValidateOld then I would think this should be ValidateNew maybe?

cmd/migrate/main.go
```diff
@@ -0,0 +1,58 @@
...
+		if ignoreValidation {
+			logger.Log("ValidationError", err)
+		} else {
+			logger.Fatal(err)
+		}
+	}
+
+	err = m.WriteToPath(toFile)
```
msackman-synthace: I'd inline into the `if` that follows myself.

workflow/repositories.go
```diff
@@ -182,3 +182,29 @@ func (r *Repository) ensureGitRepo() error {
...
 	return nil
 }
+
+func IsAnthaFile(path string) bool {
+	return strings.EqualFold(filepath.Ext(path), ".an")
+}
+
+type ElementTypeMap map[ElementTypeName]ElementType
+
+func (r *Repository) BuildElementTypeMap(prefix RepositoryPrefix) (*ElementTypeMap, error) {
```
msackman-synthace: The name doesn't suggest which elements it finds. I think a name such as "FindAllElementTypes" would be slightly more intuitive.

workflow/readers.go
```diff
@@ -51,3 +51,11 @@ func JsonPathsWithin(dir string) ([]string, error) {
 		return results, nil
 	}
 }
+
+func ReaderFromPath(path string) (io.ReadCloser, error) {
```
msackman-synthace: this is now dead code and can be removed.

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,251 @@
...
+	m.migrateGilsonConfigs()
+	return nil
+}
+
+func (m *Migrater) migrateGilsonConfigs() {
+	c := m.migrateGilsonConfig()
+	m.Cur.Config.GilsonPipetMax.Defaults = c
+
+	for d := range m.Cur.Config.GilsonPipetMax.Devices {
+		m.Cur.Config.GilsonPipetMax.InsertDeviceConfig(d, c)
```
msackman-synthace: oh wow, so you apply the config to *all* existing gilsons. I'm not sure that's intuitive behaviour. I think I would prefer "unless the gilson cmd line flag is provided, don't do anything to gilson configs"

Also, I'd probably do:
if there is already a device that has the same name as the cmd line flag, then log out a warning, and overwrite the config.

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,251 @@
...
+		if v, found := (*rmap)[name]; found {
+			if et != nil {
+				return nil, fmt.Errorf("element type %v is found in multiple repositories", name)
+			}
+			et = &v
+		}
+	}
+
+	if et == nil {
+		return nil, fmt.Errorf("element %v could not be found in the supplied repositories", name)
```
msackman-synthace: probably should be "element type %v..." rather than "element %v..." to match the other error text here.

workflow/v1_2/migrater.go
matttgregg: Ok - that makes sense given that we'll usually expect a Gilson to be given on the command line.

workflow/v1_2/migrater.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	return et, nil
+}
+
+func (m *Migrater) migrateElementTypes() error {
+	repoMaps := make(multiRepositoriesMap)
+	seen := make(map[string]struct{}, len(m.Old.Processes))
+
+	for prefix, rep := range m.Cur.Repositories {
```
msackman-synthace: this is potentially reusable code so I would be tempted to attach "FindAllElementTypes" to the Repositories (plural) type and move these first 9 lines out (apart from the `seen`, if you see what I mean :) )

workflow/v1_2/migrater.go
msackman-synthace: EDIT: oh hang on, Cur here is a fresh empty, so you can't possibly have a collision. And fact that means the loop can't do anything at all, can it?

workflow/v1_2/workflow.go
```diff
@@ -0,0 +1,107 @@
...
+	Outputs      map[string]*wtype.Plate
+	TimeEstimate time.Duration
+}
+
+func (wf *workflowv1_2) MigrateElementParameters(name string) (*workflow.ElementParameterSet, error) {
+	v, pr := wf.Parameters[name]
+	if !pr {
+		return nil, errors.New("parameters not present for element" + name)
+	}
+	pset := workflow.ElementParameterSet(make(map[workflow.ElementParameterName]json.RawMessage))
```
msackman-synthace: just `pset := make(workflow.ElementParameterSet)` no?

workflow/v1_2/workflow.go
```diff
@@ -0,0 +1,107 @@
...
+	MixTaskResults []mixTaskResult
+}
+
+type mixTaskResult struct {
+	Instructions liquidhandling.SetOfRobotInstructions
+	Outputs      map[string]*wtype.Plate
+	TimeEstimate time.Duration
+}
+
+func (wf *workflowv1_2) MigrateElementParameters(name string) (*workflow.ElementParameterSet, error) {
```
msackman-synthace: again, with a map, no need for the *

cmd/migrate/main.go
matttgregg: I've gone fo 'Cur' (given Old/Cur names elsewhere)

workflow/v1_2/migrater.go
matttgregg: Cur could have devices from the merged files (currently pulling in everything, not just repository information.)

msackman-synthace: Oh for some reason in my head I was thinking we did the merging after the migration not before.

msackman-synthace: ahh yes

```
	cwf, err := workflow.WorkflowFromReaders(rs[:len(rs)-1]...)
```

PR #994, opened 2019-03-13
Make ShapeTypeID a struct, harder to inadvertently define a broken one

haydnKing:
Previously, the shape type id was typedefd as a string. It was easy to inadvertently define a new object with a random shapetype (e.g. "trapezoid"), which then caused a crash if we attempted to access the IsRound method. 
Now both the name and roundedness are required up-front, so this should no longer be an issue.

Testing:
currently making minor changes to update the plugins then will run antha-tester


antha/anthalib/wtype/plate.go
```diff
@@ -1228,7 +1228,7 @@ func (p *Plate) AreWellTargetsEnabled(adaptorChannels int, channelSpacing float6
 		return false
 	}
 
-	if p.Welltype.Shape().ShapeName != "box" {
+	if !p.Welltype.Shape().Type.Equals(BoxShape) {
```
msackman-synthace: 100% minor: there's always an argument for doing the constant on the lhs because then you never have to deal with a nil receiver. So BoxShape.Equals(...) is fractionally less error prone.

antha/anthalib/wtype/shape.go
```diff
@@ -31,46 +31,64 @@ import (
...
+	return sti.Name
+}
+
+// IsRound returns whether or not the shape is rounded
+func (sti ShapeTypeID) IsRound() bool {
+	return sti.Round
+}
+
+// Equals returns true if the shape types are the same
+func (sti ShapeTypeID) Equals(rhs ShapeTypeID) bool {
```
msackman-synthace: yeah, from a maximum robustness pov, I'd probably do at least the arg as a pointer, and belt-and-braces would suggest the receiver too. In fact, if it were me, I'd try to base this on pointer equality of the object, not an embedded string (your constructor might need to be more robust (or package-private))

antha/anthalib/wtype/shape.go
```diff
@@ -31,46 +31,64 @@ import (
...
+	Type       ShapeTypeID
 	LengthUnit string
 	H          float64
 	W          float64
 	D          float64
 }
 
 func (sh *Shape) Equals(sh2 *Shape) bool {
-	return sh.ShapeName == sh2.ShapeName && sh.LengthUnit == sh2.LengthUnit && sh.H == sh2.H && sh.W == sh2.W && sh.D == sh2.D
+	return sh.Type == sh2.Type && sh.LengthUnit == sh2.LengthUnit && sh.H == sh2.H && sh.W == sh2.W && sh.D == sh2.D
```
msackman-synthace: surely you should be callign .Equals on the type?

antha/anthalib/wtype/shape.go
```diff
@@ -86,16 +104,16 @@ func (sh *Shape) Depth() wunit.Length { // Z?
 }
 
 func (sh *Shape) Dup() *Shape {
-	return &(Shape{sh.ShapeName, sh.LengthUnit, sh.H, sh.W, sh.D})
+	return &(Shape{sh.Type, sh.LengthUnit, sh.H, sh.W, sh.D})
```
msackman-synthace: make prettier pls.

antha/anthalib/wtype/shape.go
```diff
@@ -86,16 +104,16 @@ func (sh *Shape) Depth() wunit.Length { // Z?
...
 
 func (sh *Shape) String() string {
-	return fmt.Sprintf("%s [%fx%fx%f]", sh.ShapeName, sh.H, sh.W, sh.D)
+	return fmt.Sprintf("%s [%fx%fx%f]", sh.Type, sh.H, sh.W, sh.D)
 }
 
 func (sh *Shape) MaxCrossSectionalArea() (area wunit.Area, err error) {
 
-	shapename := strings.ToLower(string(sh.ShapeName))
+	shapename := strings.ToLower(string(sh.Type.Name))
```
msackman-synthace: no, this is gross, I can't let this through, sorry. This shapename == "circle" crap that follows. If you move to pointers you could cheaply do == comparison with the vars.

antha/anthalib/wtype/shape.go
msackman-synthace: Pointer version would require some care with deserialisation, but should be possible I think.

antha/anthalib/wtype/shape.go
haydnKing: good catch. this is exactly why "IsRound()" exists - there's no need to do any comparison at all

antha/anthalib/wtype/shape.go
```diff
@@ -24,53 +24,77 @@ package wtype
...
+var shapeTypeByName = make(map[string]*shapeType)
+
+// shapeTypeFromName get a shape type by string name, required to map from ms-inventory
+// returns nil if the name is unknown
+func ShapeTypeFromName(name string) *shapeType {
+	return shapeTypeByName[name]
+}
+
+var (
+	UnknownShape   = (*shapeType)(nil)
```
msackman-synthace: nice!

haydnKing:
the issue that this PR was intended to fix is sorted, but there seem to be unrelated errors

haydnKing:
for posterity: turned out that the pointer address of `wtype.Shape.Type` wasn't being preserved through serialise/deserialise, leading to bad pointer comparison. I've now removed some dead LHWell serialisation code and added `wtype.Shape.(Unm|M)arshalJSON` to convert to string or serialisation and set the correct object on deserialisation.

PR #998, opened 2019-03-15
add descending option to SortConcentrations and SortVolumes functions

chr1sgrant:


antha/anthalib/wunit/sort.go
```diff
@@ -25,18 +25,25 @@ package wunit
...
+// SortVolumes in order to sort by descending order.
+const DescendingOrder = true
+
 // SortConcentrations sorts a set of Concentration values.
 // An error will be returned if no values are specified or the base units of any of the concentrations are incompatible,
 // e.g. units of X and g/l would not be compatible.
-func SortConcentrations(concs []Concentration) (sorted []Concentration, err error) {
+// An optional second argument DescendingOrder can be specified to indicate that the sorting will be in descending order.
+// By default sorting will be in ascending order.
+func SortConcentrations(concs []Concentration, descending ...bool) (sorted []Concentration, err error) {
```
haydnKing: nice, I like this method of adding optional arguments :)

I'm not gonna insist, but it might be better to have 
```go
type SortOptions int
const DescendingOrder SortOptions = iota
...
func SortConcentrations(concs []Concentration, options ...SortOptions)
```
so we can more easily add more options if we need to

chr1sgrant: Yes I was trying to think about other options which might be relevant to sorting of units but was drawing a bit of a blank so just implemented as a standalone argument with a bool.  No harm in making it more future proof though; and also force the use of the constant as argument in elements which will make it more readable than just `true`.

chr1sgrant: implemented it as a string though to make it easier to print the expected value in the error message

PR #1000, opened 2019-03-18
Add version property to Layout and Actions summary for quick validation

haydnKing:
First: there was chaos.
Then: there was a schema + validation that we matched the schema.
Now: the schema has a version number we can increment if we make changes.


haydnKing:
obligatory "1000 GET"

microArch/scheduler/liquidhandling/schemas/actions.schema.json
```diff
@@ -2,14 +2,25 @@
...
-            { "$ref": "#/definitions/transferAction" },
-            { "$ref": "#/definitions/promptAction" },
-            { "$ref": "#/definitions/waitAction" }
-        ]
+    "type": "object",
+    "required": ["version", "actions"],
+    "properties": {
+        "version": {
+            "description": "the version of the schema, for validation",
+            "const": 1.0
```
msackman-synthace: I wouldn't use a number. Generally a string. Because really what a version is is a list of numbers with . as a separator - version comparison is `strings.Split(verA, ".") < strings.Split(verB, ".")` (to butcher a whole lotta apis/languages)

haydnKing: :headdesk: fixed

PR #1002, opened 2019-03-20
Return files of sub directories when accessing files from a job. 

chr1sgrant:


antha/AnthaStandardLibrary/Packages/jobfile/jobfile.go
```diff
@@ -93,6 +94,36 @@ type File struct {
 	IsDir bool
 }
 
+// dirsToFiles will walk through all sub directories of files and append the list of
+// files returned with files found in that sub directory and sub directories of that.
+func dirsToFiles(c *Client, files []*File) ([]*File, error) {
+	for _, f := range files {
+		if f.IsDir {
+			subDir := filepath.Join(f.Dir, f.Name)
```
msackman-synthace: filepath is for local files - i.e. on windows this will use the windows separator \ Here, you want / because it's webdav, so you want to be using "path", not "path/filepath"

antha/AnthaStandardLibrary/Packages/jobfile/jobfile.go
```diff
@@ -116,6 +148,8 @@ func (c *Client) ListFiles(ctx context.Context, jobID JobID) ([]*File, error) {
 		})
 	}
 
+	files, err = dirsToFiles(c, files)
```
msackman-synthace: just `return dirsToFiles(c, files)`

chr1sgrant:
ok thanks, I've made the suggested changes now

chr1sgrant:
I don't envisage a case where there will be a huge number of files so hopefully the performance won't be a problem.

PR #1003, opened 2019-03-20
ANTHA-2628 - Improve liquid tracking in physical simulator

haydnKing:
See https://synthace.atlassian.net/browse/ANTHA-2628

Once the low-level device independent instructions have been generated, the physical simulator replays them to verify their consistency and check for any physical collisions that could occur. The simulator is also used in `microArch/scheduler/liquidhandling/summary.go` to track liquids when generating the actions summary.

As well as ignoring carry volumes, the physical simulator previously used it's own liquid mixing function `addComponent` to add liquids into a well, modelled on `*LHWell.AddComponent`, as this cut down on simulation time since `*LHWell.AddComponent` was very memory intensive due to tracking of liquid history.

This change reverts the simulator to using the normal AddComponent interface, which means that subcomponents and liquid names are set in the same way as they are during element execution. CarryVolume is also accounted for.

Test bundles have been tested using antha-tester


microArch/simulator/liquidhandling/simulator.go
```diff
@@ -1601,36 +1613,18 @@ func (self *VirtualLiquidHandler) Mix(head int, volume []float64, platetype []st
...
-				continue
-			}
-			err = addComponent(wells[ch], com)
-			if err != nil {
-				self.AddErrorf("Unexpected well error - %s", err.Error())
-				continue
-			}
+		if tip := arg.adaptor.GetChannel(ch).GetTip(); !tip.IsEmpty() {
+			nonEmptyChannels = append(nonEmptyChannels, ch)
+			nonEmptyVolumes = append(nonEmptyVolumes, tip.CurrentVolume().MustInStringUnit("ul").RawValue())
```
matttgregg: Do we need to cast down to "ul" here? If just for reporting can we leave as a complete measurement and use the standard string version?

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -1601,36 +1613,18 @@ func (self *VirtualLiquidHandler) Mix(head int, volume []float64, platetype []st
 		}
 	}
 
-	//do the mixing
+	// tips should be empty
+	nonEmptyChannels := make([]int, 0, len(arg.channels))
+	nonEmptyVolumes := make([]float64, 0, len(arg.channels))
```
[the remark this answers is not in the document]
haydnKing: Yeah, I agree `map[int]wunit.Volume` would be cleaner as far as the code goes... the one argument in favour of doing it this way is that (foolishly) when writing the simulator I wanted to make the error messages as human friendly as possible and so the slices here are the correct format for `summariseChannels`/`summariseVolumes` used below

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -811,7 +814,8 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
 
 		if wells[i] == nil {
 			no_well = append(no_well, i)
-		} else if wells[i].CurrentWorkingVolume().LessThan(v) {
+		} else if wells[i].CurrentWorkingVolume().LessThan(v) && !wunit.SubtractVolumes(v, wells[i].CurrentWorkingVolume()).IsZero() {
```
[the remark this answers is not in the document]
haydnKing: I think it's a good idea - it's something that didn't quite make the cut when looking at `wunit` last time around

microArch/simulator/liquidhandling/simulator.go
haydnKing: again, it's just so it gets formatted nicely. I'm not too much of a fan of the pattern I've created here, but figure it's probably best to follow the convention within this package rather than end up with a mix of behaviour

matttgregg: Slightly tempted to suggest that there could be a version of `summariseVolumes` which works on volumes as well (share the code - only one version need do real work, the other can handle a float <-/-> volume transformation and pass it over.) But only very slightly, if you think it would be of benefit.

microArch/simulator/liquidhandling/simulator.go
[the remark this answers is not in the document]
msackman-synthace: ok, here we have `i < v && v - i != 0`. It seems surprising that the `v-i` step is needed. It would read nicer to be if this was `i < (v-ε)` maybe (I'm assuming that's the point here)?

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -823,12 +827,20 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
...
+		} else if err := tip.AddComponent(c); err != nil {
 			self.AddErrorf("%s: unexpected tip error \"%s\"", describe(), err.Error())
 		}
 	}
 
+	// silently remove the carry volume for each well - this may leave the well volume lower than the residual volume
+	// do this after removing everything else so carry volumes don't interfere with multichannel aspiration
+	for _, well := range wells {
+		if well != nil {
+			well.RemoveCarry(self.properties.CarryVolume())
```
msackman-synthace: well well well...

microArch/simulator/liquidhandling/simulator.go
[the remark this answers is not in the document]
haydnKing: Yeah, that is exactly the point. Unfortunately the units somewhat prevent defining a constant wunit.Epsilon and doing that directly... 
Ultimately I think this is a bit out of scope - elsewhere we do something like
```
if delta := wunit.SubtractVolumes(a, b); !delta.IsPositive() && !delta.IsZero() { 
    // a is meaningfully less than b
}
```
though I concede this still isn't great.

antha/anthalib/wunit/wunit.go
```diff
@@ -264,7 +267,7 @@ func (cm *ConcreteMeasurement) IsNil() bool {
 }
 
 func (cm *ConcreteMeasurement) IsZero() bool {
-	if isNil(cm) || math.Abs(cm.Mvalue) < 0.00000000001 {
+	if isNil(cm) || math.Abs(cm.Mvalue) < Epsilon {
 		return true
 	}
 	return false
```
msackman-synthace: I mean, obviously not assigning blame, but this isn't great.

`return isNil(cm) || math.Abs(cm.Mvalue) < Epsilon`

haydnKing: jesus I didn't even look

haydnKing: I'm actually a bit surprised this isn't caught by linting...

msackman-synthace: Humans are the best linters :) They just get grumpy too.

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -811,7 +814,8 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
 
 		if wells[i] == nil {
 			no_well = append(no_well, i)
-		} else if wells[i].CurrentWorkingVolume().LessThan(v) {
+		} else if wells[i].CurrentWorkingVolume().LessThan(v.MinusEpsilon()) {
+			// IsZero returns true for very, very small volumes
```
msackman-synthace: comment is bitrotted. kill.

PR #1004, opened 2019-03-22
add time estimates and cumulative time estimates for each instruction…

chr1sgrant:
… in the actions.json file.

Started to poke around here whilst trying to consume this in elements and saw a quick win; let me know if you think this is ok. 

1. One question would be the redundancy of the DurationInSeconds field of the promptAction (which is not used currently)

2. Another Question is whether we also need to update cumulativeTime in the default case of SumarizeActions 

```go 

default:
			// output anything else to the simulator to keep the state up to date

                          cumulativeTime += timer.TimeFor(act.Instruction())
			if err := vlh.Simulate(act.Leaves()); err != nil {
				return nil, err
			}
```


haydnKing:
This is great! I actually had this on my to do list: https://synthace.atlassian.net/browse/ANTHA-2656

The one thing this is missing is an update to the JSON-schema document, see `microArch/scheduler/liquidhandling/schemas`, which is used to validate the produced JSON and is compiled into typescript by the frontend. If you try generating an actions summary on the command line in your branch you get:
```
antha run --driver go://github.com/Synthace/instruction-plugins/PipetMax --bundle ELISA.json --mixSummary actions.json
...
generated an invalid action summary: rule "" broken: failed to set property "/actions/1": did not match any of the specified OneOf schemas
rule "" broken: failed to set property "/actions/2": did not match any of the specified OneOf schemas
```
This isn't picked up because the test liquidhandler doesn't have a timer so no timing info gets added and validation is successful, so I'd recommend:
1) get the timer working for the test liquidhandler, this test should then fail: https://github.com/antha-lang/antha/blob/master/microArch/scheduler/liquidhandling/planning_test.go#L72
2) add the new timing properties to the JSON-schema, along with a description and let @mkonicek know about the update

chr1sgrant:
OK great, I'll get those sorted:

Not too much adaptation for when the mixer wait comes in, although the Asp and Dsp wait instructions won't generate a time estimate at present but but should do if this is cherry-picked: https://github.com/antha-lang/antha/pull/988/commits/1229083ff60531f4c281423a9721ea1229ebe43b

I'll also go ahead and add the code above to the default case (although it may not make much difference at present... and I suppose, if it does increase cumulative time based on instructions which aren't included in actions.json then the risk is that one step seems far longer than it actually is)

chr1sgrant:
Ok, this seems to be working now; I didn't need to add a Timer to the test liquid handler as it already has one.

microArch/scheduler/liquidhandling/summary.go
```diff
@@ -104,27 +109,44 @@ func SummarizeActions(initialState *driver.LHProperties, itree *driver.ITree) ([
 
 	actions := make(actionsSummary, 0, len(acts))
 	for _, act := range acts {
-
+		var timeForInstruction time.Duration
+		if timer != nil {
+			lowLevelInstructions := act.Leaves()
+			for _, leave := range lowLevelInstructions {
```
haydnKing: leaf is the singular ;-)

microArch/scheduler/liquidhandling/summary.go
```diff
@@ -104,27 +109,44 @@ func SummarizeActions(initialState *driver.LHProperties, itree *driver.ITree) ([
...
+			cumulativeTime += timeForInstruction
+		}
 		switch act.Instruction().Type() {
 		case driver.MSG:
 			// record messages as a prompt action
 			if action, err := newPromptAction(vlh, act); err != nil {
 				return nil, err
 			} else {
+				if action.DurationSeconds > 0 {
+					nano := 1e9
```
haydnKing: Is there a good reason for having this in nanoseconds? seems to imply a _lot_ more precision that we have to me...

microArch/scheduler/liquidhandling/summary.go
```diff
@@ -104,27 +109,44 @@ func SummarizeActions(initialState *driver.LHProperties, itree *driver.ITree) ([
...
+					action.TimeEstimate = time.Duration(action.DurationSeconds * nano)
+					action.CumulativeTimeEstimate = cumulativeTime
+				}
 				actions = append(actions, action)
 			}
 		case driver.TFR:
 			// record transfer instructions as transfer actions
 			if action, err := newTransferAction(vlh, act); err != nil {
```
haydnKing: So I think if you pass `timer` and `cumulativeTime` down here you can do something similar to generate timestamps for the lower level instructions

microArch/scheduler/liquidhandling/summary.go
chr1sgrant: haha true, whoops

microArch/scheduler/liquidhandling/summary.go
chr1sgrant: this is now implemented in the latest commit

microArch/scheduler/liquidhandling/summary.go
chr1sgrant: it's just a consequence of how the time.Duration is stored in go; I'll update

PR #1005, opened 2019-03-25
Improve "insufficient source volume" errors

haydnKing:
"Insufficient source volume" errors are notorious for being difficult to debug, and often consume a great deal of core developers time unnecessarily. This is because the error message produced contains little information and is difficult to interpret.
This change improves the quality of the error messages and simplifies the "GetSourcesFor" mechanism.

With this change:
```
error planning: Insufficient source volumes for components 2a9e802f-28f8-4590-a37f-c4f55cfae198:0.721 X Primary Solution+5.12e-05 v/v Standard+0.279 X Wash Buffer;
```
becomes:
```
error planning: found insufficient sources for 1 requested liquids
  eacc0024-a9ca-4c77-8239-1d15eb545ab1 ["0.721 X Primary Solution+5.12e-05 v/v Standard+0.279 X Wash Buffer"]:
    available: 76.5 ul
    requested: 79 ul
    shortfall: 2.5 ul
```

### GetSourcesFor
This function takes a request for a liquid and figures out where those liquids can come from given the current robot state.
It does this by first finding all the liquids which could satisfy the request, then greedily matching sources to requests, checking at each stage that enough source volume is available to service the request previously `sourceVolumesOK`.
Previously this was inconsistent: Liquids were selected based on `Liquid.NameOrID` (which returns the ID if the liquid is an instance, i.e. the result of a mix or explicitly supplied and thus not virtual, or the name otherwise), but then `sourceVolumesOK` checked using `Liquid.FullyQualifiedName`, which returns ID _and_ name - and was thus more restrictive.
In addition, when there wasn't enough source to match, there was an entire mechanism to try again setting `ignoreInstances`, but this would never identify any new sources and was thus guaranteed to fail, and thus has been removed.

### Testing
This has been tested successfully using `antha-tester`


haydnKing:
Had some ideas about how to improve this further, so marking as [WIP]

haydnKing:
Updated the error message to provide a hint in the case that the ID has been updated:
```
error planning: found insufficient sources for 2 requested liquids
  3538df6f-2c29-4891-881d-9c49ccedc32b ["IL4"]:
    available: nothing, but did find potential source(s) with matching name:
        "323a2eb0-5b58-4f50-aa91-4513c6c46cb7:IL4"
    requested: [1 ul 1 ul 1 ul 1 ul 1 ul] 5 ul total
    shortfall: 5 ul
  d370e62f-c908-403c-8d14-2028a833786b ["CD68"]:
    available: nothing, but did find potential source(s) with matching name:
        "829ed1e6-9c74-42ac-941a-ca6ae4c40263:CD68"
    requested: [1 ul 1 ul 1 ul] 3 ul total
    shortfall: 3 ul
```

antha/anthalib/wtype/plate.go
```diff
@@ -1258,47 +1246,25 @@ func (p *Plate) DeclareSpecial() {
...
+	ret := make(ComponentVector, 0, len(wv))
 
-		// ignoreInstances can only work for initial inputs
-		if ignoreInstances && cv[i].Generation() == 0 {
-			identifier = cv[i].CName
-		}
-
-		if cv[i] != nil && wants[identifier] {
-			fcv[i] = cv[i]
+	for _, wc := range wv {
```
msackman-synthace: I would use the index here, remove the 0 from make, and then switch the appends to direct indices. That way it's 100% obvious the input indices and output indices are the same.

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -97,34 +53,38 @@ func getPlateIterator(lhp *wtype.Plate, ori wtype.ChannelOrientation, multi int)
...
+		if p, ok := lhp.Plates[ipref]; ok {
+
 			it := getPlateIterator(p, ori, multi)
 			for wv := it.Curr(); it.Valid(); wv = it.Next() {
-				// cmps needs duping here
-				mycmps := p.GetVolumeFilteredContentVector(wv, cmps, minPossibleVolume, ignoreInstances) // dups components
-				if mycmps.Empty() {
-					continue
+
+				found := p.AvailableContents(wv)
```
msackman-synthace: sorry, in my head `found` is always a boolean. `available` maybe?

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -97,34 +53,38 @@ func getPlateIterator(lhp *wtype.Plate, ori wtype.ChannelOrientation, multi int)
...
+				for _, filter := range filters {
+					found = found.Filter(filter)
```
msackman-synthace: I would be tempted to make filter into `(filters ...func(*Liquid) bool)`

antha/anthalib/wtype/componentvector.go
```diff
@@ -14,14 +14,24 @@ func (cv ComponentVector) String() string {
...
+// does not guarantee to call every filter on every liquid
+func (cv ComponentVector) Filter(filters ...func(*Liquid) bool) ComponentVector {
+	ret := make(ComponentVector, 0, len(cv))
+
+	for _, l := range cv {
+		passed := true
+		for _, filter := range filters {
+			passed = passed && filter(l)
+			if !passed { // don't bother checking the rest once one fails
+				break
```
msackman-synthace: ahh, I would use a break with a label here

```go
OUTER:
 for _, l := range cv {
  for _, f := range filters {
   if !f(l) {
    break OUTER
   }
  }
  ret = append(ret, l)
 }
```

antha/anthalib/wtype/plate.go
msackman-synthace: much nicer now :)

antha/anthalib/wtype/componentvector.go
```diff
@@ -14,14 +14,23 @@ func (cv ComponentVector) String() string {
...
+func (cv ComponentVector) Filter(filters ...func(*Liquid) bool) ComponentVector {
+	ret := make(ComponentVector, 0, len(cv))
+
+Outer:
+	for _, l := range cv {
+		for _, filter := range filters {
+			if !filter(l) {
+				// don't bother checking the rest once one fails
+				// continue the outer for loop
+				continue Outer
```
msackman-synthace: yeah, the most confusing aspect of using labels imo: if the label is on the other line then it's a break, not a continue... I always have to double check these myself.

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -252,35 +241,46 @@ func cmpsEqual(c1, c2 *wtype.Liquid) bool {
...
+
+	// build the filters that we'll use later
+	cmpNames := make(map[string]bool, len(cmps))
+	for _, l := range cmps {
+		if l != nil {
+			cmpNames[l.IDOrName()] = true
+		}
+	}
+	nameFilter := func(l *wtype.Liquid) bool {
+		if l == nil {
```
msackman-synthace: up to you but you could collapse this to a 1 liner: `return l == nil || cmpNames[l.IDOrName()]`

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -252,35 +241,46 @@ func cmpsEqual(c1, c2 *wtype.Liquid) bool {
...
+			return false
+		}
+		return cmpNames[l.IDOrName()]
+	}
+
+	mpvMinusE := lhp.MinPossibleVolume().MinusEpsilon()
```
msackman-synthace: I only just noticed that you didn't use a greek ε !

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -252,35 +241,46 @@ func cmpsEqual(c1, c2 *wtype.Liquid) bool {
...
+	volumeFilter := func(l *wtype.Liquid) bool {
+		if l == nil {
+			return false
+		}
+		// allow volumes which are trivially less than minPossibleVolume
+		return !l.Volume().LessThan(mpvMinusE)
```
msackman-synthace: same - this could be collapsed to 1 liner. Personally, I think it's reasonable for developers to know that boolean operators do shortcutting and value the brevity over the slightly more complex expression. But I can see others would disagree with me.

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -290,44 +290,43 @@ func (lhp *LHProperties) GetComponents(opt GetComponentsOptions) (GetComponentsR
 		}
 
 		bestMatch := wtype.Match{Sc: -1.0}
-		var bestSrc wtype.ComponentVector
+		bestSrc := -1
```
msackman-synthace: I'd prefer the name of this var to indicate it's a slice index. `bestSrcIdx` for example.

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -290,44 +290,43 @@ func (lhp *LHProperties) GetComponents(opt GetComponentsOptions) (GetComponentsR
 		}
 
 		bestMatch := wtype.Match{Sc: -1.0}
```
msackman-synthace: Hmm. I've not read through the scoring stuff and am slightly guessing here. But this appears to be a float which in my head means the default value should be the lowest float possible (`math.Inf(-1)`), not -1.0

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -290,44 +290,43 @@ func (lhp *LHProperties) GetComponents(opt GetComponentsOptions) (GetComponentsR
 		}
```
msackman-synthace: right above here is a comment and a break. But I don't see where the comment ... is resolved. The break could be replaced by a return from my reading - there's no special sauce after the loop.

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -290,44 +290,43 @@ func (lhp *LHProperties) GetComponents(opt GetComponentsOptions) (GetComponentsR
...
+		for i, src := range srcs {
+			if !src.IsEmpty() {
 
-			match, err := wtype.MatchComponents(currCmps, src, opt.Independent, false)
+				match, err := wtype.MatchComponents(currCmps, src, independent, false)
 
-			if err != nil && err.Error() != wtype.NotFoundError {
-				return rep, err
-			}
+				if err != nil && err.Error() != wtype.NotFoundError {
```
msackman-synthace: IME, the `os.IsNotExist(err) bool` pattern is much nicer:

```go
package wtype
func IsNotFound(err error) bool {
  return err != nil && ...
}
```

msackman-synthace: otoh, feel free to postpone this if applying this change consistently has big consequences for this PR. That said, I can only find about 3 hits in the code base myself.

microArch/driver/liquidhandling/getComponents.go
haydnKing: from what I can see in `antha/anthalib/wtype/matchComponent.go` scores are initialised at zero and only ever added to... tbh I'm a bit loathe to touch this matching without a good reason as I've not worked all the way through it so can't claim to fully get it

microArch/driver/liquidhandling/getComponents.go
haydnKing: would tend to agree, though it always feels a bit javascripty

microArch/driver/liquidhandling/getComponents.go
haydnKing: seemed pretty straightforward so I went ahead

microArch/driver/liquidhandling/getComponents.go
msackman-synthace: Fair enough.

PR #1007, opened 2019-03-26
data tables: short-term Parquet patch

levitskovich:
Allows to read only selected columns from a Parquet file.


levitskovich:
> Seems fine, needs documentation though

https://github.com/antha-lang/antha/pull/1019

PR #1009, opened 2019-03-27
Fix GetComponents issue

haydnKing:
did the wrong thing with boolean logic earlier, and failed to run through antha-tester again after. Am highly spooked by the fact that this wasn't caught by unit testing.

Have run this fix through `antha-tester`, and all ran successfully.


haydnKing:
update: it turns out that previously instead of removing potential sources with the wrong names or volumes below the machine's minimum, they were set to nil.
This makes sense as it allows the following matching logic to take the layout into account when deciding which sources to choose since a less good match which allows multichannel may score more highly than a worse match which doesn't.

haydnKing:
testing: latest commit has been tested successfully with antha-tester

microArch/driver/liquidhandling/getComponents.go
```diff
@@ -152,9 +152,21 @@ func (vbln volumesByLiquidName) Summary(lhp *LHProperties) string {
...
+				if l != nil && nameMap[l.CName] {
+					return l
+				}
+				return nil
 			})
+
+			// filter out sources which are all nils
+			other := make([]wtype.ComponentVector, 0, len(sources))
+			for _, s := range sources {
+				if filtered := s.Filter(func(l *wtype.Liquid) bool { return l != nil }); len(s) > 0 {
```
msackman-synthace: woah this is odd do you really mean `len(s) > 0` in the conditional? surely `len(filtered) > 0`?

haydnKing: I am _not_ on form at the moment

msackman-synthace:
Let's gamble; try merging.

PR #1010, opened 2019-03-28
ANTHA-2394 Support non-contiguously loaded tips

haydnKing:
A feature of the hamilton is that it is able to pick tips up independently, leaving gaps. Since antha doesn't support spanning heads yet, this is potentially useful (though admittedly we may run into issues further down the line with this, which will require more exploration).
Currently this behaviour causes a panic as the code to drop tips assumes tips are loaded contiguously. This small change fixes the panic to allow non-contiguous loading, though physical experimentation will be required to validate it.


antha/anthalib/wtype/lhtipwaste.go
```diff
@@ -138,24 +138,22 @@ func (lht *LHTipwaste) Empty() {
 	lht.Contents = 0
 }
 
+// Dispose return a slice of well coordinates which specify where to eject each loaded tip
+// elements in channels which are nil imply no tip is loaded in that location, and the return
```
msackman-synthace: something seems to have gone wrong with the english here. Maybe just a  .  at the end of the first line?

antha/anthalib/wtype/lhtipwaste.go
```diff
@@ -138,24 +138,22 @@ func (lht *LHTipwaste) Empty() {
...
 func (lht *LHTipwaste) Dispose(channels []*LHChannelParameter) ([]WellCoords, bool) {
-	// this just checks numbers for now
 	n := 0
 
-	for _, c := range channels {
+	ret := make([]WellCoords, len(channels))
+	for i, c := range channels {
 		if c != nil {
 			n += 1
+			ret[i] = WellCoords{X: 0, Y: 0}
```
msackman-synthace: nah dude, does not compute.

ret is []WellCoords, not []*WellCoords, which means every element of the slice already has the zero value. So this line does exactly squat.

antha/anthalib/wtype/lhtipwaste.go
```diff
@@ -138,24 +138,22 @@ func (lht *LHTipwaste) Empty() {
 	lht.Contents = 0
 }
 
+// Dispose return a slice of well coordinates which specify where to eject each loaded tip
+// elements in channels which are nil imply no tip is loaded in that location, and the return
+// value in the corresponding location is undefined
```
matttgregg: I feel dirty for making comments on comments, but... maybe `returns` vs `return` plus, full stop/ capital after the first line 'tip'. Also maybe some indication of the purpose of the second return value.

antha/anthalib/wtype/lhtipwaste.go
haydnKing: inglish is hard :)

antha/anthalib/wtype/lhtipwaste.go
```diff
@@ -138,24 +138,22 @@ func (lht *LHTipwaste) Empty() {
...
 
-	//currently tipwastes only ever have one well
-	wcS := make([]WellCoords, 0, n)
-	wc := WellCoords{0, 0}
-	for i := 0; i < n; i++ {
-		wcS = append(wcS, wc)
-	}
-
-	return wcS, lht.DisposeNum(n)
+	// this just checks if there's space
```
matttgregg: This comment feels like it sells things short - it also seems to reduce the waste capacity if succesful? (i.e. there is a side effect?)

antha/anthalib/wtype/lhtipwaste.go
```diff
@@ -138,24 +138,21 @@ func (lht *LHTipwaste) Empty() {
...
 
+// Dispose attempt to eject the tips from non-nil channels into the tipwaste.
+// Returns a slice of well coordinates which specify where each tip should be ejected (undefined for nil channels),
+// and a bool which is true if the tips were disposed of successfully, false if the tipbox is over capacity
 func (lht *LHTipwaste) Dispose(channels []*LHChannelParameter) ([]WellCoords, bool) {
-	// this just checks numbers for now
 	n := 0
 
-	for _, c := range channels {
+	for i, c := range channels {
```
matttgregg: `i` unused now?

haydnKing: what a palaver

PR #1011, opened 2019-03-28
[ANTHA-2700] Get migrate command to cope with files embedded within the workflow

msackman-synthace:
This works. However it has revealed some oversights with the FileManager, so I shall address those in a different ticket tomorrow.

### DIFF
--- a/workflow/v1_2/workflow.go
+++ b/workflow/v1_2/workflow.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"github.com/antha-lang/antha/laboratory/effects"
 	"github.com/antha-lang/antha/microArch/driver/liquidhandling"
 	"github.com/antha-lang/antha/workflow"
 )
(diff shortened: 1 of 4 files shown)


workflow/v1_2/workflow.go
```diff
@@ -70,20 +71,48 @@ type mixTaskResult struct {
...
+	Bytes struct {
+		Bytes []byte `json:"bytes"`
+	} `json:"bytes"`
+}
+
+func maybeMigrateFileParam(fm *effects.FileManager, param json.RawMessage) (json.RawMessage, error) {
+	bif := &bakedInFile{}
+	if err := json.Unmarshal([]byte(param), bif); err != nil {
+		return param, nil
+	} else if bif.Name == "" && len(bif.Bytes.Bytes) == 0 {
```
matttgregg: If only one of these is empty, does that cause a failure below?

msackman-synthace: If Name is empty then the call to fm.WriteAll will still be ok because Name doesn't actually influence anything on disk.
If the bytes is empty, that's still fine - we're just writing out an empty file. So I think we're ok.

PR #922, opened 2018-11-12
[Antha-2258]/serialization between antha-lang and e.g. PipetMaxDriver

haydnKing:
This PR is a significant rewrite of the serialisation that occurs in order to communicate with instruction plugins.

The goal of the change is to make serialisation more robust to changes in structs which describe robots or physical objects within robots. This is achieved by removing custom serialization code where possible, and where not making use of custom JSON serialization such that it sits within the same package as the struct itself.

The library `drivers/liquidhandling` wraps the creation and invocation of the GRPC server, and handles conversion into protobuf defined messages for non-complex types.

More complex objects (e.g. wtype.LHPlate, liquidhandling.LHProperties), are sent using JSON serialisation - for many objects (plates, tipboxes) this already existed and needed maintaining, so adds no extra burden. Serialisation was added for LHProperties, being careful to make sure that pointers refer to the correct underlying object in the deserialised object.

PR #1014, opened 2019-03-29
Antha 2586 liquid purpose

chr1sgrant:
Provisional PR to evaluate the interface and implementation before fully annotating.

Linked to this doc:

https://paper.dropbox.com/doc/Querying-the-history-of-a-liquid--AaP~a2NtEVzT75g9SmJ8gMwpAg-icdaSqfRo78rRSN85stLR

### DIFF
--- a/antha/anthalib/wtype/liquid.go
+++ b/antha/anthalib/wtype/liquid.go
@@ -102,6 +102,91 @@ func (c *Liquid) HasSubComponent(subComponentName string) bool {
 	return hasSubComponent(c, subComponentName)
 }
 
+const liquidPurposeKey = "LiquidPurpose"
+
+type LiquidPurpose string
+
+const (
+	NegativeControl LiquidPurpose = "NegativeControl"
+	PositiveControl LiquidPurpose = "PositiveControl"
+	Standard        LiquidPurpose = "Standard"
+	Diluent         LiquidPurpose = "Diluent"
+	TestSample      LiquidPurpose = "TestSample"
+)
+
+func (c *Liquid) SetLiquidPurpose(purpose LiquidPurpose) {
+	c.Extra[liquidPurposeKey] = purpose
+}
+
+func (c *Liquid) SetAsNegativeControl() {
+	c.SetLiquidPurpose(NegativeControl)
+}
+
+func (c *Liquid) SetAsPositiveControl() {
+	c.SetLiquidPurpose(PositiveControl)
+}
+
+func (c *Liquid) SetAsStandard() {
+	c.SetLiquidPurpose(Standard)
+}
+
+func (c *Liquid) SetAsDiluent() {
+	c.SetLiquidPurpose(Diluent)
+}
+
+func (c *Liquid) SetAsTestSample() {
+	c.SetLiquidPurpose(TestSample)
+}
+
+func (c *Liquid) GetLiquidPurpose() LiquidPurpose {
+	liquidPurpose, found := c.Extra[liquidPurposeKey]
+	if !found {
+		return ""
+	}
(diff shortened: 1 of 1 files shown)


chr1sgrant:
@haydnKing it looks like this[https://github.com/antha-lang/antha/pull/922/files] means I don't need to update any protobuf file as a consequence to changes to the Liquid struct?

antha/anthalib/wtype/liquid.go
```diff
@@ -697,13 +799,29 @@ func (cmp *Liquid) MixPreserveTvol(cmp2 *Liquid) {
...
+	// Step 1. Check if both components are classified as Diluent
+	cmpIsDiluent := cmp.IsDiluent()
+	cmp2IsDiluent := cmp2.IsDiluent()
+
+	// Step 2. add all tags from cmp2 to cmp
+	cmp.addTagsFromLiquid(cmp2)
+
+	// Step 3. unless both were classified as Diluent we'll remove the Diluent tag
+	// since it's unlikely it's still a diluent after being mixed with a non diluent.
+	if !cmpIsDiluent || !cmp2IsDiluent {
```
m-i-sadowski: why not `if !cmp.IsDiluent() || !cmp2.IsDiluent()` ?

antha/anthalib/wtype/liquid.go
```diff
@@ -697,13 +799,29 @@ func (cmp *Liquid) MixPreserveTvol(cmp2 *Liquid) {
...
+// The product of Mix inherits tags from both liquids,
+// "Diluent" is a special case which is removed as a tag unless both liquids are tagged "Diluent".
 func (cmp *Liquid) Mix(cmp2 *Liquid) {
 	wasEmpty := cmp.IsZero()
 	cmp.Smax = mergeSolubilities(cmp, cmp2)
 	// determine type of final
 	cmp.Type = mergeTypes(cmp, cmp2)
 	// add cmp2 to cmp
 
+	// Manage the merging of tags:
```
m-i-sadowski: I'm a bit dubious about this - I'd be inclined to see tags as inherently highly volatile and destroyed by mixing with something else. It really seems quite risky allowing them to propagate indefinitely like this

antha/anthalib/wtype/liquid.go
```diff
@@ -62,6 +62,7 @@ type Liquid struct {
 	Loc                string // refactor to PlateLocation
 	Destination        string
 	Policy             LHPolicy // Policy is where a custom liquid policy is stored
+	Tags               []LiquidPurpose
```
m-i-sadowski: With all the accessor stuff below you're 99% of the way to this being a private variable. 
I would strongly encourage going the other 1% and dealing with the serialization issue as well then making this private access

antha/anthalib/wtype/liquid.go
chr1sgrant: By this point in the code the tags have already been merged so cmp will contain all tags from ccmp2

antha/anthalib/wtype/liquid.go
chr1sgrant: I'd like it to be a visible part of the data structure of the Liquid in the output though

antha/anthalib/wtype/liquid.go
chr1sgrant: I've experimented with this a bit having the tags volatile and then in elements copying over tags after mixing and it's very laborious and error prone to keep track of, the reverse of .ClearTags() would be easier but I'd be inclined to have best of both by changing the type to have an IsPersistant bool field in order to control which tags get inherited.

PR #1015, opened 2019-04-01
ANTHA-2689: Show correct transfer volumes in actions-summary

haydnKing:
Previously, if antha attempted to aspirate from a well that didn't have enough volume, the physical simulator reduced the aspirated volume to the maximum possible and raised a warning (which the users wouldn't see from the UI). This causes confusion as the new mix graph then sees the wrong volume being transferred.

Aspirating from empty or nearly empty wells can happen due to [ANTHA-2704](https://synthace.atlassian.net/browse/ANTHA-2704), where an accumulation of carry volumes consumes more volume than expected. In most cases, this is unlikely to be an issue in physical execution, since the carry volume is deliberately conservative and a full 0.5ul is unlikely to be spirited away on each transfer.

Fixing the underlying issue is likely to be quite a bit more work, see [ANTHA-2704](https://synthace.atlassian.net/browse/ANTHA-2704) for a full summary. 

This proposed piecemeal step changes the physical simulator so that it:
- no longer reduces transferred volumes so mix-graph works as expected
- raises a (non-fatal) warning if a transfer invades the carry volume
- raises an error (fatal) if a transfer exhausts the total volume in a well
Thus the user is insulated from trivial errors by the residual volume, but in more serious cases they are given an error and forced to work around the issue.

See [ANTHA-2689](https://synthace.atlassian.net/browse/ANTHA-2689) for more.

## Testing

Previously this issue was never fatal, so it was possible for test bundles to exist which actually aspirated wells to dryness.

Fortunately, there is only one test bundle which fails due to these changes, `DemoWorkflows/qPCR/Primer+MastermixCombo_QPCR_DemoWorkflow.json`, which @jnunoleitao added recently. After discussion, we agreed that the bundle _should_ fail and thus is being removed from the test bundles, pending successful resolution of 2704. 
Otherwise, antha-tester ran successfully.


microArch/simulator/liquidhandling/simulator.go
```diff
@@ -781,61 +781,56 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
...
-				describe(), tip.MinVol)
-			//will get an error here, but ignore it since we're already raising a warning
-			tip.AddComponent(c) //nolint
-		} else if err := tip.AddComponent(c); err != nil {
-			self.AddErrorf("%s: unexpected tip error \"%s\"", describe(), err.Error())
+		} else {
+
+			v := wunit.NewVolume(volume[i], "ul")
+			tip := arg.adaptor.GetChannel(i).GetTip()
+			fv := tip.CurrentVolume()
```
msackman-synthace: what does the `f` stand for? `tv` I could understand. What's `fv`?

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -781,61 +781,56 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
...
+			fv.Add(v)
+
+			if fv.GreaterThan(tip.MaxVol) {
```
msackman-synthace: Why no Epsilon?

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -781,61 +781,56 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
...
+				self.AddErrorf("%s: channel %d contains %s, command exceeds maximum volume %s",
+					describe(), i, tip.CurrentVolume(), tip.MaxVol)
+			} else if c, err := wells[i].RemoveVolume(v); err != nil {
+				self.AddErrorf("%s: unexpected well error \"%s\"", describe(), err.Error())
+			} else if fv.LessThan(tip.MinVol.MinusEpsilon()) {
```
msackman-synthace: again, I would have guessed this should be PlusEpsilon - my thinking is we should be taking the most conservative approach, not the most gung-ho!

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -781,61 +781,56 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
...
+				self.AddWarningf("%s: minimum tip volume is %s", describe(), tip.MinVol)
+				//will get an error here, but ignore it since we're already raising a warning
+				tip.AddComponent(c) //nolint
+			} else if err := tip.AddComponent(c); err != nil {
```
msackman-synthace: I don't like the dup of AddComponent. I would experiment along the lines of:

```go
} else {
  addErr := tip.AddComponent(c)
  if fv.LessThan(tip.MinVol.PlusMinusEpsilon()) {
    self.AddWarningf(...)
  } else if addErr != nil {
    ...
  }
}
```

_might_ be easier to read?

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -781,61 +781,56 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
 	}
 
 	//check total volumes taken from each unique well
-	uniqueWells := make(map[string]*wtype.LHWell)
-	uniqueWellVolumes := make(map[string]float64)
-	uniqueWellVolumeIndexes := make(map[string][]int)
+	volumeTakenByWell := make(map[*wtype.LHWell]float64, len(wells))
```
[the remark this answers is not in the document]
haydnKing: I think this should be safe since the map is populated with wells taken directly from the plate, and there's no use of `DupKeepID` or anything like that here - definitely a valid concern if this map was more long-lived though

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -781,61 +781,56 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
...
-			reductionUl := reduction.ConvertToString("ul")
-			for _, i := range uniqueWellVolumeIndexes[id] {
-				volume[i] -= reductionUl
-			}
-			self.AddWarningf("%s: well %s only contains %s working volume, reducing aspirated volume by %v",
-				describe(), well.GetName(), well.CurrentWorkingVolume(), reduction)
+	}
+	for well, v := range volumeTakenByWell {
+		volume := wunit.NewVolume(v, "ul")
+		if volume.GreaterThan(well.CurrentVolume().PlusEpsilon()) {
```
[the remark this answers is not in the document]
haydnKing: yeah, it's just accounting for rounding floating point errors. In this case, we allow the volume we want to take to be up to a small $\epsilon$ larger than the (total|available) volume in the well without causing a fuss.
It's a bit weird to think about, but I think this is better than the check if the difference is greater than zero stuff I was doing before

microArch/simulator/liquidhandling/simulator.go
```diff
@@ -781,61 +781,56 @@ func (self *VirtualLiquidHandler) Aspirate(volume []float64, overstroke []bool,
...
+			//
+			// Once that issue has been resolved, this warning should become fatal
+			self.AddWarningf("%s: taking %s from %s which contains only %s working volume, possible aspiration of residual (see ANTHA-2704)",
+				describe(), volume, well.GetName(), well.CurrentWorkingVolume())
 		}
 	}
 
 	//move liquid
 	no_well := []int{}
 	for _, i := range arg.channels {
```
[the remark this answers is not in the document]
haydnKing: I think these are a sanitised version of the arguments or something? I wrote this a good while ago...

microArch/simulator/liquidhandling/simulator.go
haydnKing: it's probably meant to be final volume, but isn't clear so I'll rename

microArch/simulator/liquidhandling/simulator.go
haydnKing: I think the idea is that if the minimum tip volume is 0.5 that we should be tolerant of a final tip volume of 0.499999999, because for all practical purposes it's the same thing

PR #1016, opened 2019-04-01
datatables: eliminating the hack used for reading nulls from JSON

levitskovich:
...which turned out to be failing sometimes. Unfortunately, at the cost of ~10% loss of performance:

Before:

goos: windows
goarch: amd64
pkg: github.com/antha-lang/antha/antha/anthalib/data
BenchmarkJsonUnmarshal-8   	      50	  38380364 ns/op	16300902 B/op	  314353 allocs/op

After:

goos: windows
goarch: amd64
pkg: github.com/antha-lang/antha/antha/anthalib/data
BenchmarkJsonUnmarshal-8   	      30	  41466706 ns/op	15515073 B/op	  295725 allocs/op


levitskovich:
An interesting fact is that the number of allocations hasn't grown.

twoodwark:
is merged

PR #1020, opened 2019-04-02
ANTHA-2608: Apply/Foreach

levitskovich:
- Opted for calling it `Foreach` (`Apply` might make users think that it somehow modifies the table); Toby, if you don't like this, let's rename back.
- Like `Filter`, `table.Foreach.On(...).Interface` takes optional schema assertions - not sure this'll be ever used, but theoretically speaking might be useful.
- Unlike `Filter`, didn't implement `Foreach` without columns list - like `table.Foreach().By(func(r Row))` (not sure it's needed)


antha/anthalib/data/gen.py
```diff
@@ -728,6 +728,56 @@
...
+// If a null value is encountered, the function fails.
+// If given any SchemaAssertions, they are called in the beginning and may have side effects.
+func (o *ForeachOn) {{ t['Type'] }}(fn func(...{{ t['Raw'] }}), assertions ...SchemaAssertion) error {
+	// schema checks
+	if err := o.checkSchema(type{{ t['Type'] }}, assertions...); err != nil {
+		return errors.Wrapf(err, "can't iterate over %+v", o.t)
+	}
+
+	// iterating over the projected table
+	projected := o.t.Must().Project(o.cols...)
```
twoodwark: doesn't need Must here, we can just wrap the error and return it

antha/anthalib/data/gen.py
```diff
@@ -728,6 +728,56 @@
...
+package data
+// Code generated by gen.py. DO NOT EDIT.
+
+import "github.com/pkg/errors"
+
+{% for t in env['TYPE_SUPPORT'] -%}
+// {{ t['Raw'] }}
+
+// {{ t['Type'] }} invokes a user-supplied function passing the named column values as {{ t['Raw'] }} arguments.
+// If a null value is encountered, the function fails.
```
twoodwark: Maybe say 'an error is returned and execution is aborted at the first null value encountered'

antha/anthalib/data/example_test.go
```diff
@@ -351,6 +351,40 @@ func ExampleTable_Join_inner() {
...
+		}
+		cost += v[0].(float64) * float64(v[1].(int64))
+	})
+	fmt.Printf("Total cost: %f\n", cost)
+	// Output: Total cost: 2565.000000
+}
+
+func ExampleTable_Foreach_specialized() {
+	quantity := int64(0)
+	_ = pirateBooty.Foreach().On("Quantity").Interface(func(v ...interface{}) {
```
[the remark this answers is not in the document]
levitskovich: Of course I did, sorry. Fixed it.

antha/anthalib/data/gen.py
levitskovich: fixed

antha/anthalib/data/gen.py
levitskovich: I just wanted to express something like
```
projected, err := o.t.Project(o.cols...)
assert(err == nil)
```
because `o.checkSchema` should have already checked the schema by this moment, and if `Project` fails, then it's probably some internal library error.

twoodwark: ok

PR #1022, opened 2019-04-03
Validate job IDs

simonwhitaker:
While we still use jobIds in file paths and the like, here's a first attempt at validating values to ensure they contain non-troublesome runes.

## Test plan

1. `go test ./...`
2. Run composer passing an invalid job ID

    ```
    ./composer -outdir ~/Desktop/temp -keep -linkedDrivers \
        sampleRepositories.json sampleWorkflow3.json sampleGilsonOnly.json
    ts=2019-04-04T09:01:29.483821Z fatal="Invalid jobId 'åbc': job IDs must match the pattern ^[0-9a-zA-Z_ -]+$"
    ```


workflow/workflow.go
```diff
@@ -47,6 +53,11 @@ func WorkflowFromReaders(rs ...io.ReadCloser) (*Workflow, error) {
 		}
 	}
 
+	rxValidJobId := regexp.MustCompile(ValidJobId)
+	if !rxValidJobId.MatchString(string(acc.JobId)) {
+		panic(fmt.Sprintf("Invalid jobId '%v': job IDs must match the pattern %v", acc.JobId, ValidJobId))
```
matttgregg: Would it be OK to just bubble up the error here?

workflow/workflow.go
```diff
@@ -3,16 +3,22 @@ package workflow
...
+	"regexp"
 	"strings"
 	"unicode"
 
 	"github.com/antha-lang/antha/antha/anthalib/wtype"
 	git "gopkg.in/src-d/go-git.v4"
 )
 
+const (
+	ValidJobId string = "^[0-9a-zA-Z_-]+$"
```
matttgregg: Is it worth forcing first char to be alpha - seems a small burden on the user, but might save us pain later.
The restriction to remove whitespace does mean I'll need to update some of the test workflows, but they're in flux anyway, and I've no objection to that being enforced!

workflow/workflow.go
msackman-synthace: Yeah, this would be better as not-a-panic and in validate.go in the JobId.Validate() method. Obviously, I realise this is likely only temporary, but still, would be better in the more logical place.

.gitignore
```diff
@@ -9,3 +9,5 @@
 .idea/
 
 .build/
+cmd/composer/composer
+cmd/elements/elements
```
msackman-synthace: I wish there was a way in gitignores to do `cmd/([^/]+)/\1` Any thoughts?

workflow/workflow.go
msackman-synthace: @matttgregg should be temporary - ANTHA-2736 should be in play fairly soon

.gitignore
simonwhitaker: I don't think there is.

workflow/workflow.go
simonwhitaker: > The restriction to remove whitespace does mean I'll need to update some of the test workflows, but they're in flux anyway, and I've no objection to that being enforced!

That was fairly arbitrary, so I've added a space to the regexp.

PR #1023, opened 2019-04-04
Add Workflow JSON schema

simonwhitaker:
# Test plan

Static testing:

```bash
npm i -g ajv-cli
for JSON in cmd/composer/*.json; do
    ajv validate -s workflow/schemas/workflow.schema.json -d $JSON
done
```

Runtime testing:

```
$ go build && rm -rf ~/Desktop/temp/ && ./composer -outdir ~/Desktop/temp \
    -keep -linkedDrivers ~/src/clients/synthace/sampleRepositories.json \
    sampleWorkflow3.json sampleGilsonOnly.json
```


.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+    "json.schemas": [
```
simonwhitaker: I have conflicted feelings about checking in anything in `.vscode`. Happy to remove this and document it instead. Let me know what you think.

msackman-synthace: On the basis I use emacs, I'm perfectly happy for this to live here. :)

workflow/schemas/workflow.schema.json
```diff
@@ -0,0 +1,190 @@
...
+                    "items": {
+                        "$ref": "#/definitions/elementType"
+                    }
+                },
+                "Instances": {
+                    "type": "object",
+                    "minProperties": 1,
+                    "propertyNames": {
+                        "$comment": "TODO: confirm this regex. This is the name of an element instance, e.g. 'my_aliquot'",
+                        "pattern": "^[a-zA-Z0-9_-]+$"
```
msackman-synthace: oh they definitely need spaces in here.

workflow/schemas/workflow.schema.json
```diff
@@ -0,0 +1,190 @@
...
+                },
+                "Meta": {
+                    "type": "object",
+                    "properties": {
+                        "x": {
+                            "type": "number"
+                        },
+                        "y": {
+                            "type": "number"
+                        }
```
msackman-synthace: Yeah, it's this sort of thing I'm uneasy about. I wouldn't expect UI folk to change this schema if they need to add something to meta so I'm not sure we should be enforcing anything here... I wonder if there's a way to make these schemas composible in some way?

workflow/workflow.go
```diff
@@ -39,8 +44,29 @@ func WorkflowFromReaders(rs ...io.ReadCloser) (*Workflow, error) {
 	for _, r := range rs {
 		defer r.Close()
 		wf := &Workflow{} // we're never merging _into_ wf so it's safe to have nil maps here
-		dec := json.NewDecoder(r)
-		if err := dec.Decode(wf); err != nil {
+
+		schema, err := Asset("workflow.schema.json")
```
msackman-synthace: Oh for this I would use `MustAsset` - for a const file name it really does seem appropriate to panic if the thing doesn't exist.

workflow/workflow.go
```diff
@@ -39,8 +44,29 @@ func WorkflowFromReaders(rs ...io.ReadCloser) (*Workflow, error) {
...
+		}
+
+		workflowJSON, err := ioutil.ReadAll(r)
+		if err != nil {
+			return nil, err
+		}
+
+		if errors, _ := rs.ValidateBytes(workflowJSON); len(errors) > 0 {
+			fmt.Println(rs.Path())
+			fmt.Printf("%s\n", workflowJSON)
```
msackman-synthace: Err, I assume these are debug printfs which can now be removed?

workflow/workflow.go
```diff
@@ -39,8 +44,29 @@ func WorkflowFromReaders(rs ...io.ReadCloser) (*Workflow, error) {
...
+			return nil, errors[0]
```
msackman-synthace: urgh, https://github.com/qri-io/jsonschema/blob/master/schema.go#L185 you need to check both valerror and err. That's not a great api.

Rather than do the unmarshalling twice, I wonder if it wouldn't be better to do our own unmarshal first, and then if that passes, before our Merge, just do `rs.Validate("/",wf, &errs)` as they do

workflow/schemas/workflow.schema.json
msackman-synthace: Ahh yes, I was going to suggest, if you look at cmd/elements/elements_test.go then at the bottom there we walk through every single checked in workflow in the elements repo. You could hack that code to build up a set of every single instance name in use (for the checked in ones) and so infer from that some sensible constraints on instance names.

workflow/workflow.go
```diff
@@ -39,8 +44,29 @@ func WorkflowFromReaders(rs ...io.ReadCloser) (*Workflow, error) {
...
-		if err := dec.Decode(wf); err != nil {
+
+		schema, err := Asset("workflow.schema.json")
+		if err != nil {
+			return nil, err
+		}
+
+		rs := &jsonschema.RootSchema{}
+		if err := json.Unmarshal(schema, rs); err != nil {
+			return nil, err
```
haydnKing: similar to @msackman-synthace above - this might as well be a panic since we're building in the definition as an asset

.vscode/settings.json
simonwhitaker: > I'm fine with it and do use VS code! Seems fairly restricted usefulness however if it only binds to files in `cmd/composer`?

I agree; I'd suggest we can change that as and when we have workflow files elsewhere in the codebase. But it has to match on something; either a path or a glob, and I don't want to apply it to JSON files elsewhere in the codebase that aren't of this schema type. (One alternative would be to adopt a naming convention for workflow files, e.g. they have to be called *.workflow.json, and then we can apply the schema to anything matching that filename pattern.)

workflow/schemas/workflow.schema.json
simonwhitaker: I've left `additionalProperties` defaulting to `true` there, which means other properties are allowed. I also haven't set `required`, so neither `x` nor `y` are required. So if you're editing a JSON file by hand in VS Code, the editor will suggest `x` and `y` as possible keys, but you can enter anything you like, and any value in the `Meta` field will validate. So I think this is OK as it is. But your comment highlights the fact that JSON schemas can be a little unreadable. I'll add explicit `additionalProperties` and `required` values to make this less magical.

workflow/workflow.go
simonwhitaker: > Rather than do the unmarshalling twice, I wonder if it wouldn't be better to do our own unmarshal first, and then if that passes, before our Merge, just do `rs.Validate("/",wf, &errs)` as they do

Annoyingly, this doesn't work, since `rs.Validate` expects `wf` to be of type `map[string]interface{}`, which is what you get when you unmarshal a JSON document into an `interface{}`, as `ValidateBytes` does. We unmarshal the JSON into a struct, which the JSON validator doesn't know how to handle.

msackman-synthace: That is highly annoying, but I can understand why they've written it that way. Ahh well. Probably worth commenting in the code then.

workflow/workflow.go
```diff
@@ -39,8 +43,23 @@ func WorkflowFromReaders(rs ...io.ReadCloser) (*Workflow, error) {
...
+		if err := json.Unmarshal(schema, rs); err != nil {
+			panic(err)
+		}
+
+		workflowJSON, err := ioutil.ReadAll(r)
+		if err != nil {
+			return nil, err
+		}
+
+		if errors, _ := rs.ValidateBytes(workflowJSON); len(errors) > 0 {
```
msackman-synthace: Just to make sure this doesn't get accidentally merged early, I'm still very keen for both error values to be checked here, not just the lhs, and I would ideally like to avoid just `errors[0]` being returned (perhaps going via `utils.ErrorSlice`... maybe - the copying would be annoying though..).
Also, `errors` as a var name is slightly jarring given there's a package in the stdlib of the same name - personally I would go for `errs`.

simonwhitaker:
@msackman-synthace back to you

PR #1024, opened 2019-04-04
[ANTHA-2739] Refine semantics around merging and validating repositories

msackman-synthace:


workflow/merge.go
```diff
@@ -32,18 +32,50 @@ func (a *Workflow) Merge(b *Workflow) error {
...
+			case repoA.Commit == "":
+				repoA.Commit = repoB.Commit
+			default: // both != "" and != each other
+				return fmt.Errorf("Cannot merge: repository with name '%v' redefined illegally (Commit fields not empty and not equal; %s vs %s).",
+					repoName, repoA.Commit, repoB.Commit)
 			}
+
+			// same but for branch
+			switch {
+			case repoA.Branch == repoB.Branch, repoB.Branch == "":
```
matttgregg: OCD wondering about putting these all on the same line, or at least grouping the two empty checks together! (In all cases.)

msackman-synthace: It's a tricky one - I generally believe fewer lines is normally better, but I also accept that this version is less readable than splitting `repoB.Branch == ""` onto a separate `case`. I went backwards and forwards on it a few times, and decided that with comments it's probably ok as is.

workflow/merge.go
```diff
@@ -32,18 +32,50 @@ func (a *Workflow) Merge(b *Workflow) error {
...
+			case repoA.Directory == repoB.Directory, repoB.Directory == "":
+				// all good
+			case repoA.Directory == "":
 				repoA.Directory = repoB.Directory
-			} else if repoB.Directory != "" && repoA.Directory != repoB.Directory {
-				return fmt.Errorf("Cannot merge: repository with name '%v' redefined illegally (Directory fields not empty and not equal).", repoName)
+			default: // both != "" and != each other
+				return fmt.Errorf("Cannot merge: repository with name '%v' redefined illegally (Directory fields not empty and not equal: %s vs %s).",
+					repoName, repoA.Directory, repoB.Directory)
+			}
```
simonwhitaker: This is less complex and (arguably) a little more readable:

```go
if repoA.Directory != "" && repoB.Directory != "" && repoA.Directory != repoB.Directory {
  // error
} else if repoA.Directory == "" {
  repoA.Directory = repoB.Directory
}
```

workflow/merge.go
```diff
@@ -32,18 +32,50 @@ func (a *Workflow) Merge(b *Workflow) error {
...
+			// point.  Consequently, all we can do safely here is basic
+			// string equality and allow empty fields to be set. This is
+			// slightly more restrictive than we might want, for example,
+			// it could be valid to have two different branch names that
+			// happen to resolve to the same commit, but we can't test
+			// for that here so we play it safe and enforce strict
+			// equality.
+
+			switch {
+			case repoA.Commit == repoB.Commit, repoB.Commit == "":
```
simonwhitaker: There's a common pattern here. Can you extract this out into a function? Something like:

```
if result, err := merge(repoA.SomeField, repoB.SomeField); err != nil {
  // handle err
} else {
  repoA.SomeField = result
}
```

PR #1026, opened 2019-04-05
ANTHA-2618: datatables: Update

levitskovich:
`Update` is implemented on the top of `Extend`. Their interfaces are much the same. This required modifying `Extend` a bit:

1. Allowed homogeneous specializations (e.g. `Table.Extend(...).On(...).Int64`) to return nulls (i.e. pairs `(int64, bool)`) - because:
- even previously `Extend` could write nulls implicitly (if one of the input columns values was null) - so why not allow return them explicitly?
- sometimes "partial `Extend`" is needed (extending some rows and dropping all other ones) - and this change allows for implementing this in an easier way for homogeneous types

2. Added `Table.Extend(...).On(...).Interface` - because:
- we probably need some way to Extend/Update columns of non-standard types (e.g. alias types which Chris seems to be using) without accessing the whole row
- to make the interfaces more uniform (as `Filter` has such method)


antha/anthalib/data/extend.go
```diff
@@ -164,8 +204,10 @@ func (e *ExtendOn) inputs(asType reflect.Type) ([]*Series, error) {
 			return nil, errors.Wrapf(err, "extending new column %q", e.extension.newCol)
 		}
 		ser := e.extension.t.series[colIndex]
-		if err := ser.assignableTo(asType); err != nil {
-			return nil, errors.Wrapf(err, "when extending new column %q", e.extension.newCol)
+		if asType != nil {
```
twoodwark: this level of nesting if isn't required

PR #1027, opened 2019-04-05
Antha 2541/compare

matttgregg:
Not overly sure about the structure of this (split between packages, etc.), but this functionally replicates the previous output comparison tests while removing some unused features.

Particularly wondered about use of error slices here - whether preferable to have a shorter error, and log individual failures directly to the Logger as "err".


laboratory/compare.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"math"
+
+	"github.com/antha-lang/antha/laboratory/compare"
+	"github.com/antha-lang/antha/target"
+	"github.com/antha-lang/antha/workflow"
+)
+
+// Compare compares output generated with any supplied test data in the workflow
+func (lb *LaboratoryBuilder) Compare() error {
```
msackman-synthace: Please keep the receiver varname consistent - across a lot of our code base we've not and it rather bugs me - I've `labBuild` throughout laboratory.go. I don't mind which we use, I just want it to be consistent.

laboratory/compare.go
```diff
@@ -0,0 +1,77 @@
...
+	if lb.workflow.Testing == nil {
+		lb.Logger.Log("msg", "No comparison test data supplied.")
+		return nil
+	}
+
+	mixIdx := 0
+
+	for i, instr := range lb.instrs {
+		switch t := instr.(type) {
+		case *target.Mix:
```
msackman-synthace: hmm, seeing as we're only coping with Mix instructions, you could drop the switch and just do:

```go
if t, ok := instr.(*target.Mix); ok {
  ...
}
```

don't really mind...

laboratory/compare.go
```diff
@@ -0,0 +1,77 @@
...
+			lb.Logger.Log("msg", fmt.Sprintf("[%d] Checking mix instruction, %f seconds", i, t.GetTimeEstimate()))
+			lb.compareTimings(t, mixIdx)
```
msackman-synthace: fwiw, I'd lift the call to expectedMix up to here so that mixIdx stays local to this method and the compare* methods are directly comparing the values they're supplied with and doing no indirection.

laboratory/compare.go
```diff
@@ -0,0 +1,77 @@
...
+			lb.compareOutputs(t, mixIdx)
```
msackman-synthace: how come you're throwing away the errors from `compareTimings` and `compareOutputs` ? Add them to a `utils.ErrorSlice` ?

laboratory/compare.go
```diff
@@ -0,0 +1,77 @@
...
+	} else if err = compareToPercent(em.TimeEstimate.Seconds(), m.GetTimeEstimate(), 10); err != nil {
+		return fmt.Errorf("timing check failed, %v", err)
+	}
+
+	lb.Logger.Log("msg", fmt.Sprintf("Passed timing check. Expected %.3gs, found %.3gs.", em.TimeEstimate.Seconds(), m.GetTimeEstimate()))
+	return nil
+}
+
+func compareToPercent(expected float64, actual float64, percent float64) error {
+	if math.Abs(expected-actual) > math.Abs(0.01*percent*expected) {
```
msackman-synthace: I'd pull the 0.01 out to a const - basically just so that it's got a name and can be found more trivially (not that it's exactly hidden as is...)

laboratory/compare/compareplates.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/antha-lang/antha/utils"
+)
+
+type wellInfo struct {
+	PlateName string
+	PlateType string
+	VolumeUl  string
+}
+
+func infoForWell(w *wtype.LHWell, p *wtype.Plate, idGen *id.IDGenerator) (wellInfo, error) {
```
msackman-synthace: fwiw, I always default for structs in returns to the pointer version so that you can return nil in the error case. Just strikes me as slightly more robust / defensive (i.e. `(*wellInfo, error)` in this case)

msackman-synthace: A custom I picked up years ago when doing a lot of Haskell is to order parameters most logically if you take into account your needs from the pov of currying/partial-application. I know in go we don't have these features, but, I would still put the `IDGenerator` first (same for `Plates` below and others here too) because I could well imagine wanting to do eg: `map (Plates idGen) $ zip gots expecteds` (if you'll excuse the abuse of syntax!). I.e. arguments that are less likely to change between successive calls to the same function can more likely be considered as _state_ (for example, they could be baked into the receiving object if there is one) and so should appear more to the lhs.

laboratory/compare/compareplates.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+
+	for pos := range got {
+		if _, ok := eSeen[pos]; !ok {
+			errs = append(errs, fmt.Errorf("found unexpected plate at position '%s' in output", pos))
+		}
+	}
+
+	return errs.Pack()
```
msackman-synthace: I wonder in this case whether to allow better composition this could be one of the few places where it makes sense for the return type to be `utils.ErrorSlice` and return `errs` directly without the `Pack`

laboratory/compare/compareplates.go
```diff
@@ -0,0 +1,98 @@
...
+func plateWellInfo(p *wtype.Plate, idGen *id.IDGenerator) (map[wellInfo]int, error) {
+	pwi := make(map[wellInfo]int)
+	for _, col := range p.Cols {
+		for _, w := range col {
+			if wi, err := infoForWell(w, p, idGen); err != nil {
+				return nil, err
+			} else if v, ok := pwi[wi]; ok {
+				pwi[wi] = v + 1
+			} else {
+				pwi[wi] = 0
```
msackman-synthace: nah, you can rely on the default value here being 0, so can collapse these last 2 branches: just
```go
} else {
  pwi[wi] = pwi[wi] + 1
}
```
will do.

laboratory/compare/compareplates.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+
+	for w, v := range gwi {
+		if _, ok := eSeen[w]; !ok {
+			errs = append(errs, fmt.Errorf("Saw %d of unexpect well %v", v, w))
+		}
+	}
+
+	return errs.Pack()
```
msackman-synthace: same with this one. I'm just thinking of the fact that at the top level it makes more sense to have a flat list of errors rather than lists-of-lists, and so you'd want to be able to do `errs := append(errs, err...)` essentially, if you see what I mean

workflow/merge.go
```diff
@@ -16,8 +16,12 @@ func (a *Workflow) Merge(b *Workflow) error {
 		return nil
 	case a.JobId == "":
 		a.JobId = b.JobId
-	case a.JobId != b.JobId && b.JobId != "":
-		return fmt.Errorf("Cannot merge: different JobIds: %v vs %v", a.JobId, b.JobId)
+	case b.JobId == "":
+		a.JobId = a.JobId
```
msackman-synthace: perfectly legal to have an empty case that does nothing - you don't need a noop line :)

workflow/validate.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	ValidJobId string = "^[a-zA-Z][0-9a-zA-Z_ -]*$"
+	ValidJobId string = "^[0-9a-zA-Z_ -]*$"
```
msackman-synthace: probably don't want this change?

laboratory/compare.go
msackman-synthace: Yeah so I think I would build up a great big errorslice here, and then do a test:

* if len(errs) != 0 then
  * write it to some sensible file
  * return a fmt.Errorf containing a sensible message and a reference to the file

that way we would still get a sensible error msg on stderr and thus the call to labBuild.Fatal and thus the correct process exit code. But we would not go mad trying to format a bunch of errors via stderr

laboratory/compare.go
matttgregg: :D Much nicer.

laboratory/compare.go
matttgregg: That's much neater, collecting the slices to a higher level works well. I've definitely seen Logger fail when I've abused it too.

laboratory/compare/compareplates.go
matttgregg: I did wonder about that! Done - prefer that to a default struct which should be ignored.

matttgregg: Yes! I like this - helps add meaning to the argument lists. (I've seen similar in F# to make the pipe operators play nicely - which is essentially the same thing.)

laboratory/compare.go
```diff
@@ -0,0 +1,109 @@
...
+		if t, ok := instr.(*target.Mix); ok {
+			labBuild.Logger.Log("msg", fmt.Sprintf("[%d] Checking mix instruction.", i))
+
+			if expected, err := expectedMix(labBuild.workflow, mixIdx); err != nil {
+				errs = append(errs, err)
+			} else {
+				if err := labBuild.compareTimings(t, expected); err != nil {
+					errs = append(errs, err)
+				}
+				errs = append(errs, labBuild.compareOutputs(t, expected)...)
```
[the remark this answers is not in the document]
msackman-synthace: Otoh, I see the danger with Pack() that you only get an `error` out and so later on you can't call any interesting methods on it... hmmm. Makes me wonder whether doing a json.Marshal when it comes to serialising would be the best bet.

laboratory/compare.go
```diff
@@ -0,0 +1,109 @@
...
+			return fmt.Errorf("errors in comparison tests, details in %s", ef)
+		}
+	}
+
+	labBuild.Logger.Log("msg", "Comparison test data passed.")
+	return nil
+}
+
+func (labBuild *LaboratoryBuilder) errorsToFile(errs utils.ErrorSlice) (string, error) {
+	filename := filepath.Join(labBuild.outDir, "comparisons.txt")
```
msackman-synthace: `errorsToFile` is a bit too broad because it looks too similar to `SaveErrors`...

Also generally the APIs that have been built take the path as an input rather than returning it, which if we did that here would mean the method wouldn't need to be attached to labBuild at all.

...both of which make me wonder: should there be a `WriteToFile(p string) error` method on ErrorSlice itself, which would internally call `Pack()` and continue iff the result of that is non nil, and then I'm now wondering if a json.Marshal would be the best approach.

And then I'm also wondering if it would be a good idea to always write out comparisons.txt even if errs is empty so that the empty list is recorded and any downstream consumer can just read that file and know a) comparisons happened b) whether there were any errors or not.

laboratory/compare/compareplates.go
```diff
@@ -0,0 +1,94 @@
...
+
+func plateWellInfo(idGen *id.IDGenerator, p *wtype.Plate) (map[wellInfo]int, error) {
+	pwi := make(map[wellInfo]int)
+	for _, col := range p.Cols {
+		for _, w := range col {
+			wi, err := infoForWell(idGen, w)
+			if err != nil {
+				return nil, err
+			}
+			pwi[*wi] = pwi[*wi] + 1
```
msackman-synthace: I quite like this - it makes it super clear to my eyes that we really are relying on structural equality for some reason, not pointer equality.

workflow/merge.go
```diff
@@ -26,9 +28,24 @@ func (a *Workflow) Merge(b *Workflow) error {
 		a.Elements.merge(b.Elements),
 		a.Inventory.merge(b.Inventory),
 		a.Config.merge(b.Config),
+		a.mergeTesting(b),
 	}.Pack()
 }
 
+func (a *Workflow) mergeTesting(b *Workflow) error {
+	switch {
+	case b == nil:
```
msackman-synthace: Oh this surprises me - I would have expected `a.Testing.merge(b.Testing)` Any reason why you've done it this way around?

laboratory/compare.go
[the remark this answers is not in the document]
matttgregg: I'm happy with using `Pack` - do we use more complex error types much in the code?

laboratory/compare.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if len(errs) > 0 {
+		if ef, err := labBuild.errorsToFile(errs); err != nil {
+			return errors.New("errors in test comparisons, and failed to create error file")
+		} else {
+			return fmt.Errorf("errors in comparison tests, details in %s", ef)
+		}
+	}
+
+	labBuild.Logger.Log("msg", "Comparison test data passed.")
```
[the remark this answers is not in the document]
matttgregg: I must have an old `golint` running - it looks like this was dropped from the linter in November. (For readability and scoping reasons.)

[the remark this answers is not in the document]
msackman-synthace: hah, I have very mixed feelings about linters and I don't tend to run them locally myself. I think for things like unchecked errs I would enforce via a linter. But concerns about control flow and readability I think are often best left to human eyes :)

laboratory/compare.go
msackman-synthace: Thinking about it, if we ever expect the front end to present this stuff (and I think we do), then we should be using a machine readable format so we should be standardising on JSON I feel. So I think I'll change the standard SaveErrors() to also save json - just a []string would be enough here, which I suspect the default json.Marshal will do for us anyway.

workflow/merge.go
matttgregg: Ah - mainly because I'd allowed Testing to be optional, so thought that I needed to check existence at the workflow level? (Or will go handle nil pointer receivers OK?)

msackman-synthace: Ahh I see.The problem is if `a` is nil then you can't assign to it anyway.

Hmm, an alternative would be for `Workflow.Testing` to be of type `Testing` rather than `*Testing` and then you could work of `len(MixTaskChecks)` being 0 or not.

workflow/schemas/workflow.schema.json
```diff
@@ -10,7 +10,7 @@
     "properties": {
         "WorkflowId": {
             "type": "string",
-            "pattern": "^[a-zA-Z][a-zA-Z0-9_ -]*$"
+            "pattern": "^[a-zA-Z0-9_ -]*$"
```
matttgregg: Suggesting we make this change to make pre-existing/autogenerated workflows easier to handle.

utils/errors.go
```diff
@@ -42,3 +46,35 @@ func (efs ErrorFuncs) Run() error {
...
+		return es.Write(f)
+	}
+}
+
+// Write writes the contents of an error slice as json.
+func (es ErrorSlice) Write(w io.Writer) error {
+	if js, jsErr := json.MarshalIndent(es, "", "  "); jsErr != nil {
+		return jsErr
+	} else {
+		fmt.Fprintf(w, "%v\n", string(js))
```
msackman-synthace: very minor: Fprintf does have an error return which we might one day be grateful to have checked :)

laboratory/compare.go
```diff
@@ -13,7 +13,8 @@ import (
 
 // Compare compares output generated with any supplied test data in the workflow
 func (labBuild *LaboratoryBuilder) Compare() error {
-	if labBuild.workflow.Testing == nil {
+
+	if labBuild.workflow.Testing.MixTaskChecks == nil || len(labBuild.workflow.Testing.MixTaskChecks) == 0 {
```
msackman-synthace: ahh, in Go, len is perfectly safe to be passed a nil collection (slice or map) - it'll safely return 0 in both cases.

laboratory/compare.go
```diff
@@ -0,0 +1,83 @@
...
+				errs = append(errs, err)
+			} else {
+				errs = append(errs, labBuild.compareTimings(t, expected))
+				errs = append(errs, labBuild.compareOutputs(t, expected)...)
+			}
+			mixIdx++
+		}
+	}
+
+	filename := filepath.Join(labBuild.outDir, "comparisons.txt")
```
msackman-synthace: s/.txt/.json/

workflow/merge.go
```diff
@@ -24,9 +24,31 @@ func (a *Workflow) Merge(b *Workflow) error {
 		a.Elements.merge(b.Elements),
 		a.Inventory.merge(b.Inventory),
 		a.Config.merge(b.Config),
+		a.Testing.merge(&b.Testing),
 	}.Pack()
 }
 
+func (a *Testing) merge(b *Testing) error {
+	if a == nil || b == nil {
```
msackman-synthace: I think this should be `(a *Testing) merge(b Testing)`. Which would mean b itself is never nil. And then in the above call you can drop the `&` which makes it all look the same and uniform.

workflow/merge.go
```diff
@@ -24,9 +24,31 @@ func (a *Workflow) Merge(b *Workflow) error {
...
+	switch {
+	case b.MixTaskChecks == nil || len(b.MixTaskChecks) == 0:
+		return nil
+	case (a.MixTaskChecks == nil || len(a.MixTaskChecks) == 0) && b.MixTaskChecks != nil:
+		a.MixTaskChecks = make([]MixTaskCheck, len(b.MixTaskChecks))
+		for i, v := range b.MixTaskChecks {
+			a.MixTaskChecks[i] = v
+		}
+		return nil
+	default:
```
msackman-synthace: so with `len()` working fine on `nil` I think this is just:
```go
if len(a.MixTaskChecks) == 0 {
  a.MixTaskChecks = b.MixTaskChecks
} else if len(b.MixTaskChecks) != 0 {
  // error sensibly
}
```
or am I missing something?

workflow/schemas/workflow.schema.json
msackman-synthace: totally agree

PR #1028, opened 2019-04-05
ANTHA-2742 : Improve prompt aggregation

haydnKing:
Prompt aggregation was highly expensive for more than a handful of prompts, this algorithm change makes it much faster.

Should have no external effects. Am running antha-tester now to be on the safe side.


haydnKing:
`antha-tester` passed happily as expected

microArch/scheduler/liquidhandling/executionhelpers.go
```diff
@@ -97,33 +97,12 @@ func thisNode(ic *wtype.IChain, n graph.Node, tg graph.Graph, deps map[graph.Nod
...
-func related(i1, i2 *wtype.LHInstruction, topolGraph graph.Graph) bool {
-	if ancestor(graph.Node(i1), graph.Node(i2), topolGraph) || ancestor(graph.Node(i2), graph.Node(i1), topolGraph) {
-		return true
-	}
-
-	return false
-}
-
-func canAggHere(ar []*wtype.LHInstruction, ins *wtype.LHInstruction, topolGraph graph.Graph) bool {
+// unreachable return true ins cannot reach, or cannot be reached from, any of ar
```
msackman-synthace: english has gone wrong here!

microArch/scheduler/liquidhandling/executionhelpers.go
```diff
@@ -97,33 +97,12 @@ func thisNode(ic *wtype.IChain, n graph.Node, tg graph.Graph, deps map[graph.Nod
...
+func unreachable(ar []*wtype.LHInstruction, ins *wtype.LHInstruction, reachability graph.Reachability) bool {
```
msackman-synthace: hmm, very minor, but just as I don't like negatives in flag names, here I think the `un` is better off lost, and then in the callsite, `!reachable(...)`.

haydnKing: yeah fair enough, otherwise at some point someone's bound to write `!unreachable(...)`

PR #1029, opened 2019-04-05
[ANTHA-2736] Introduce separate WorkflowId and SimulationId

msackman-synthace:
Hopefully sensible levels of enforcement and generation.


msackman-synthace:
Depends on https://github.com/Synthace/instruction-plugins/pull/15

workflow/workflow.go
```diff
@@ -15,7 +19,8 @@ import (
 
 type Workflow struct {
 	SchemaVersion SchemaVersion `json:"SchemaVersion"`
-	JobId         JobId         `json:"JobId"`
+	WorkflowId    SimpleId      `json:"WorkflowId,omitempty"`
+	SimulationId  SimpleId      `json:"SimulationId,omitempty"`
```
haydnKing: Some comment here about the distinction between `WorkflowId` and `SimulationId` would be nice

workflow/workflow.go
```diff
@@ -124,21 +155,17 @@ func (m *Meta) MarshalJSON() ([]byte, error) {
 	return json.Marshal(all)
 }
 
-type JobId string
+type SimpleId string
```
haydnKing: my one reservation about this name is that when you see `simId SimpleId` in the code it's unclear if it's just a random `SimpleId` or actually a contraction of `simulationId`... would renaming `SimpleId` to just `Identifier` be better?

workflow/workflow.go
```diff
@@ -90,6 +106,21 @@ type Meta struct {
 	Rest map[string]interface{} `json:"-"`
 }
 
+func (m *Meta) NameAsIdentifier() string {
```
haydnKing: `NameAsGoIdentifier`? or maybe a comment, since someone might expect this to be an Id

workflow/workflow.go
```diff
@@ -90,6 +106,21 @@ type Meta struct {
...
+	res := []rune{}
+	for _, r := range m.Name {
+		switch {
+		// see https://golang.org/ref/spec#identifier
+		// However, we allow the first rune to be a digit
```
haydnKing: why?

laboratory/laboratory.go
```diff
@@ -36,6 +36,9 @@ type Element interface {
 }
 
 type LaboratoryBuilder struct {
+	WorkflowId   workflow.SimpleId
+	SimulationId workflow.SimpleId
```
haydnKing: are these duplicating the ids in `LaboratoryBuilder::workflow.(Workflow|Simulation)Id`?

workflow/workflow.go
msackman-synthace: I agree with your concern. Identifier is too broad for me though. How about BasicId ?

workflow/workflow.go
msackman-synthace: Deal.

laboratory/laboratory.go
msackman-synthace: yes... I'll have a rethink.

workflow/validate_test.go
```diff
@@ -4,29 +4,30 @@ import (
...
-		{JobId("*"), false},
-		{JobId("åbc"), false},
-		{JobId("🤡"), false},
+		{SimpleId(""), true},
+		{SimpleId("abc"), true},
+		{SimpleId("ABC"), true},
+		{SimpleId("a b"), false},
+		{SimpleId("a"), true},
+		{SimpleId("a0"), true},
+		{SimpleId("0a"), false},
```
matttgregg: I think some of these have changed, this one in particular - allowing leading numerals? Which I'm all in favour of despite suggesting the original restriction myself. (Turns out I do have a few existing ids like '1x...' which I think is fairly common automatic naming from the front end at the moment.)

workflow/workflow.go
haydnKing: Hmm, it's not clear to me that `BasicId` really limits the scope any more than `Identifier`... though I'm not sure what the right word is to encapsulate `WorkflowId` and `SimulationId`...
`RuntimeId`, `SystemId`, `ExecutionId`?
Or is it meant to signify that it's just a random string with no extra information encoded in it? In which case maybe `BaseId`, `UniqueId`?

workflow/workflow.go
matttgregg: I *think* the automatic IDs from the front end are all things like `1x Aliquot`... (if I'm putting the right pieces together), so may be important than we allow these?

workflow/workflow.go
msackman-synthace: It makes no guarantees about uniqueness really. Identifier overlaps with the concept of Identifiers from programming languages and seeing as we have the Go parser/tokeniser/ast in this code base which has a different meaning of Identifier, I don't want to collide with that.

Because of the prefix arg, it is used eg in the simulation id to create hierarchical ids - i.e. simId is prefixed by wfId.

workflow/workflow.go
msackman-synthace: Yarp. Also the use site is always prefixed eg `TestWorkflow{{wf.Name}}``. I'll add this gobbet to the comments :)

laboratory/laboratory.go
msackman-synthace: Removed. Ult it comes down to what we decide elements themselves can see. Before now, they couldn't actually see and inspect the workflow itself at runtime, but they might need the WfId and the SimId, hence the copying. But thinking about it, that strikes me as premature bullshittery, so I've just exposed the whole of the Workflow to elements.

workflow/validate_test.go
msackman-synthace: I think you've not spotted the changes to migrater - the old Name now stays as a simple Name and is not an Id at all. So the old "1x ..." shouldn't appear as an Id anymore.

But yes, you're entirely right - I'd failed to rerun the tests. Sorry!

workflow/workflow.go
haydnKing: fair enough then

PR #1031, opened 2019-04-09
[AC-41] Generate element assets for upload to microservices

simonwhitaker:
The plan here is:

- Extend `elements describe` to make the output format configurable; add protobuf as an output format. This will output elements as protobuf-serialised versions of microservice Element structs.
- Add `elements defaults`, which dumps a JSON doc containing defaults in the format expected by the defaults service
- (In a separate PR) Add a thin wrapper script that uses these, along the lines of:

    ```sh
    #!/bin/sh
    elements describe -format=protobuf > elements.pb
    elements defaults > metadata.json
    ```

In the short term we can then simply invoke the existing `update-metadata` command to send these to the microservice layer.

## Test plan

See https://synthace.atlassian.net/browse/AC-41


simonwhitaker:
@msackman-synthace This one's still in draft, but any comments you might have will be welcome.

cmd/elements/describe.go
```diff
@@ -118,39 +120,187 @@ func describe(l *logger.Logger, args []string) error {
...
+				{
+					elem, err := getProtobufElement(antha, et, ewm)
+					if err != nil {
+						return err
+					}
+					pbElements = append(pbElements, elem)
+				}
+			default:
+				{
+					return fmt.Errorf("Unknown format: '%v'", outputFormat)
```
msackman-synthace: I'd prefer to be a bit more eager validating this flag's value.

cmd/elements/defaults.go
```diff
@@ -0,0 +1,88 @@
...
+	"path"
+	"path/filepath"
+	"reflect"
+	"regexp"
+
+	"github.com/antha-lang/antha/logger"
+	"github.com/antha-lang/antha/workflow"
+)
+
+func defaults(l *logger.Logger, args []string) error {
```
msackman-synthace: hmm. Interesting. I hadn't realised that the defaults service was distinct from the elements service.
Nevertheless, I would think the defaults command is surely only useful for machines, not humans, and would always be called at the same time as metadata.
The reason I'm saying all this is I'd written the whole antha/antha/compile/meta.go stuff assuming you'd never need one without the other, and the API there has not allowed you to use that because you don't have to do the transpilation for defaults.

Basically, I'm wondering if it would be sensible to combine defaults with describe so that describe with format=json or format=protobuf always writes out a defaults.json file too. Less boilerplate/code-duplication, but a more complex command maybe? Not sure what's best... thoughts?

simonwhitaker: I went back and forth on this a couple of times. I ended up making the defaults tool separate because I prefer tools that do one thing, and I like tools that write to STDOUT. It's easier to take any tool that writes to STDOUT and redirect its output to a file than it is to take a tool that outputs multiple files and then so something different with the file you're interested in. So my choice would be to leave this as a separate tool for now.

In the long term, I'd love to have elements and their defaults mastered in a single (logical) service in the webserver layer, and refactor these tools to output a single JSON file that contains data about elements and their defaults all in the same place (imagine that!), at which point the defaults tool could die.

msackman-synthace: Yup, ok, makes sense to me, thanks.

cmd/elements/describe.go
simonwhitaker: Done in e7833fc

cmd/elements/defaults.go
```diff
@@ -0,0 +1,88 @@
...
+		bs, err := json.Marshal(defaults)
+		if err != nil {
+			return err
+		}
+		w := bufio.NewWriter(os.Stdout)
+		defer w.Flush()
+		_, err = w.Write(bs)
+		if err != nil {
+			return err
+		}
```
msackman-synthace: hmm, I wonder if this would be neater using the json.Encoder:

```go
err := json.NewEncoder(bufio.NewWriter(os.Stdout)).Encode(defaults)
```

I'm curious as to what led you to adding the Flush()

cmd/elements/describe.go
```diff
@@ -118,39 +145,183 @@ func describe(l *logger.Logger, args []string) error {
...
+						return err
+					}
+					pbElements = append(pbElements, elem)
+				}
+			}
+		}
+
+		w := bufio.NewWriter(os.Stdout)
+		defer w.Flush()
+
```
msackman-synthace: I reckon you could also bring out `var bs []byte` and also `err error` and then avoid duplication of the error checking and w.Write bits. You would just need to add a `default:` which `returns nil`

cmd/elements/defaults.go
simonwhitaker: I can't remember. I guess it'll get called implicitly when the writer goes out of scope? I'll remove it.

cmd/elements/defaults.go
```diff
@@ -0,0 +1,88 @@
...
+		defaults := make(map[string]interface{})
+
+		for _, repo := range wf.Repositories {
+			err := repo.Walk(func(f *workflow.File) error {
+				dir := filepath.Dir(f.Name)
+				if (!workflow.IsAnthaMetadata(f.Name)) || !regex.MatchString(dir) {
+					return nil
+				}
+
+				bs, err := ioutil.ReadFile(path.Join(repo.Directory, f.Name))
```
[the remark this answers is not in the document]
simonwhitaker: Done

cmd/elements/defaults.go
```diff
@@ -0,0 +1,91 @@
...
+				if (!workflow.IsAnthaMetadata(f.Name)) || !regex.MatchString(dir) {
+					return nil
+				}
+
+				r, err := f.Contents()
+				if err != nil {
+					return err
+				}
+
+				bs, err := ioutil.ReadAll(r)
```
msackman-synthace: gah, sorry, I can't let this one through. Using the json.Decoder means you don't need the intermediate `bs` and can save at least 4 loc:

```go
defer r.Close()
var doc map[string]interface{}
if err := json.NewDecoder(r).Decode(&doc); err != nil {
...
```

PR #1032, opened 2019-04-10
ANTHA-2690: Respect user Liquid names in actions summary

haydnKing:
Sometimes after issuing a mix instruction, elements will rename output liquids using `liquid.SetName(...)`. Previously this was not respected in the generated `actions.json`, this change addresses that.

This is done by keeping track of the last reference to each well within a transferblock and updating the name of the liquid as given by the instruction - similarly to how this is already done for the `FinalProperties` object within `liquidhandler.Plan`, but for each transferblock.


microArch/scheduler/liquidhandling/summary.go
```diff
@@ -385,7 +385,7 @@ type liquidSummary struct {
 // newLiquidSummary create a liquid summary
 func newLiquidSummary(l *wtype.Liquid) *liquidSummary {
 	return &liquidSummary{
-		Name:        l.CName,
+		Name:        l.MeaningfulName(),
```
m-i-sadowski: I'm interested to know if this behaviour is always correct. Currently hard to see how we move beyond it but I'm not totally sure if the lifecycle of names is correct here or not, or whether users might sometimes want one and sometimes the other. 

This is a question to be tackled after refactoring Liquid a bit I think.

antha/anthalib/wtype/liquid.go
```diff
@@ -494,6 +494,16 @@ func (lhc *Liquid) SetName(name string) {
 	lhc.CName = trimString(name)
 }
 
+// MeaningfulName returns the name of the liquid if one has been explicitly set (e.g. with SetName)
```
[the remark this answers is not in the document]
haydnKing: Yeah, I agree this isn't very nice. I've been thinking about a replacement for (/sane implementation of) subcomponents, and it might make sense to do it then

microArch/scheduler/liquidhandling/summary.go
haydnKing: Yeah, it's possible this will have to evolve as the uses for actions.json grow.

Currently, name is only needs to be shown to the user if the element explicitly blessed the Liquid with a name, or if there aren't any subcomponents (e.g. it's an autogenerated input)

PR #1034, opened 2019-04-11
Support maps and arrays for embedded data

matttgregg:
Wondering if this is more complex than needs to be, maybe around swallowing errors and separate 'success' flag.


workflow/v1_2/workflow.go
```diff
@@ -79,20 +80,90 @@ type bakedInFile struct {
...
+}
+
+func maybeMigrateFileArray(fm *effects.FileManager, param json.RawMessage) (bool, json.RawMessage, error) {
+	bifArr := make([]bakedInFile, 0)
+
+	if err := json.Unmarshal([]byte(param), &bifArr); err != nil {
+		return false, nil, nil
+	}
+
+	fmt.Printf("This is an array! %v\n", bifArr)
```
msackman-synthace: debug :)

workflow/v1_2/workflow.go
```diff
@@ -79,20 +80,90 @@ type bakedInFile struct {
...
+		return nil, err
+	} else if success {
+		return js, nil
+	} else {
 		return param, nil
+	}
+}
+
+func maybeMigrateFileMap(fm *effects.FileManager, param json.RawMessage) (bool, json.RawMessage, error) {
+	bifMap := make(map[string]bakedInFile, 0)
```
msackman-synthace: ok, I think this should be `map[string]*bakedInFile` so that if in the input we have a null we will keep that all the way through

workflow/v1_2/workflow.go
```diff
@@ -79,20 +80,90 @@ type bakedInFile struct {
...
-		if f, err := fm.WriteAll(bif.Bytes.Bytes, bif.Name); err != nil {
-			return nil, err
-		} else if fbs, err := json.Marshal(f.AsInput()); err != nil {
-			return nil, err
+		return true, msg, nil
+	}
+}
+
+func maybeMigrateFileArray(fm *effects.FileManager, param json.RawMessage) (bool, json.RawMessage, error) {
+	bifArr := make([]bakedInFile, 0)
```
msackman-synthace: similarly, I think this should be `[]*bakedInFile`

workflow/v1_2/workflow.go
```diff
@@ -79,20 +80,90 @@ type bakedInFile struct {
...
+	} else if f, err := moveDataToFile(fm, bif); err != nil {
+		return false, nil, err
+	} else if msg, err := json.Marshal(f); err != nil {
+		return false, nil, err
+	} else {
+		return true, msg, nil
+	}
+}
+
+func moveDataToFile(fm *effects.FileManager, bif bakedInFile) (*wtype.File, error) {
```
msackman-synthace: So I think this should be doing some validation too - and then move validation out of ...FileFlat. I.e. we probably want to check Name and Bytes in all cases - certainly empty Name is always going to be a problem when it comes to running.

matttgregg: Yes - this one hit me. Migrater ended up trying to move out things like gene sequences because they were in arrays.

workflow/v1_2/workflow.go
```diff
@@ -78,21 +78,99 @@ type bakedInFile struct {
...
+
+	if msg, err := json.Marshal(js); err != nil {
+		return false, nil, err
+	} else {
+		return true, msg, nil
+	}
+}
+
+// maybeMigrateFileFlat returns bool indicating whether this could be serialised as a file reference, then either the original json or transformed json if successful.
+func maybeMigrateFileFlat(fm *effects.FileManager, param json.RawMessage) (bool, json.RawMessage, error) {
```
msackman-synthace: Ok, so this comment applies to all three of maybeMigrateFile* functions:

* the bool and the error can be combined. I 100% agree to ignore the error in the case that the initial unmarshal fails.
* but from there on, it should always be the case that err != nil implies not success.
* in all 3 functions, there's a situation where you check f == nil, after checking the err != nil. That *should* be impossible - I would arrange for the contract of moveDatatoFile to be that if err == nil then the file is non nil.
* finally, by reducing the return type to (json.RawMessage, error) I think you can then just `return json.Marshal(f)` as the final step, which would make all of these shorter still.

workflow/v1_2/workflow.go
```diff
@@ -78,21 +78,99 @@ type bakedInFile struct {
...
+
+// maybeMigrateFileArray returns bool indicating whether this could be serialised as an array of file references, then either the original json or transformed json if successful.
+func maybeMigrateFileArray(fm *effects.FileManager, param json.RawMessage) (bool, json.RawMessage, error) {
+	bifArr := make([]*bakedInFile, 0)
+
+	if err := json.Unmarshal([]byte(param), &bifArr); err != nil {
+		return false, nil, nil
+	}
+
+	js := make([]wtype.File, len(bifArr))
```
msackman-synthace: very minor: in both this case of `js` and the one above, I would use `*wtype.File` in the collection. Just avoids the `*f` bit, which always makes me think "huh, is there some special reason why we're copying out a pointer to a value?".

workflow/v1_2/workflow.go
```diff
@@ -78,21 +78,99 @@ type bakedInFile struct {
...
+	} else if f == nil {
+		return false, nil, nil
+	} else if msg, err := json.Marshal(f); err != nil {
+		return false, nil, err
+	} else {
+		return true, msg, nil
+	}
+}
+
+func moveDataToFile(fm *effects.FileManager, bif *bakedInFile) (*wtype.File, error) {
```
msackman-synthace: * I would be very tempted to make this a method on `bakedInFile`
* Thinking about it carefully, I can understand why we need the first branch checking for `nil`, but I'd add a comment please just saying something like "have to be robust against nil/nulls coming in from the old bundle in collections of files"
* Again, the contract of `WriteAll` is that the non-error results are guaranteed to be non-nil if err is nil. This is relatively standard for Go, so there is no reason for the `f == nil` case.

workflow/v1_2/workflow.go
```diff
@@ -78,21 +78,99 @@ type bakedInFile struct {
...
-	bif := &bakedInFile{}
-	if err := json.Unmarshal([]byte(param), bif); err != nil {
-		return param, nil
-	} else if bif.Name == "" && len(bif.Bytes.Bytes) == 0 {
+
+	if success, js, _ := maybeMigrateFileFlat(fm, param); success {
+		return js, nil
+	} else if success, js, _ := maybeMigrateFileArray(fm, param); success {
+		return js, nil
+	} else if success, js, _ := maybeMigrateFileMap(fm, param); success {
```
msackman-synthace: Yeah, ignoring all the errors here does not seem the right thing to do at all as this will mask errors when writing out files. Combining success-bool with error would make this simpler.

workflow/v1_2/workflow.go
```diff
@@ -78,21 +78,66 @@ type bakedInFile struct {
...
+		return nil, err
 	} else {
-		if f, err := fm.WriteAll(bif.Bytes.Bytes, bif.Name); err != nil {
-			return nil, err
-		} else if fbs, err := json.Marshal(f.AsInput()); err != nil {
+		return param, nil
+	}
+}
+
+// maybeMigrateFileArray returns bool indicating whether this could be serialised as an array of file references, then either the original json or transformed json if successful.
```
msackman-synthace: I would say something like:
> returns a nil json.RawMessage unless: no error was encountered during unmarshalling *and* every value within the slice unmarshals as a valid wtype.File. returns a non-nil error only if an error was encountered when writing a file value to disk.
Reason being this api is somewhat non-standard (though I do understand why!)

workflow/v1_2/workflow.go
```diff
@@ -78,21 +78,66 @@ type bakedInFile struct {
...
 }
 
+// maybeMigrateFileParam returns either the original json representations, or a json representation as serialized files, if possible.
 func maybeMigrateFileParam(fm *effects.FileManager, param json.RawMessage) (json.RawMessage, error) {
-	bif := &bakedInFile{}
-	if err := json.Unmarshal([]byte(param), bif); err != nil {
-		return param, nil
-	} else if bif.Name == "" && len(bif.Bytes.Bytes) == 0 {
-		return param, nil
+
```
msackman-synthace: I would also add in here a comment along the lines of "here be dragons: we do no type checking so this code has to cope with being sent arbitrary element parameters and must be very careful about modifications it makes"

PR #1035, opened 2019-04-12
WIP: datatables: enhancing Rows and Observations

levitskovich:
```
anthaObjectLabel := r.MustObservation("AnthaObjectLabel").MustString()
```

(or maybe even rename it into `.MustCol()` for brevity?)

A minor change to make it possible to get values from a `Row` in one line of code. Hopefully this will make the whole row versions of `Filter`, `Extend`, `Update` and `Foreach` a bit more usable.

I used to think that these versions aren't so important since they are slow - however, after conversations with Marion and Xander about datatables basics it became clear that it's great that we have these easier-to-understand versions - for them to begin with.


levitskovich:
Please don't merge it - going to add some changes on Rows and Observations on Monday. Probably now it's the last opportunity to improve all this a bit - until we have much elements code depending on datatables.

twoodwark:
Perhaps a good chance to use a different word than "Observation".
`Cell` or `Value` are possibilities

PR #1037, opened 2019-04-15
ANTHA-2641: `num` - a library of float arithmetic functions of general use which Antha users might need

levitskovich:
For now, `num` contains:
1. Wrappers on the top of Add, Sub, Mul, Div functions from `gonum`. The point is that `gonum` doesn't have:
* functions like `Add(v1 []float64, v2 []float64) []float64` (there only their versions which require an output vector as well, which are not very readable)
* functions like `Add(v1 []float64, scalar float64) []float64`
2. Several functions from `numpy` - e.g. Linspace and Convolve.

Considered adding more element-wise functions (e.g. `Exp(v)` and `Pow(v1, v2)`), but not sure it makes any sense.
Complex formulas like `y = exp(-1 / (1 - x^2)))` look absolutely unreadable in this syntax:
`y = num.Exp(num.DivConst(-1, num.SubFromConst(1, num.Pow(x, 2))))`
So looks like it's easier to do things like this just in a for loop.


antha/anthalib/num/numpy.go
```diff
@@ -0,0 +1,94 @@
...
+	switch mode {
+	case Full:
+		// `full` mode - returning the whole array
+		return result
+	case Same:
+		// `same` mode - returning the array of length `len(a)`
+		toCut := len(v) / 2 // is this correct? at least, this works for the sample data
+		return result[toCut : toCut+len(a)]
+	case Valid:
+		// `same` mode - returning the array of length `len(a)`
```
twoodwark: Can you fix the comment here

antha/anthalib/num/numpy.go
```diff
@@ -0,0 +1,94 @@
...
+	for i := range result {
+		result[i] = start + float64(i)*step
+	}
+	return result
+}
+
+// ConvolutionMode defines convolution output array length.
+type ConvolutionMode int
+
+const (
```
twoodwark: Please document the mode values

antha/anthalib/num/num_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+func assertPanics(t *testing.T, fn func(), msg string) {
+	if !doesPanic(fn) {
+		t.Errorf(msg + ": expected to panic")
+	}
+}
+
+func doesPanic(fn func()) (b bool) {
```
twoodwark: replace this with https://godoc.org/github.com/stretchr/testify/assert#Panics

antha/anthalib/num/numpy.go
```diff
@@ -0,0 +1,94 @@
...
+	Full = iota
+	Same
+	Valid
+)
+
+// Convolve is a (very naive) implementation of convolution.
+// The results are numerically equivalent to `np.convolve(a, v, mode)` - it looks like that `np.convolve` uses precise convolution as well (but not an FFT approximation).
+// TODO: optimize the implementation - the current one has O((M+N)^2) time complexity.
+func Convolve(a, v []float64, mode ConvolutionMode) []float64 {
+	if len(a) == 0 {
```
twoodwark: Comment should probably refer to the algorithm as *"discrete convolution"*

antha/anthalib/num/num_test.go
levitskovich: great, thanks, didn't know about it

antha/anthalib/num/numpy.go
levitskovich: I remember some linter we don't use (golint?) used to detect things like this

levitskovich:
> NB So one obvious thing to do if performance somehow becomes critical in future is to make sure we deploy with any SSE/AVX support added by gonum. I imagine we'd need a lot of benchmark evidence to justify this, because it would require `unsafe` access to sub-arrays.

I'm afraid that if we at some point are really concerned about performance, then we should get rid of all the functions I've recently added to `floats.go` - because they imply additional memory allocations.

Re SSE/AVX support - as far as I understand, using assembler implementations is turned on by default in `gonum`, so it should be working.

antha/anthalib/num/numpy.go
```diff
@@ -38,20 +38,23 @@ func Linspace(start, end float64, num int) []float64 {
...
+	// Full - returns the convolution at each point of overlap, i.e. of length N+M-1.
 	Full = iota
+	// Same - returns the output of length max(M, N).
 	Same
+	// Valid - returns the output of length max(M, N) - min(M, N) + 1.
 	Valid
 )
 
-// Convolve is a (very naive) implementation of convolution.
+// Convolve is a (very naive) implementation of previce discrete convolution.
```
twoodwark: typo here

levitskovich: fixed

PR #1039, opened 2019-04-16
Fix/prompts to multiple args

m-i-sadowski:
This PR changes the signature of MixerPrompt from 

```MixerPrompt(component *Liquid, message string) Liquid```

to 

```MixerPrompt(message string, components ...*Liquid) []Liquid```

with the analogous change for ExecuteMixes to 

```ExecuteMixes(components ...*Liquid) []Liquid```

There are two reasons for this

1) internally, MixerPrompt already works like this. It makes no sense to restrict this to a single liquid

2) this should make it easier on users who want barriers to operate on particular component versions, since everything mentioned in the prompt will always be synced at the appropriate point, giving users more control over staging of mixes

more details in this ticket https://synthace.atlassian.net/browse/ANTHA-2636

So far the testing based on that ticket's bundles seems good. Jordan is doing further testing. However it would be useful to see whether this idea passes muster in parallel. 

There is a corresponding branch in elements land here: 

https://repos.antha.com/antha-ninja/elements-westeros/tree/fix/promptArgs

with MR https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/772

### CODE REVIEW

**Comment by m-i-sadowski** [comment]
PS ignore merge conflicts above, I will fix these iff this doesn't die in flames

File: microArch/scheduler/liquidhandling/executionhelpers.go [hunk 1]
```diff
@@ -161,16 +168,27 @@ func aggregatePromptsWithSameMessage(inss []*wtype.LHInstruction, topolGraph gra
...
+	// what I think is happening is:
+	// we have hashed out promps by message
+	// then we have taken steps to ensure that those at different
+	// levels of the chain are not merged
+	// now we are making (effectively) dummy instructions to
+	// replace them, using PassThrough
+	// thing is that this really isn't necessary once Prompts
+	// are N-N anyway
+	// which they are now.
+	// so I think PassThrough can DIAF
```
haydnKing: My suspicion is that once users can issue prompts on groups of liquids rather than just individually we probably don't need to do prompt aggregation at all since the element is already telling you the grouping... I think it would be well worth putting in a little time to figuring that out at this stage and actually getting rid of this stuff if we possibly can

m-i-sadowski: I think that's dependent on how workflows get written. To me it makes sense to aggregate if you see the same message at the same level of the tree over and over again rather than repeatedly prompting, and as long as workflows can be written that way I think it's necessary to support it. 

But at least the PassThrough stuff can be removed since now all that's needed is the inputs and outputs of the elements

m-i-sadowski:
Companion branch for elements here https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/772

codegen/codegen.go
```diff
@@ -424,11 +424,11 @@ func Compile(ctx context.Context, t *target.Target, roots []ast.Node) ([]ast.Ins
 
 	root, err := makeRoot(roots)
 	if err != nil {
-		return nil, fmt.Errorf("invalid program: %s", err)
+		return nil, fmt.Errorf("1: invalid program: %s", err)
 	}
 	ir, err := build(root)
 	if err != nil {
-		return nil, fmt.Errorf("invalid program: %s", err)
+		return nil, fmt.Errorf("2: invalid program: %s", err)
```
haydnKing: being a bit more descriptive here would be nice

execute/intrinsics.go
```diff
@@ -97,44 +96,55 @@ func Incubate(ctx context.Context, in *wtype.Liquid, opt IncubateOpt) *wtype.Liq
...
-	ComponentIn *wtype.Liquid
-	Message     string
-	WaitTime    wunit.Time
+	Components   []*wtype.Liquid
+	ComponentsIn []*wtype.Liquid
+	Message      string
+	WaitTime     wunit.Time
+}
+
+func newCompsFromComps(ctx context.Context, in []*wtype.Liquid) []*wtype.Liquid {
```
haydnKing: `newCompsFromComps` means nothing to me... how about `changeLiquidIDs`?

codegen/codegen.go
m-i-sadowski: Yes, it would. I'll see if I can figure something out - we can't be very specific at this point, but at least a bit of detail of what is being tried and why would help

the main point here is to distinguish which of two identical messages was being triggered!

execute/intrinsics.go
m-i-sadowski: I think I prefer `updateLiquidIDs`. I've changed newCompFromComp as well

haydnKing:
Looking very close, just wanted to ask about testing... would be nice to have something in `microArch/scheduler/liquidhandling/liquidhandling_test.go` ideally in `TestExecutionPlanning`

m-i-sadowski:
Yes I had an attack of conscience about this earlier. Will do

m-i-sadowski:
BTW I've run this all through antha-tester vs the relevant elements branch and everything is fine. 
I think discussion on that thread has led to the conclusion that the refactoring work required for this in any unmerged branches will be relatively small. Still need to figure out what to do re documentation

(this is covered in the test plan commit comments but just for visibility)

m-i-sadowski:
OK I have added some unit tests which I hope show both that the system works as anticipated and the difference between single and multiple argument prompting.

execute/intrinsics.go
```diff
@@ -49,7 +48,7 @@ type IncubateOpt struct {
 	PreShakeRadius wunit.Length
 }
 
-func newCompFromComp(ctx context.Context, in *wtype.Liquid) *wtype.Liquid {
+func updateLiquidID(ctx context.Context, in *wtype.Liquid) *wtype.Liquid {
```
matttgregg: Mainly for my understanding - `update...` sounds destructive, wheras this is (I think...) creating (a duplicate with) a new id? (Although I prefer it over `newCompFromComp` for sure.)

matttgregg:
Just a few comments, mainly to clarify my understanding. Otherwise no objections.

antha/anthalib/wtype/componentvector.go
```diff
@@ -90,6 +90,9 @@ func (cv ComponentVector) GetVols() []wunit.Volume {
 	for i, c := range cv {
 		if c == nil {
 			ret[i] = wunit.ZeroVolume()
+		} else if c.Vunit == "" {
+			ret[i] = wunit.NewVolume(c.Vol, "ul")
```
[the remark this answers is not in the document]
m-i-sadowski: it's a bit of a tidy-up (too small to be another 'thesis' IMO ;) - debugging messages about issues with the iChain were getting cluttered up with noise about unsupported unit types. 

Definitely a good idea to centralise. I think that really is a separate thesis though. But one we should definitely pursue.

antha/anthalib/wtype/ichain.go
```diff
@@ -114,8 +114,8 @@ func (it *IChain) Print() {
 				fmt.Printf("-- ")
 			} else if it.Values[j].Type == LHIPRM {
 				fmt.Println("PROMPT ", it.Values[j].Message, "-- ")
-				for in, out := range it.Values[j].PassThrough {
-					fmt.Println(in, ":::", out.ID, " --")
+				for i := range it.Values[j].Inputs {
```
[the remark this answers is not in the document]
m-i-sadowski: it's necessary with the removal of PassThrough that Inputs and Outputs match exactly in the sense that 
- both are the same length
- outputs[i] == 'after version' of inputs[i] 
a check for this could be a good idea?

execute/intrinsics.go
m-i-sadowski: yeah, that's correct. If you have a better name I'm willing to consider it.

PR #1042, opened 2019-04-16
datatables: more docs on data import and export

levitskovich:
Summary:
After conversations with Chris, Marion and Xander it turned out that it was unclear for them what full range of options for transforming their data into datatables is.
So decided to add an explicit "Data import" and "Data export" paragraphs to the library overview.

Test plan:
* `godoc -http=":6060"`
* http://localhost:6060/pkg/github.com/antha-lang/antha/antha/anthalib/data/


twoodwark:
https://godoc.org/github.com/antha-lang/antha/antha/anthalib/data#hdr-Data_import

PR #1043, opened 2019-04-17
ANTHA-2761 - support non-contiguous channels

haydnKing:
Occasionally the planner will attempt to multichannel using e.g. channels 0 and 2 on independent platforms like the Hamilton. This is possible, except that several places lower down in the instruction tree and the summary generation assume that channels are contiguous. This change removes this assumtion.

Jira: https://synthace.atlassian.net/browse/ANTHA-2761

It's possible that we might actually prefer to "compress" the channel selection and use the robot's spanning capability (i.e. use channels 0 and 1 instead of 0 and 2), but this would be a different PR and depends on robot testing.

Testing:
 - antha tester ran successfully
 - simulated the bundle attached to the jira ticket and ran it through the hamilton software successfully


microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -463,32 +446,7 @@ func (ins *ChannelTransferInstruction) GetParameter(name InstructionParameter) i
...
-		c += 1
-		suckinstruction.AddTransferParams(ins.Params(i))
-		blowinstruction.AddTransferParams(ins.Params(i))
-	}
-
-	ret = append(ret, suckinstruction)
-	ret = append(ret, blowinstruction)
-
-	return ret, nil
+	return []RobotInstruction{NewSuckInstruction(ins), NewBlowInstruction(ins)}, nil
```
m-i-sadowski: This seems a bit odd to me - every other instruction has a different signature for New. I'm not sure the inconsistency is necessary here, or useful. 

I could at least see NewSuckInstruction taking a TransferParams object instead - what I most disagree with here is the dependency introduced between Suck, Blow and ChannelTransfer, as evidenced by the slightly weird code in the tests below

microArch/driver/liquidhandling/lhpolicy_test.go
```diff
@@ -56,29 +65,22 @@ func TestDNAPolicy(t *testing.T) {
...
-	tp := TransferParams{
-		What:    "peg",
-		Volume:  wunit.NewVolume(190.0, "ul"),
-		Channel: getChannelForTest(),
-	}
-
-	ins1 := NewSuckInstruction()
-	ins1.AddTransferParams(tp)
+	ins1 := getSingleChannelSuck("peg", wunit.NewVolume(190.0, "ul"))
+	fmt.Println(ins1.What)
```
m-i-sadowski: Stray debug here?

microArch/driver/liquidhandling/compositerobotinstruction.go
haydnKing: Yeah, I think on balance I agree that the inconsistency isn't a good idea. 

That said, one thing I'd like to fix is that it's really hard to know what type of instruction generates what other type - at least if that's encoded in the constructors it's a little more discoverable - but certainly if we decide to do something about it we should do it in a bit more of a coordinated way.

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -1340,38 +1298,52 @@ type SuckInstruction struct {
...
-		PltFrom:         []string{},
-		WellFrom:        []string{},
-		Volume:          []wunit.Volume{},
-		FPlateType:      []string{},
-		FVolume:         []wunit.Volume{},
-		Component:       []string{},
+func NewSuckInstruction(cti *ChannelTransferInstruction) *SuckInstruction {
+	// we're assuming here that the same channel and tip is used for each channel
+	var prms *wtype.LHChannelParameter
+	for _, cp := range cti.Prms {
```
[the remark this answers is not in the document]
haydnKing: Yeah good point - I think this might be worthy of an in person chat. 

Are there robots that we currently or may soon target for which loading different tips on the same head at the same time is possible? Since not even hamilton can do this I wonder it's a piece of complexity we can really do without for now

haydnKing:
Ok, so for posterity - there turned out to be a bit more here, namely:
 - an issue with figuring out if the head would fit in the well when the channels used are non-contiguous which affected troughs
 - several points in lower level instructions which assumed that `len(ins.ARRAY)` should equal to ins.Multi

microArch/driver/liquidhandling/compositerobotinstruction.go
m-i-sadowski: I guess the best thing would be an explicit formal grammar. Combined with, maybe, code generation for at least the skeleton of the structure? Maybe that's crazy but at least a grammar would be nice.

haydnKing:
ok, I think this is ready for another look. I've added the error as suggested and tested again with `antha-tester` to be sure.
I think ChooseChannels will probably need looking at again quite soon in the context of independent liquidhandlers, but that should be a different issue.

haydnKing:
Also ran `cd175d8` through antha-tester to double check - no errors

PR #1044, opened 2019-04-17
ANTHA-2690: Store custom component names within the VLH to preserve them

haydnKing:
Previously the lhsummary was extended to respect custom CNames given to liquids by elements, however when liquids were later sampled from (e.g. by split sample) the custom name was lost and replaced with the empty string.

This change fixes that by storing the custom name within the virtual liquid handler so it continues to be used in future sampling.

Testing:
- added a unit test to verify that the names are persisted


chr1sgrant:
Why not add Meaningful name to the liquid directly?

haydnKing:
Liquid already has `CName` and a whole bunch of other properties that are important at different times so I didn't want to muddy the waters further.
The logic of determining whether the `CName` is meaningful will become easier once the subcomponents rewrite is done and there's no automatic naming though.

microArch/scheduler/liquidhandling/summary.go
```diff
@@ -775,6 +775,10 @@ func newTransferAction(vlh *simulator.VirtualLiquidHandler, act *driver.ITree, t
 			pl := output.PlateLocation()
 			if content, ok := lastUpdate[wellLocation{DeckItemID: pl.ID, Row: pl.Coords.Y, Column: pl.Coords.X}]; ok {
 				content.Name = output.MeaningfulName()
+				// update the name withing the virtual liquidhandler so that it persists if the component is aspirated from later
```
m-i-sadowski: typo  : *withing*

PR #1049, opened 2019-04-25
ANTHA-2747 Add parquet metadata support to data tables.

twoodwark:
The use case for this is adding antha-specific JSON document into the footer.  documentation on the expected usage here: https://paper.dropbox.com/doc/Empirical-data-schemas--Ab8UyJdC2_NJJPkjMEPsY4vlAg-RvV9MPtb5vRwFCvuSIGEM

As part of this, we are no longer redundantly seeking in the file (in reader).

### DIFF
--- a/antha/anthalib/data/parquet/filemeta.go
+++ b/antha/anthalib/data/parquet/filemeta.go
@@ -7,7 +7,7 @@ import (
 // FileKeyValueMetadata represents keys in file-level Parquet
 // metadata, as defined in:
 // https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L924 .
-// In the presence of duplicate keys ion the file, behavior is undefined.
+// In the presence of duplicate keys in the file, behavior is undefined.
 type FileKeyValueMetadata map[string]string
 
 // Write gives a WriteOpt that writes all the given keyvalues into file metadata.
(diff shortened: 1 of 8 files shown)


twoodwark:
Note when we land this we will need to update the master branch AMBR element in westeros.

antha/anthalib/data/parquet/filemeta.go
```diff
@@ -0,0 +1,40 @@
+package parquet
+
+import (
+	"github.com/xitongsys/parquet-go/parquet"
+)
+
+// FileKeyValueMetadata represents keys in file-level Parquet
+// metadata, as defined in:
+// https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L924 .
+// In the presence of duplicate keys ion the file, behavior is undefined.
```
levitskovich: in?

antha/anthalib/data/parquet/parquet_test.go
```diff
@@ -58,7 +58,23 @@ func TestParquet(t *testing.T) {
 			return nil, err
 		}
 
-		return TableFromReader(buffer, columns...)
+		return TableFromReader(buffer, Columns(columns...))
+	})
+
+	// bytes: write + read, also setting arbitrary keyvalue metadata
+	parquetTest(t, "Bytes + kv meta", table, []data.ColumnName{}, func(src *data.Table, columns ...data.ColumnName) (*data.Table, error) {
```
levitskovich: Maybe in future - in order to test opts in all 3 modes (file, bytes, io) - it may be worth rewriting these tests into something like
```
func TestParquet(t *testing.T) {
	runSubTests(t, func(t *testing.T, readAndWrite func(*data.Table, []parquet.ReadOpt, []parquet.WriteOpt)) {
		// test pure readAndWrite + test all the opts we have
	}
}
```
(where there are 3 implementations of `readAndWrite` - file, buffer, io) - but not sure it's worth spending time on it now

PR #1051, opened 2019-04-29
[AC-60] Document workflow

simonwhitaker:


workflow/workflow.go
```diff
@@ -226,14 +226,25 @@ func RandomBasicId(prefix BasicId) (BasicId, error) {
...
 type Repositories map[RepositoryName]*Repository
 
+// Repository is a local, checked-out clone of a Git repository.
+//
+// Directory is the absolute path to the repository clone on the local file
+// system
+// Branch (optional) is the name of the branch to use.
+// Commit (optional) is the SHA1 hash of the commit to use.
+//
+// You can provide Branch or Commit, but not both.
```
simonwhitaker: Things that we should document here, but which I can't figure out from reading the code:

1. Are abbreviated commit hashes OK?
2. Is it OK to provide _neither_ Branch _nor_ Commit? What's the behaviour in that instance? (Presumably, we use whatever is currently the HEAD revision?)

msackman-synthace: If neither are provided then we just use the raw files we find in the directory and don't treat the directory as a git repo at all. This is specifically supported for element devs to save them from having to check in every tiny change just in order to run their code.

My read of the git code is that we are only supporting the full sha1 hash - all 20 bytes. We can of course add support for shortened ones if necessary but I guess at least in the short term I'm hoping it's not necessary.

workflow/workflow.go
```diff
@@ -226,14 +226,25 @@ func RandomBasicId(prefix BasicId) (BasicId, error) {
...
 
+// Repositories is a map of Repository values, keyed by RepositoryName. The keys
+// should be unique, and no key should be a prefix of another. For example, if
+// the keys in this map are github.com/foo/bar and github.com/foo/bar/baz, that
+// will trigger a validation error.
 type Repositories map[RepositoryName]*Repository
 
+// Repository is a local, checked-out clone of a Git repository.
+//
+// Directory is the absolute path to the repository clone on the local file
```
msackman-synthace: I actually suspect you could specify this as a "./foo/bar" and it'd probably get interpreted as relative to CWD... but yikes, let's not do that :)

Another thing worth mentioning is that the value of the Directory is passed through filepath.FromSlash before being interpreted, which means on Windows systems, it can still be specified using / rather than \ as separators. That said, thinking about this now, I'm not sure that's really what we want - this is machine local data, so maybe we should expect it to be given in machine-local format? Lemme know if you agree and I'll raise a ticket to change that.

workflow/workflow.go
```diff
@@ -242,6 +253,24 @@ type Repository struct {
 	gitRepo *git.Repository
 }
 
+// ElementInstanceName is the name of an element instance. It cannot start with
+// a '.' character, and it cannot contain the strings '/' or '..'.
+type ElementInstanceName string
+
+// ElementPath is the relative path to an element type from the root of the
+// repository in which it is defined, e.g. `Elements/New-MVL/Aliquot_Liquid`
+type ElementPath string
```
msackman-synthace: Might be worth noting that the separator must be / and not \ (`Name()` will do the wrong thing otherwise due to the use of `path.Base()` and not `filepath.Base()`. ElementType values are meant to be portable.

workflow/workflow.go
```diff
@@ -250,21 +279,42 @@ type Elements struct {
...
 type ElementType struct {
 	RepositoryName RepositoryName `json:"RepositoryName"`
 	ElementPath    ElementPath    `json:"ElementPath"`
 }
 
+// Name returns the nominal name of an ElementType, taken to be the last part of
+// its path.
+//
+// For example, if an ElementType's path is `Elements/New-MVL/Aliquot_Liquid`,
+// Name() returns `Aliquot_Liquid`
```
msackman-synthace: Might be worth noting that all element types inhabit the same namespace. So two element types cannot have the same Name() even if they come from different repositories (see validate.go line 98)

workflow/workflow.go
simonwhitaker: No strong opinion on this, but I suspect if I was a Windows user I'd want to provide paths in native format (not least so I could copy/paste them from PowerShell etc). But it sounds like the fact that we pass it through FromSlash means we could use either forward-slash or backslash separators on Windows? In which case there's not much to do here?

msackman-synthace: Yeah, essentially we're being too robust - we're allowing unix users to specify in windows format, and vice versa. The effect being that if a unix user has a legit \ in their directory name then things will break, and similarly if a windows user has a legit / in their directory name, it will break. Potentially surprising behaviour but probably unlikely. I'll raise a ticket but set it to do not do.

PR #1054, opened 2019-04-30
[AC-62] Fix TODOs in workflow JSON schema

simonwhitaker:


workflow/schemas/workflow.schema.json
```diff
@@ -76,7 +84,7 @@
         "Config": {
             "type": "object",
             "propertyNames": {
-                "$comment": "TODO: check this regex. This is the name of a device class, e.g. 'GilsonPipetMax'. It may be the case that an enum is preferable.",
+                "$comment": "This is the name of a device class, e.g. 'GilsonPipetMax'.",
```
msackman-synthace: Heh, I wonder: this is essentially an explicit enum in the code, reflecting exactly which devices are supported. I wonder if that should be reflected here in order to avoid accidental typos which would otherwise end up being silently ignored.

workflow/schemas/workflow.schema.json
```diff
@@ -189,7 +197,7 @@
                 "Devices": {
                     "type": "object",
                     "propertyNames": {
-                        "$comment": "TODO: check this regex. This is the name of a device instance, e.g. 'Doris'",
+                        "$comment": "This is the name of a device instance, e.g. 'Doris'",
```
msackman-synthace: I wouldn't be at all surprised if we find that at least to start with, we populate this from the devices microservice with the Ids (guids) of the devices rather name. Though I would prefer, long term, to use name.
Also, I would definitely allow spaces in here, and possibly parenthesis too - I would expect biologists to use descriptive names for their machines.

workflow/schemas/workflow.schema.json
simonwhitaker: Ah, got it. Will change.

msackman-synthace:
erm, don't you need to re-`go generate` and check that in before merging? ;)

PR #1059, opened 2019-04-30
ANTHA-2791: generate tiploads summaries for all channels

haydnKing:
This was a classic case of use of ins.Multi instead of len(ins.Position) causing truncation of the array in the case of non-contiguous tip loading.


haydnKing:
(this was tested by running an offending bundle locally and inspecting the generated actions.json)

PR #1060, opened 2019-04-30
ANTHA-2809: implementing Savitzky-Golay filter in our numeric library

levitskovich:
Summary:

The implementation roughly follows the current SciPy one (https://github.com/scipy/scipy/blob/v1.3.0rc1/scipy/signal/_savitzky_golay.py#L227), but with a few "buts":
* Different least square linear equation solvers are used. Both use LAPACK functions for this. However, we use `gels` since it's the only one implemented in `gonum` (https://github.com/gonum/gonum/blob/master/lapack/lapack64/lapack64.go#L120), while Python code uses on of `gelsd`, `gelsy`, `gelss` (https://github.com/scipy/scipy/blob/v1.3.0rc1/scipy/linalg/basic.py#L1069). But, AFAICT, their results are numerically the same.
* Different convolution functions are used. We use `num.Convolve` which is equivalent to `np.convolve`, they use `scipy.ndimage.convolve1d`. Their results are numerically slightly different on the edges, but the same on the most of the interval.

Test plan: added `TestSavGolFilter` which follows the test in Python code comments.


antha/anthalib/num/savgol.go
```diff
@@ -0,0 +1,122 @@
...
+	// retrieving the size of work space needed (this is how LAPACK interfaces are designed:
+	// if we call the function with lwork=-1, it returns the work size needed in work[0])
+	work := make([]float64, 1)
+	gels(work, -1)
+	lwork := int(math.Ceil(work[0]))
+
+	// solving the equation itself
+	result := gels(make([]float64, lwork), lwork)
+	if !result {
+		panic("gels: computation hasn't converge")
```
twoodwark: "didn't converge"

and can we output the parameter values in this case

twoodwark:
as discussed there are ways we could reduce the recomputation of coefficients, but until someone has an inventive to optimize this let's not bother

cosmonautspiff:
Agree w Toby. Looks fine to me for now.

cosmonautspiff:
I think, as we learn more details the nuances of how the filter is used, we could update the function descriptions to explicitly state the shortfalls of the current implementation.

levitskovich:
Btw, maybe we could merge this? (in spite of the fact it is not used in the the elements code now)

PR #1062, opened 2019-05-01
WIP dockerfile changes

twoodwark:


README_Dockerfile.md
```diff
@@ -50,28 +50,23 @@ message like:
      ---> 2ce136c2261c
     Successfully built 2ce136c2261c
 
-That last number is the Id of the image you've just built. If you run
-`docker images` you should see that Id and it should tell you you've
-recently created that image.
+Run `docker images` to confirm.
```
msackman-synthace: Given you're now tagging it you should explain what you're looking for in the output of `docker images`.

cloudbuild.yaml
```diff
@@ -14,7 +14,11 @@ steps:
   - --key=read-github
 
 - name: 'gcr.io/cloud-builders/docker'
-  args: ['build', '--force-rm', '-t', 'eu.gcr.io/antha-images/antha-core:$COMMIT_SHA', '--build-arg', 'COMMIT_SHA=$COMMIT_SHA', '--build-arg', 'NETRC="$(cat .netrc)"', '.']
+  entrypoint: 'bash'
+  args:
+  - '-ceu'
+  - |
+    docker build '--force-rm' -t "eu.gcr.io/antha-images/antha-core:$COMMIT_SHA" --build-arg COMMIT_SHA="$" --build-arg NETRC="$(cat .netrc)" .
```
msackman-synthace: "$" ??!

twoodwark: typo

Dockerfile
```diff
@@ -1,16 +1,21 @@
...
 COPY scripts/. /antha/.
 RUN ./antha-test.sh
-RUN rm $HOME/.netrc
+
+# Final stage: drop the layers with the netrc credentials. we still need the go
+# compiler, etc, so use the same base image.
+FROM eu.gcr.io/antha-images/golang:1.12.4-build 
+WORKDIR /antha
+COPY --from=build /antha/. .
+COPY --from=build /go/. /go/.
```
msackman-synthace: We need ~/.cache keeping. It's where go stores intermediate build results. Losing this cache will substantially hurt.

msackman-synthace: https://groups.google.com/d/msg/golang-dev/qfa3mHN4ZPA/X2UzjNV1BAAJ (top posting)
TLDR: I wonder whether we should explicitly set GOCACHE in the base image?

README_Dockerfile.md
twoodwark: I don't think it helps, removed.

msackman-synthace:
No, absolutely not. This will not work for several reasons.

1. I want the commit sha baked into the binary. Go `runtime/debug.ReadBuildInfo()` will get that data, and antha core uses this (see composer/generate_mod.go:AnthaModule()) but only if the source is built into the GOPATH directory structure.

2. What you have done is ... break antha-core. Composer takes a workflow and generates not only a main.go but also a go.mod, which includes the exact version required of antha-core. This information is retrieved from the above. Because with your changes the antha code is not actually being added to GOPATH, the generated module for any workflow will be unable to find antha-core so it will attempt to download it fresh, for every single workflow.

twoodwark:
ok.. If we want any support for local docker build (without pushing branches) we ought to build from the working copy which is what this does

for #1 we can handle that by restoring an arg for that purpose so externalizing this 

for #2, we can use a replace -> /antha/  for clearer semantics no?

msackman-synthace:
for the cloud images, I very strongly feel that all the source should be within /go and correctly managed and versioned. BuildInfo() allows us to detect that the code has been built properly within /go and is much cleaner than tricks with link flags as we have done in the past.

composer does detect if the binary has been built from source outside of /go (this information is available via BuildInfo) and so it will adjust the generated go.mod files adding in an extra replace to point to the location where composer was built. This is kinda interesting because really what it's getting at is that the binary is highly coupled to its source, which is not surprising, but nevertheless worth bearing in mind.

I would say that the Dockerfile should ADD . iff COMMIT_SHA is not set. I.e. in the cloud we set COMMIT_SHA and everything gets installed properly. For local builds, if you leave that out then you get the local source.

msackman-synthace:
Ok, thinking about this further, I actually don't think it's reasonable to build a docker image of something uncommitted and pushed.

As discussed, the source and binaries are highly coupled - antha is both a library and so should be installed as a library, plus some binaries to help orchestrate use of that library.

Now on a local developer machine, you may have local changes to source all over the place, in different repos, even local changes to 3rd party libraries. With correct use of gohack and friends you could have all this working locally. But, it's then really really hard to get all the sources you depend on into the docker context. Things like `go mod vendor` are appealing, but I've seen odd behaviour with that in a few places, and then we have the problem that always in Go only the top level vendor directory is obeyed, not nested ones. So from the pov of a workflow-generated-main, there is no top level vendor directory... the whole thing just becomes a complete mess. https://tip.golang.org/cmd/go/#hdr-Modules_and_vendoring

Now to be sure, an awful lot of the blame here lies with Go and the design of modules - it's basically far too hard to install code into GOPATH now. I think if there is demand for this use case then the simplest to start with is going to be a Dockerfile.local

msackman-synthace:
Closing this because I believe I've addressed the concerns directly on feature/future_sanity.

PR #1063, opened 2019-05-01
A little de-bit-rotting for the Dragonfly / any HighLevelLiquidHandler

m-i-sadowski:
the earlier visitor change to shrinkVolumes was incorrect for the TransferInstructions, so it was deleting all plates

Can confirm this now works for the Echo and Dragonfly (tested via running a simple bundle in each case) 

Additionally I've added the dragonfly type to the tryer in auto so it is now a Known Driver (shoot me)

### DIFF
--- a/microArch/scheduler/liquidhandling/liquidhandler.go
+++ b/microArch/scheduler/liquidhandling/liquidhandler.go
@@ -343,12 +343,16 @@ func (this *Liquidhandler) shrinkVolumes(rq *LHRequest) error {
 			HandleTransfer: func(ins *liquidhandling.TransferInstruction) {
 				for _, mtf := range ins.Transfers {
 					for _, tf := range mtf.Transfers {
-						if plate, ok := this.Properties.PlateLookup[tf.PltFrom].(*wtype.LHPlate); ok {
-							usedPlates[plate] = true
-							if well := plate.Wellcoords[tf.WellFrom]; well.IsAutoallocated() {
-								useVol(well, tf.Volume, wunit.ZeroVolume())
+						for _, pos := range []string{tf.PltFrom, tf.PltTo} {
+							plateID := this.Properties.PosLookup[pos]
+							if plate, ok := this.Properties.PlateLookup[plateID].(*wtype.LHPlate); ok {
+								usedPlates[plate] = true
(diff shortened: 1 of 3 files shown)


microArch/scheduler/liquidhandling/liquidhandler.go
```diff
@@ -343,12 +343,16 @@ func (this *Liquidhandler) shrinkVolumes(rq *LHRequest) error {
...
 					for _, tf := range mtf.Transfers {
-						if plate, ok := this.Properties.PlateLookup[tf.PltFrom].(*wtype.LHPlate); ok {
-							usedPlates[plate] = true
-							if well := plate.Wellcoords[tf.WellFrom]; well.IsAutoallocated() {
-								useVol(well, tf.Volume, wunit.ZeroVolume())
+						for _, pos := range []string{tf.PltFrom, tf.PltTo} {
+							plateID := this.Properties.PosLookup[pos]
+							if plate, ok := this.Properties.PlateLookup[plateID].(*wtype.LHPlate); ok {
+								usedPlates[plate] = true
+								if well := plate.Wellcoords[tf.WellFrom]; well.IsAutoallocated() {
```
haydnKing: not sure I understand this change... 
IIRC at this point we're tallying up how much volume has been used from each autogenerated well using the helper `useVol`, and also marking which autogenerated plates were actually used with usedPlates so we can remove plates which only contain unused autogenerated liquids later on.
Since destination volumes (PltTo) aren't autogenerated, there shouldn't be any reason to record them here - analogously, there's no handler for `Dispense` instructions since it's not relevant for autogenerated liquids

m-i-sadowski: this is a bit of a band-aid, I'll admit, but for some reason in this case the destination plate was also going!

I'll dig in further and solve the root cause in due course if it isn't too hideous, or explain why I'm not going to and brace myself for the flak. Given we're aiming to deprecate this interface in the near(?!) future it seems like it's probably not worth expending much effort though.

m-i-sadowski: Further investigation (because breaking the tests did bother me) leads me to conclude that the issue is most likely that we're not actually filling wells in destination plates, hence they get deleted. 

This will be required for getting this into the front-end so I'll revert this and fix that when the time comes

PR #1064, opened 2019-05-01
ANTHA-2797: Physically simulating LLD

haydnKing:
Previously there was a bunch of ultimately unsuccessful work to support Liquid Level Following on the gilson, including the physical simulator estimating the location of the liquidlevel based on the gilson quadratic model. This has caused issues with hamilton because the model is pretty bad with full wells being significantly above the top of the well which causes physical simulation to fail.

This change:
1. changes the simulator behaviour to just assume the liquidlevel is halfway up the well since the actual level will be detected by the machine
2. removes all of the now completely dead code surrounding gilson style LLF

see corresponding change for the Pipetmax below


antha/anthalib/wtype/aparam.go
```diff
@@ -129,10 +129,7 @@ func MakePolicyItems() AParamSet {
 		"NO_AIR_DISPENSE":             AParam{Name: "NO_AIR_DISPENSE", Type: typemap[Bool], Desc: "Prevent dispensing anywhere other than under liquid?"},
 		"CAN_MSA":                     AParam{Name: "CAN_MSA", Type: typemap[Bool], Desc: "Permissible to aspirate from multiple sources? -- currently non functional"},
 		"DESCRIPTION":                 AParam{Name: "DESCRIPTION", Type: typemap[String], Desc: "Summary of LHPolicy to present to the user"},
-		"LLFBELOWSURFACE":             AParam{Name: "LLFBELOWSURFACE", Type: typemap[Float64], Desc: "Distance below surface for Liquid Level Following (LLF) when aspirating"},
-		"LLFABOVESURFACE":             AParam{Name: "LLFABOVESURFACE", Type: typemap[Float64], Desc: "Distance below surface for Liquid Level Following (LLF) when dispensing"},
 		PolicyNameField:               AParam{Name: PolicyNameField, Type: typemap[String], Desc: "Name of the Liquid Policy"},
-		"USE_LLF":                     AParam{Name: "USE_LLF", Type: typemap[Bool], Desc: "Use Liquid-level following if plate has a model for liquid height-volume relations and the driver can use it."},
```
m-i-sadowski: Hmmm.... shouldn't we at least model that the possibility exists? 
To put it another way, how will control over LLF work for, e.g., Hamilton?

haydnKing: Currently in hamilton, the liquidlevel is followed if LLD was enabled, based on the assumptions that 
1. if you're aspirating from just below the liquid surface you'll want to follow it down 
2. if you're dispensing to the surface you'll want to follow it up
3. if you're aspirating or dispensing at the top or bottom of the well you don't know anything about the liquid surface so should stay put
I admit this might not cover every use case that could possibly exist, but I figured it would be better to clean the slate and start again rather than try and repurpose this

m-i-sadowski: I guess that's reasonable enough. Personally I don't entirely agree with the delete-things-for-the-sake-of-it approach but I'm aware I'm in the minority there.

However it does occur to me there's a more material point about this and all the other policy item deletions: this is essentially a published interface, before we take this step can we confirm there are no externalities in terms of these items appearing in user assets (i.e. policy excel files)

If they do then I think the system would just die when presented with previously working files, so we also need to ensure that deleted policy items aren't choked on or we manage the transition some other way (which is obviously a separate change)

haydnKing: I just asked @MattRedpath about documentation. 

From what I can see in `wtype/liquidtype/PolicyMaker.go` it seems that unknown policies generate warnings, but it also seems like those warnings get raised as error further up the call stack, so I think you're right and a policy file with e.g. "USE_LLF" would cause an error - though I think this is better than users expecting a certain behaviour that silently doesn't work.

There aren't any instances of "USE_LLF" in elements_westeros and I just ran antha-tester to make sure (this did catch an issue in the PM driver, now fixed)

m-i-sadowski: OK let's just check what happens if we try it and make sure the error message is sane

haydnKing: The error message is: `Invalid PolicyCommand specified in design file: USE_LLF` - seems ok to me?

PR #1065, opened 2019-05-01
[AC-45] Refactor migration tool

simonwhitaker:
AC-45 requires us to be able to migrate SimulateRequest protobuf files to the new workflow format. This preparatory PR refactors the migrate tool to make it easier to support alternative input formats.

### Unit tests

```
go build ./...
go test ./...
```

### Integration tests

```
$ git checkout feature/future_sanity
Switched to branch 'feature/future_sanity'
Your branch is behind 'origin/feature/future_sanity' by 10 commits, and can be fast-forwarded.
  (use "git pull" to update your local branch)
$ cd cmd/migrate/
$ go build .
$ ./migrate -from ~/Desktop/AC-45/qPCR\ precision\ -\ Validation\ Final.json -gilson-device=gillian -outdir ~/Desktop/AC-45/pre ../composer/repositories.json
ts=2019-05-02T11:29:53.026521Z outdir=/Users/simon/Desktop/AC-45/pre
$ git checkout simon/AC-45/sanitise-interface
Switched to branch 'simon/AC-45/sanitise-interface'
Your branch is up to date with 'origin/simon/AC-45/sanitise-interface'.
$ go build .
$ ./migrate -from ~/Desktop/AC-45/qPCR\ precision\ -\ Validation\ Final.json -gilson-device=gillian -outdir ~/Desktop/AC-45/post ../composer/repositories.json
ts=2019-05-02T11:30:13.543243Z outdir=/Users/simon/Desktop/AC-45/post
$ jq -S '.' ~/Desktop/AC-45/post/workflow/workflow.json > ~/Desktop/AC-45/post/workflow/workflow.sorted.json
$ jq -S '.' ~/Desktop/AC-45/pre/workflow/workflow.json > ~/Desktop/AC-45/pre/workflow/workflow.sorted.json
$ diff ~/Desktop/AC-45/pre/workflow/workflow.sorted.json ~/Desktop/AC-45/post/workflow/workflow.sorted.json
```


msackman-synthace:
Just recording my testing notes:

Been looking for nasty bundles in repos.antha.com/antha-ninja/elements-westeros that have Testing steps and also files, and migrating them with old migrater and new.

    for x in $(for t in $(grep -rl RobotInstructions); do grep -l bytes $t; done); do migrate -from $x -gilson-device=gillian -outdir /tmp/OLD/$x /home/matthew/work/Synthace/antha/cmd/composer/repositories.json; done
    # then update to this branch and rebuild the migrater, then:
    for x in $(for t in $(grep -rl RobotInstructions); do grep -l bytes $t; done); do migrate -from $x -gilson-device=gillian -outdir /tmp/NEW/$x /home/matthew/work/Synthace/antha/cmd/composer/repositories.json; done
    # and then to diff:
    for x in $(for t in $(grep -rl RobotInstructions); do grep -l bytes $t; done); do echo $x && diff -u <(jq -S '.' /tmp/OLD/${x}/workflow/workflow.json) <(jq -S '.' /tmp/NEW/${x}/workflow/workflow.json); done |less

Basically looks all good to me. Need to review the code though :)

workflow/workflow.go
```diff
@@ -208,6 +208,9 @@ func (m *Meta) Set(key string, val interface{}) error {
 	if bs, err := json.Marshal(val); err != nil {
 		return err
 	} else {
+		if m.Rest == nil {
+			m.Rest = make(map[string]json.RawMessage)
+		}
```
matttgregg: This was always a bit of a pain (and the other maps) - I think elsewhere the assumption is always that empty maps have been initialised by the creator. (e.g. as EmptyWorkflow - not sure if it's worth following that here too - i.e. creating Meta with Rest set up properly)

workflow/v1_2/provider.go
```diff
@@ -0,0 +1,295 @@
...
+	"io"
+
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"github.com/antha-lang/antha/antha/anthalib/wtype/liquidtype"
+	"github.com/antha-lang/antha/laboratory/effects"
+	"github.com/antha-lang/antha/logger"
+	"github.com/antha-lang/antha/workflow"
+)
+
+type V1_2WorkflowProvider struct {
```
matttgregg: golang naming throws me - but maybe just `Provider` and `NewProvider`? (Or maybe +`Workflow`?) On the basis that v1_2 is in the package name.

workflow/v1_2/provider.go
```diff
@@ -0,0 +1,295 @@
...
+func (p *V1_2WorkflowProvider) GetWorkflowID() (workflow.BasicId, error) {
+	if id, err := workflow.RandomBasicId(""); err != nil {
+		return "", err
+	} else {
+		return id, nil
+	}
+}
+
+func (p *V1_2WorkflowProvider) GetMeta() (workflow.Meta, error) {
+	meta := workflow.Meta{}
```
matttgregg: May prefer to initialise the 'Rest' part here (or follow the rabbit hole of adding a `workflow.EmptyMeta()`)?

workflow/v1_2/provider_test.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, err
+	}
+	fm, err := effects.NewFileManager(tmpDir, tmpDir)
+	if err != nil {
+		return nil, err
+	}
+
+	repo := &workflow.Repository{
+		Directory: "/tmp",
```
matttgregg: If I read right, the value `/tmp` is completely ignored? Maybe worth commenting.

cmd/migrate/main.go
```diff
@@ -1,52 +1,138 @@
...
 )
 
+const (
+	fromFormatJSON     = "json"
+	fromFormatProtobuf = "protobuf"
+)
+
+var validFromFormats = []string{fromFormatJSON, fromFormatProtobuf}
+
 func main() {
```
matttgregg: Not sure if it's worth refactoring this function now? Feels like we're doing several different things. (As per code comments.)

workflow/migrate/migrator.go
```diff
@@ -0,0 +1,73 @@
...
+// NewMigrator creates and returns a migrator
+func NewMigrator(logger *logger.Logger, provider provider.WorkflowProvider) *Migrator {
+	return &Migrator{
+		logger:   logger,
+		provider: provider,
+	}
+}
+
+// Workflow returns the workflow resulting from exercising this migrator
+func (m *Migrator) Workflow() (*workflow.Workflow, error) {
```
matttgregg: Not obvious what benefit the `Migrator` object provides now - but accept may be due to the protobuf work to follow? In which case ignore. 
(Versus, e.g., WorkflowProvider providing a full workflow rather than the parts?)

workflow/migrate/migrator.go
```diff
@@ -0,0 +1,73 @@
...
+// Migrator migrates data from a previous format to the v2.0 format
+type Migrator struct {
+	logger   *logger.Logger
+	provider provider.WorkflowProvider
+}
+
+// NewMigrator creates and returns a migrator
+func NewMigrator(logger *logger.Logger, provider provider.WorkflowProvider) *Migrator {
+	return &Migrator{
+		logger:   logger,
```
matttgregg: logger seems unused now? (But possibly waiting on protobuf work?)

cmd/migrate/main.go
```diff
@@ -1,52 +1,138 @@
...
+		logger.Fatal(l, err)
+	}
+
+	// Merge the generated v2.0 workflow into the stuff we got from the snippets
+	if err = cwf.Merge(wf); err != nil {
+		logger.Fatal(l, err)
+	}
+
+	// validate the resulting workflow
+	if err = cwf.Validate(); err != nil {
```
msackman-synthace: shouldn't this happen iff validate (the flag)?

workflow/migrate/migrator.go
```diff
@@ -0,0 +1,73 @@
...
+		provider: provider,
+	}
+}
+
+// Workflow returns the workflow resulting from exercising this migrator
+func (m *Migrator) Workflow() (*workflow.Workflow, error) {
+
+	wf := &workflow.Workflow{
```
msackman-synthace: `workflow.EmptyWorkflow()` is preferable because you can then make assumptions about embedded maps not being nil and causing panics.
... hmm seeing as you overwrite _everything_ maybe it's not essential. But it's still worth using `workflow.CurrentSchemaVersion` to avoid duplicating a magic number.

workflow/migrate/provider/interface.go
```diff
@@ -0,0 +1,17 @@
...
+// to generate a workflow, e.g. a v1.2 workflow JSON file
+type WorkflowProvider interface {
+	GetWorkflowID() (workflow.BasicId, error)
+	GetMeta() (workflow.Meta, error)
+	GetRepositories() (workflow.Repositories, error)
+	GetElements() (workflow.Elements, error)
+	GetInventory() (workflow.Inventory, error)
+	GetConfig() (workflow.Config, error)
+	GetTesting() (workflow.Testing, error)
+}
```
msackman-synthace: totally cosmetic: I would have put this in the migrate package - I can't see that anything is gained by having this in its own package.

workflow/v1_2/provider.go
```diff
@@ -0,0 +1,295 @@
...
+func (p *V1_2WorkflowProvider) getGilsonPipetMaxConfig() (workflow.GilsonPipetMaxConfig, error) {
+	if p.gilsonDeviceName == "" {
+		return workflow.GilsonPipetMaxConfig{}, nil
+	}
+
+	devices := map[workflow.DeviceInstanceID]*workflow.GilsonPipetMaxInstanceConfig{}
+	devID := workflow.DeviceInstanceID(p.gilsonDeviceName)
+
+	if _, found := devices[devID]; found {
+		p.logger.Log("warning", fmt.Sprintf("Gilson device %s already exists, and will have configuration replaced with migrated configuration.", p.gilsonDeviceName))
```
msackman-synthace: Yeah, so this check is now impossible - this is essentially dead code.
The old code built up on top of the merged snippets and so here we could detect whether we were about to overwrite an existing mixer device.
But now, you're building a workflow from scratch and then merging that on top of the snippets, which would, for example, just error hard if the snippets already defined the same gilson.
Which may well all be fine - there are no particular use cases of merging devices together that I've identified, but certainly this test as it stands can never be hit.

workflow/workflow.go
```diff
@@ -208,6 +208,9 @@ func (m *Meta) Set(key string, val interface{}) error {
 	if bs, err := json.Marshal(val); err != nil {
 		return err
 	} else {
+		if m.Rest == nil {
+			m.Rest = make(map[string]json.RawMessage)
```
msackman-synthace: ahh right, yes, this would be not needed if you use EmptyWorkflow

workflow/migrate/provider/interface.go
```diff
@@ -0,0 +1,17 @@
+package provider
+
+import (
+	"github.com/antha-lang/antha/workflow"
+)
+
+// WorkflowProvider is an entity capable of providing the necessary information
+// to generate a workflow, e.g. a v1.2 workflow JSON file
```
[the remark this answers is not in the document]
simonwhitaker: Yeah sorry, the English here is terrible. Will rework it!

cmd/migrate/main.go
simonwhitaker: Don't understand the question

msackman-synthace: There is a validate command line flag which now appears to be completely unused. I believe it should be used to guard the call to Validate().

cmd/migrate/main.go
```diff
@@ -1,52 +1,138 @@
...
+	dataDir := filepath.Join(outDir, "data")
+	fm, err := effects.NewFileManager(dataDir, dataDir)
+	if err != nil {
 		logger.Fatal(l, err)
 	}
 
-	if err := m.ValidateCur(); err != nil {
-		if validate {
+	// Sanity check: we can only read from STDIN once
+	inputPaths := append(flag.Args(), fromFile)
```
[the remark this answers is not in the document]
simonwhitaker: I don't have strong opinions on this one, but the motivation behind the current implementation was:

a) The logic for sanitising command-line arguments belongs in the command, not in a library function (my opinion)
b) I personally find the slicing of `rs` (in your example) back into its component parts over-complicated and hard to read. The goal is not to have the fewest lines of code, it is to have comprehensible code. I find the refactored code a lot easier to read, but then again maybe I would, because I wrote it.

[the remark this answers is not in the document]
msackman-synthace: I understand where you're coming from. If we had pattern matching then we could do

```
fromReader : rs := ReadersFromPaths(fromFile : rs)
```

which appeals to my inner haskell.

I am totally conflicted on this I guess. On the one hand, I do tend to believe that in general the less code you write, the better. On the other hand, one of the things I like about Go is that the lack of powerful abstraction mechanisms means you have to duplicate yourself a little more which tends to lead to less surprising code IME.

Personally I find the code duplication distasteful, possibly because it's slightly different to the code in readers.go, and possibly because we're already in a function that only just fits on screen. For me, the awkwardness of the slicing is preferable to the duplication, but for sure, this is just personal opinions and our individual sense of aesthetics. If you don't want to change it then don't - I won't hold this PR up because of it.

workflow/v1_2/provider.go
```diff
@@ -0,0 +1,295 @@
...
+	return &V1_2WorkflowProvider{
+		owf:              wf,
+		fm:               fm,
+		repoMap:          repoMap,
+		gilsonDeviceName: gilsonDeviceName,
+		logger:           logger,
+	}, nil
+}
+
+func (p *V1_2WorkflowProvider) GetWorkflowID() (workflow.BasicId, error) {
```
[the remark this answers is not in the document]
matttgregg: Sleeping on this, decided that, in truth, I wouldn't _necessarily_ expect a general `Provider` object to provide the same result on multiple invocations so feel free to ignore the original comment.

workflow/migrate/migrator.go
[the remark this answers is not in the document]
simonwhitaker: Done in 1c34d3ba

workflow/migrate/provider/interface.go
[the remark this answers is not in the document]
simonwhitaker: Reworded in b037124e

workflow/workflow.go
simonwhitaker: Ah nice, thanks - I don't need this now I'm using `EmptyWorkflow`, will remove.

simonwhitaker: Have added an `EmptyMeta()` in ebefdaff and used that in the provider's `GetMeta()` function.

workflow/v1_2/provider.go
simonwhitaker: Done in 4d1aec80

workflow/v1_2/provider.go
simonwhitaker: Yep, EmptyMeta added in ebefdaff

workflow/migrate/migrator.go
simonwhitaker: It's mostly an aid to readability. When I look at that class it's clear what it does any how it does it: it takes a provider (which is in turn a type that provides each of the top-level parameters of a workflow) and uses it to create and return a new workflow. Enforcing that level of granularity feels about right to me: some of the provider getters still get fairly complicated, but they're more modular (and hence readable) than they might be if the provider just had a GetWorkflow() function.

workflow/migrate/migrator.go
simonwhitaker: Hmm... I can't remember why I added it there now. :) Maybe something I subsequently removed. I'll remove it until we need it.

workflow/migrate/migrator.go
simonwhitaker: Now using `EmptyWorkflow()` as of 1c34d3ba

workflow/workflow.go
simonwhitaker: Actually it is, because we call the provider's GetMeta() to get the meta chunk of the workflow, then apply that on top of what EmptyWorkflow gave us. I've added an EmptyMeta function in ebefdaff.

workflow/migrate/provider/interface.go
simonwhitaker: 84c1dfe7

workflow/v1_2/provider.go
simonwhitaker: 2eb352a4

cmd/migrate/main.go
simonwhitaker: 753850ff

workflow/migrate/migrator.go
simonwhitaker: 31dbd78d

cmd/migrate/main.go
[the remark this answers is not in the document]
simonwhitaker: e0236b1e

workflow/migrate/migrator.go
matttgregg: 👍

workflow/v1_2/migrater.go
```diff
@@ -1,334 +0,0 @@
...
-		return nil, nil, err
-	}
-
-	cwf, err := workflow.WorkflowFromReaders(rs[:len(rs)-1]...)
-	if err != nil {
-		return nil, nil, err
-	} else if err := cwf.EnsureWorkflowId(); err != nil {
-		return nil, nil, err
-	}
-	cwf.Meta.Name = filepath.Base(migrate)
```
msackman-synthace: It's a pity to have lost this line: we have a lot of test workflows in the elements repo that have no Name and so having a default of Name = filename was useful. We'll have to add this back in when we next merge in from master and have to re-migrate every single test workflow.

PR #1066, opened 2019-05-07
ANTHA-2678: Liquid.sources

haydnKing:
In the mix graph, we want to show users what a particular liquid is made up of, for example we might want to show something like: 
- "reaction 1" contains
  - v_1 ul "Master Mix"
  - v_2 ul "DNA Template"
  - v_3 ul "dNTPs"
  - v_4 ul "Polymerase"

Previously there was some work to represent liquid sources using the "subcomponents" mechanism, though this has never worked properly as it is poorly documented. Alas, this PR does not fully replace subcomponents, but satisfies the requirements for the mix graph and provides a step towards removing subcomponents.

This new mechanism keeps track of all the "sources" of a liquid - i.e. which named liquids were combined to create this liquid and in what proportion. `Liquid.sources` stores the tree which can be accessed via `Liquid.Sources()` (which scales the LiquidSource volumes correctly). `Liquid.sources` is updated when two or more liquids are mixed, when a liquid is sampled, or when an explicit name is set for a liquid.

Since this touches a fair bit of code around liquid creation, mixing and sampling, there's also a fair amount of tidying up and standardising these areas.

## Testing

Several tests and examples are provided which demonstrate the use of `Liquid.Sources()` API, see `wtype/liquid_test.go`, which hopefully prove useful. In addition there are several tests of the new summary JSON generated in `microarch/summary/liquidhandling/summary.go`.

I'm currently working on two test bundle issues, will update when done.


haydnKing:
Alright, this passes now. Errors fixed were:
1. A legitimate issue with simulating blowouts, since it looks like you're sampling from an empty tip
2. A completely wild and bizarre issue with the old subcomponents. It was crashing at `makeTransfers`, ultimately because the component concentration isn't maintained at that stage since it's normally meant to be used by `SampleForConcentration`, but SubComponents seems to overload it in some cases. After about 2.5hrs of fruitless chasing I gave in and nipped it in the bud by avoiding trying to sum concentrations with "v/v". Since SubComponents is now unused once we're out of elements world I think this is OK.
I want to delete it _so much_

antha/anthalib/wtype/liquid.go
```diff
@@ -1154,3 +1213,153 @@ func (l1 *Liquid) EqualTypeVolumeID(l2 *Liquid) bool {
...
+	srcs := ls.Sources.Dup()
+	srcs.merge(rhs.Sources)
+	return &LiquidSource{
+		Volume:  wunit.AddVolumes(ls.Volume, rhs.Volume),
+		Sources: srcs,
+	}
+}
+
+// LiquidSources all the source volumes which make up a liquid
+type LiquidSources map[string]*LiquidSource
```
m-i-sadowski: oof... pointer here? Shouldn't these be immutable?

antha/anthalib/wtype/liquid.go
```diff
@@ -469,7 +518,7 @@ func (lhc *Liquid) DNASequences() ([]DNASequence, error) {
 	return lhc.getDNASequences()
 }
 
-// SetVolume adds a volume to the component
+// SetVolume sets the volume to the component, inflating or deflating source volumes to match
```
m-i-sadowski: Seems slightly misleading that this comment says it changes something which only changes as a side-effect. I'd be inclined to find a better place to explain this side effect as this looks really confusing as it stands. 

typo: to --> of

haydnKing: oops, this changed and the comment didn't get updated - no side effects here

antha/anthalib/wtype/liquid.go
```diff
@@ -491,7 +540,16 @@ func (lhc *Liquid) Name() string {
 
 // SetName adds the specified component name to the component.
 func (lhc *Liquid) SetName(name string) {
```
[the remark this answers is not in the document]
haydnKing: So I think the point here (which isn't captured by the comment) is that `cmp.SetName` isn't just meant to change `cmp.CName`, but also to mark that this liquid is meaningful and should show up in the final sources - e.g. if you make a master mix and then do `cmp.SetName("awesome master mix")` you'll see some volume of "awesome master mix" show up in the mix view

I suppose this could be handled by the mixing behaviour (if you're mixing a liquid which has a meaningful name, create a new `SourceLiquid` to store it). I experimented a little with that, but ran into issues due to the strange things Subcomponents does to CName.

antha/anthalib/wtype/liquid.go
```diff
@@ -1154,3 +1213,153 @@ func (l1 *Liquid) EqualTypeVolumeID(l2 *Liquid) bool {
 
 	return l1.ID == l2.ID
 }
+
+// Sources returns a representation of the source components which were combined to produce this liquid
+func (l *Liquid) Sources() LiquidSources {
+	return l.sources.scaleToVolume(l.Volume())
+}
+
+// LiquidSource represents a source component of a liquid, that is either a volume of an input liquid or a liquid with a meaningful name
```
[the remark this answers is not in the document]
haydnKing: yeah, that would probably make sense

antha/anthalib/wtype/liquid.go
haydnKing: hmm, so I think it's safe because we make a copy before giving these to the user and also make sure to replace when we need to... but actually making them values would probably be simpler

antha/anthalib/wtype/liquid.go
m-i-sadowski: Ah. I assumed something was going on when volumes of subcomponents were asked for which would make this true... glad I was wrong

antha/anthalib/wtype/liquid.go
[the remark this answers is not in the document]
haydnKing: Yes, I completely agree - the problem is that we seem to have this mechanism whereby Subcomponents will update the CName automatically upon mixing to a name that isn't actually very helpful for users. I'd like to get rid of it, but because I'm not in a position to completely replace subcomponents, I've not tried... let's talk this through

msackman-synthace:
the go standard is test artifacts should be in a directory called `testdata`. See `go help test | grep testdata`

microArch/scheduler/liquidhandling/schemas/layout.schema.json
```diff
@@ -265,25 +265,25 @@
...
-				"components": {
-					"description": "The sub-components which make up this liquid",
+				"sources": {
+					"description": "describes the liquids which were combined to form this liquid in arbitrary order",
 					"type": "array",
 					"items": {
-						"description": "A sub component",
+						"description": "A source volume",
 						"type": "object",
-						"required": ["name", "concentration", "unit"],
```
chr1sgrant: why not include both volume and concentration?

chr1sgrant: This data-mvl will rely on being able to access sub component concentration information; preferably from this actions.json file

haydnKing:
Closing as I think it'd be better to have another run at this, probably using the slice method that Sid suggested

PR #1067, opened 2019-05-07
[AC-71] Reform file creation

msackman-synthace:
In a large number of places we create files and then write to them. We expect those files to not already exist, which means we should be doing O_CREAT|O_EXCL, which we were not.
    Additionally, we were using 0400, 0600 and 0700 as the mode to set on the created files. This is awkward: we should be using the defaults (which I believe are 0444, 0666 and 0777) and then relying on the system imposing the user's umask (typically 0022) to reduce those.
    
    Testing:
    The normal `rm -rf /tmp/foobar && ./composer -outdir /tmp/foobar -keep -linkedDrivers repositories.json workflow3.json.sample gilsonOnly.json.sample` works before and after with sensible permissions appearing.
    Setting umask to 0077 and rerunning does correctly reduce the mode of the created files.

### DIFF
--- a/composer/compile.go
+++ b/composer/compile.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/antha-lang/antha/utils"
 	"github.com/antha-lang/antha/workflow"
 )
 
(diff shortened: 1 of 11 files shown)


msackman-synthace:
https://console.cloud.google.com/cloud-build/builds/c2081cd3-9f6e-4dd5-9636-9f1015e944e4?project=antha-images

build passed.

PR #1068, opened 2019-05-07
[AC-45] Add protobuf support

simonwhitaker:
This adds a provider plugin for converting antha-runner's SimulateRequest protobuf file type to a 2.0 workflow. See AC-45 for details.

## NOTE: INCOMPLETE IMPLEMENTATION

There is still a little work to do translating plate and plate type values between the inventory microservice namespace and the antha.wtype namespace. See TODO comments in the code. I don't understand the background well enough to do this intricate bit of work correctly. The code in https://github.com/Synthace/antha-runner/blob/master/aconv/lhplatetype.go suggests that this is non-trivial.

### Unit tests

```
go build ./...
go test ./...
```

### Integration tests

Download a request.pb file from the execution details for a job, then:

```
$ migrate -from request.pb -format protobuf \
    -gilson-device gillian -outdir /tmp/AC-45 \
    /path/to/repositories.json
```


workflow/simulaterequestpb/provider.go
```diff
@@ -0,0 +1,303 @@
...
+			ref := param.GetReference()
+			if len(ref) == 0 {
+				return pset, fmt.Errorf("Param %v has no data and no reference", param.GetName())
+			}
+			// TODO: do something with ref. But what? Looks like it could be a datarepo: URL
+			// (https://github.com/Synthace/antha-runner/search?q=Param_Reference&unscoped_q=Param_Reference)
+			// - but that code path no longer exists in the feature/future_sanity world.
+			//
+			// 1. Is that use case still valid?
+			// 2. Are there any other use cases for reference params?
```
simonwhitaker: @msackman-synthace Any thoughts on this? Have we in fact killed the `reference` use case in feature/future_sanity, or are there things elsewhere that rely on it?

msackman-synthace: Those lines that you've found are actually on the output side - i.e. after the simulation has run and we're doing the nutty rewriting again of the simulation response.
The corresponding stuff on the input side is in microservice: https://github.com/Synthace/microservice/blob/master/cmd/scheduler/request/populate.go#L287
and if you look a few lines up at https://github.com/Synthace/microservice/blob/master/cmd/scheduler/request/populate.go#L254 then that stuff is coming from GetRaw(), not GetReference().
So, I *think* that GetReference() is just not used at all.

simonwhitaker: > So, I _think_ that GetReference() is just not used at all. So I think I would error if we ever come across a Reference.

Nice, thanks. It definitely smelt like a premature complication to me.

workflow/simulaterequestpb/provider.go
```diff
@@ -0,0 +1,307 @@
...
+	for i, plate := range plates {
+		result[i] = &wtype.Plate{
+			ID:         plate.Id,
+			Type:       wtype.PlateTypeName(plate.Type),
+			PlateName:  plate.Name,
+			Loc:        plate.Location,
+			WlsX:       int(plate.WlsX),
+			WlsY:       int(plate.WlsY),
+			Wellcoords: nil, // FIXME! Expecting []*wtype.LHWell, plate.WellCoords is []string
+			// TODO: Do we need to do anything with plate.Contents and plate.Barcodes??
```
simonwhitaker: We need someone more knowledgeable than I am in the lab side of things to finish this part of the implementation.

workflow/simulaterequestpb/provider.go
```diff
@@ -0,0 +1,307 @@
...
+		Types:                types,
+		InstancesConnections: connections,
+	}, nil
+}
+
+func translatePlateTypes(plateTypes []*inventorypb.PlateType) wtype.PlateTypes {
+	result := wtype.PlateTypes{}
+
+	// TODO: translate from one plate type format to the other. Not trivial from
+	// what I can see.
```
simonwhitaker: We need someone more knowledgeable than I am in the lab side of things to finish this part of the implementation.

workflow/migrate/params.go
```diff
@@ -0,0 +1,79 @@
...
+	} else if f, err := bif.moveDataToFile(fm); err != nil {
+		return nil, err
+	} else {
+		return json.Marshal(f)
+	}
+}
+
+// hasData checks that structure is a valid bakedInFile (rather than an empty struct)
+func (bif *bakedInFile) hasData() bool {
+	return bif != nil && bif.Name != "" && len(bif.Bytes.Bytes) > 0
```
msackman-synthace: I suspect this is wrong. I think if someone uploads an empty file then this will return false because of the len check. I think this should be `&& bif.Bytes.Bytes != nil`. I'll manually test.

workflow/simulaterequestpb/provider.go
```diff
@@ -0,0 +1,307 @@
...
+			Tipboxes:  mc.DriverSpecificTipPreferences,
+			Tipwastes: mc.DriverSpecificTipWastePreferences,
+			Washes:    mc.DriverSpecificWashPreferences,
+		}
+	}
+	return &config, nil
+}
+
+func (p *Provider) getGilsonPipetMaxConfig() (workflow.GilsonPipetMaxConfig, error) {
+	if p.gilsonDeviceName == "" {
```
msackman-synthace: Huh?!

```
message SimulateRequest {
  // Devices to consider when planning this request.
  repeated Device available = 1;
```

Devices is defined in the protobuf message. You don't want `gilsonDeviceName` in this provider at all.

workflow/migrate/params.go
simonwhitaker: An interesting philosophical point. If the file is empty, does the parameter have data?

Anyway, I didn't write that one, just moved it from workflow/v1_2/workflow.go. It may still be incorrect, of course.

workflow/simulaterequestpb/provider.go
```diff
@@ -0,0 +1,382 @@
...
+			result.Labcyte.Devices[id] = p.getLabcyteInstanceConfig("520")
+		case "LabCyteEcho525":
+			result.Labcyte.Devices[id] = p.getLabcyteInstanceConfig("525")
+		case "LabCyteEcho550":
+			result.Labcyte.Devices[id] = p.getLabcyteInstanceConfig("550")
+		case "LabCyteEcho555":
+			result.Labcyte.Devices[id] = p.getLabcyteInstanceConfig("555")
+		case "TecanLiquidHandler":
+			result.Tecan.Devices[id] = p.getTecanInstanceConfig("Evo")
+
```
msackman-synthace: I think the conclusion from a discussion with @m-i-sadowski yesterday was that in the case where GetAvailable() returns > 1 mixer then we only migrate the first one, and just ignore the rest. That will match the behaviour currently on master.

simonwhitaker: Right-ho.

Are all devices mixers? e.g. if the first device I get is e.g. a QPCRDevice or a WriteOnlyPlateReader, do I need to migrate any more? If not, what's the logic around what is and is not a mixer in this context?

msackman-synthace: Currently, from the point of view of antha, the mixers are only Gilson, Tecan, Hamilton, Labcyte, CyBio.
Ahh, I'd failed to spot that you'd removed `assertOnlyOneMixer`. I would bring that back and thus ensure we meet that requirement.

simonwhitaker: 7a7faf8

PR #1076, opened 2019-05-08
[AC-45] Use aconv in antha-runner to do conversion of plate types and plates.

msackman-synthace:
Currently untested, so almost certainly not wise to merge!

@m-i-sadowski does this look right to you? I'm both happy and suspicious that it's so short!

Goes hand-in-hand with https://github.com/Synthace/antha-runner/pull/417


antha/anthalib/wtype/plate.go
```diff
@@ -643,6 +643,40 @@ func NewLHPlate(idGen *id.IDGenerator, platetype PlateTypeName, mfr string, nrow
 	return &lhp
 }
 
+func (p *Plate) ToPlateType() *PlateType {
```
msackman-synthace: This exists because the corresponding PR in antha-runner calls this!

msackman-synthace:
ok, certainly the inventory migration does appear to work - I get stuff out like this:

```
	"Inventory": {
		"PlateTypes": {
			"pcrplate_skirted_riser18": {
				"Name": "pcrplate_skirted_riser18",
				"Manufacturer": "Bio-Rad",
				"WellShape": "cylinder",
				"WellH": 5.5,
				"WellW": 5.5,
				"WellD": 15,
				"MaxVol": 200,
				"MinVol": 5,
				"BottomType": 1,
				"BottomH": 0,
				"WellX": 5.5,
				"WellY": 5.5,
				"WellZ": 15,
				"ColSize": 8,
				"RowSize": 12,
				"Height": 33.5,
				"WellXOffset": 9,
				"WellYOffset": 9,
				"WellXStart": 14.28,
				"WellYStart": 11.24,
				"WellZStart": 19.25,
				"Extra": {
					"ll_model": "{\"Quadratic\":\"\",\"A\":0.402,\"B\":7.069,\"C\":0}"
				}
			}
		}
	},
```

msackman-synthace:
Argh, I thought from the discussion yesterday that #1075 was meant to be abandoned. The problem there is that the antha inventory needs to be maintained so it can be xref'd when we build the input plates from the mixer config. I.e. we are passing state through, which I've tried to convey by the comments in this PR.

antha/anthalib/wtype/plate.go
```diff
@@ -643,6 +643,40 @@ func NewLHPlate(idGen *id.IDGenerator, platetype PlateTypeName, mfr string, nrow
...
+		Manufacturer: p.Mnfr,
+		WellShape:    p.Welltype.WShape.Type.Name,
+		WellH:        p.Welltype.WShape.H,
+		WellW:        p.Welltype.WShape.W,
+		WellD:        p.Welltype.WShape.D,
+		MaxVol:       p.Welltype.MaxVol,
+		MinVol:       p.Welltype.Rvol,
+		BottomType:   p.Welltype.Bottom,
+		BottomH:      p.Welltype.Bottomh,
+		WellX:        p.Welltype.Bounds.Size.X,
```
simonwhitaker: Aside: this is so confusing. X, Y, Z are typically coordinates. I would not expect to model well coordinates in a plate type, since a plate has many wells, and each well has different coordinates in at least two planes.

Maybe this is modelling the size of the well, as suggested by `p.Welltype.Bounds.Size`, but in that case the X, Y and Z are confusing, and I don't understand what the difference is (if anything) between WellX/Y/Z and WellH/W/D a few lines earlier.

If you happen to understand any of this, please add a comment here to help us all understand it in future.
I'd expect a *Size* to have width, height and depth.

msackman-synthace:
ok, I've just been testing this, it does work. I'll sort out the branch and add a suitable test.

PR #1077, opened 2019-05-09
Append: adding columns names based Append

levitskovich:
Summary: 

Adding columns names based Append, changing the interfaces to support both position based and columns names based Append.

Three new columns names based Append modes: `Inner` (returns only intersecting columns), `Outer` (returns all the columns, fills gaps with nulls) and `Exact` (requires the same schemas of all the tables (except columns order)).

Columns names based `Append` matches columns by their name, type and (in case of multiple columns with the same name and type) ordinal - i.e. first column `a int64` is matched with first column `a int64`, second one with second one, etc. The consequence of thus approach is that columns with same names but different types are simply not matched (and Inner/Outer append returns no errors). On the positive side, any table can be matched with itself.

Interface now looks like this: `t1.Append(t2).Inner()`, `t1.Append(t2).Positional()`.

Also renamed the global `AppendMany` -> `Append`.

Test plan: `go test ./...` (changed `TestAppend` accordingly)


antha/anthalib/data/append.go
```diff
@@ -4,67 +4,285 @@ import (
...
+
+// Inner append: appends tables vertically, matches their columns by names, types and ordinals.
+// The resulting table contains intersecting columns only.
+func (s *AppendSelection) Inner() (*Table, error) {
+	return s.append(appendInner)
+}
+
+// Outer append: appends tables vertically, matches their columns by names, types and ordinals.
+// The resulting table contains union of columns.
+func (s *AppendSelection) Outer() (*Table, error) {
```
chr1sgrant: under what circumstances would this return an error?

levitskovich: Yes, it is not evident from the interface. Now `Inner` and `Outer` append return an error only in case when tables list is empty. This actually means that their non-global versions (`table.Append.Inner` and `table.Append.Outer`) will never return an error. But since we use the same `AppendSelection` object for both global Append and table.Append, there is no way to reflect this in the interfaces. But yes, at least I can update the comments.

levitskovich: Done

antha/anthalib/data/append.go
```diff
@@ -78,22 +296,39 @@ func compareSchemasForAppend(schema1 Schema, schema2 Schema) error {
...
+			tSources = []appendSource{{
+				table: t,
+				proj:  newIdentityProjection(t.schema.NumColumns()),
+			}}
+		}
+
+		// applying the corresponding projection to the sources
+		for _, source := range tSources {
+			sources = append(sources, appendSource{
+				table: source.table,
```
twoodwark: So, here why are we not actually doing a `Project` on the `source.table`?

Our iterator should ideally be independent of series that we don't need from the source due to the projection.

( requires us to change the implementation of `Project`, a bit )

chr1sgrant:
Looks great, I've just incorporated into the Append_Data element and tested Inner, Outer, Exact and Positional variants and the results generated were as expected.

PR #1079, opened 2019-05-13
ANTHA-2772: file series data type

levitskovich:
Summary:

The `FileSeries` data type is added in order to support using multiple files (selected with queries to datarepo) as workflows input. The main reason for introducing a new type (instead of just using `[]wtype.File`) is that now it's probably the only way to signal to the front end that it's needed to fill the list of files by a query (rather than manually).

Test plan: run a workflow https://www.dropbox.com/s/l5yf1utc0jpb8xv/ReadFileSeries.json with the branch https://repos.antha.com/antha-ninja/elements-westeros/tree/iuk/fileseries .


levitskovich:
Looks like it makes sense to push this code to master now - even without appropriate changes in the microservices - because, at least, it would allow us to create test bundles in the UI.

levitskovich:
P.S. This is the code for the old (pre-antha-core) world. Looks like we anyway have to start with it - because the UI currently almost doesn't support the new world.

antha/anthalib/wtype/fileseries.go
```diff
@@ -0,0 +1,6 @@
+package wtype
+
+// FileSeries represents a set of files - likely produced by the same device.
```
twoodwark: Let's add some more context here.

we should say that it's specifically intended for data processing usecases where 
* an entity such as a device is related to all the files
* the files implicitly form a series with the same data schema

PR #1080, opened 2019-05-14
Remove driver from LHProperties

mabn:
Driver inside LHProperties is pretty weird given that they are being duplicated.
I've moved it to the liquid handler to make it nicer. It also makes it more
obvious that tests run without any driver.


msackman-synthace:
FWIW, this looks fine by me. One thing though: I would probably make the field public, then you can avoid the wrapper method around GetOutputFile - there are no special semantics around accessing that field, so no need to hide it imo.

mabn:
I'd rather keep it hidden, driver has lost of methods and they shouldn't be accessible.

PR #1083, opened 2019-05-15
[AC-74] Detect when commit object is not found, and error more helpfully

msackman-synthace:
This also forced me to do some light refactoring and remove some duplication.
Tested with all permutations (both legal and illegal) of repositories.json.


msackman-synthace:
Happy to add the repository name itself if you would like (I've just seen you've suggested this in the ticket). I've only not currently because doing so would require a little remodelling of the structs. But happy to do that if you want me to.

PR #1084, opened 2019-05-16
Sid/multi arg prompts

m-i-sadowski:
This PR is the antha-core companion to https://github.com/antha-lang/antha/pull/1039 and should bring antha-core into parity with master as far as prompt syntax is concerned.

It is co-dependent on https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/807 as elements need to be updated to modify the calls to the various functions affected. 

In order to get this to work I have had to reinstate one previously changed behaviour in codegen, the check for 'bundles' which, if it doesn't find a single device which can perform all actions, returns a human driver. 

This is an area of the code which is now poorly understood and, I think, largely non-functional in terms of its original intentions. As such it is a little difficult to guarantee this is safe, however I can confirm that the expected behaviour still occurs for the workflows which were previously broken as I have manually compared the generated sqlite files. 

Additionally all test workflows pass as before.


execute/intrinsics.go
```diff
@@ -87,50 +85,60 @@ func Incubate(lab *laboratory.Laboratory, in *wtype.Liquid, opt IncubateOpt) *wt
...
-	Message     string
-	WaitTime    wunit.Time
+	Components   []*wtype.Liquid
+	ComponentsIn []*wtype.Liquid
+	Message      string
+	WaitTime     wunit.Time
+}
+
+func updateLiquidIDs(lab *laboratory.Laboratory, in []*wtype.Liquid) []*wtype.Liquid {
+	r := []*wtype.Liquid{}
```
msackman-synthace: I would always try to size these perfectly up front and then make use of the index in the subsequent range - to my eyes makes it more obvious that the input order and size corresponds to the output order and size. Plus it guarantees only 1 malloc.

instructions/maker.go
```diff
@@ -2,28 +2,24 @@ package instructions
 
 import (
 	"encoding/json"
-	"sync"
-
 	"github.com/antha-lang/antha/antha/anthalib/wtype"
+	"sync"
```
msackman-synthace: There are a couple of places where this has happened. Obviously, cosmetic, but us all using goimports to organise these in the same way does help keep diffs a little shorter.

microArch/driver/liquidhandling/messageinstruction.go
```diff
@@ -23,10 +23,15 @@ func NewMessageInstruction(lhi *wtype.LHInstruction) *MessageInstruction {
...
+	pt := make(map[string]*wtype.Liquid)
+
 	if lhi != nil {
+		for i := 0; i < len(lhi.Inputs); i++ {
+			pt[lhi.Inputs[i].ID] = lhi.Outputs[i]
+		}
 		msi.Message = lhi.Message
 		msi.WaitTime = lhi.WaitTime
-		msi.PassThrough = lhi.PassThrough
+		msi.PassThrough = pt
```
msackman-synthace: Ahh that's a shame - I was hoping PassThrough was going to be completely removed by this change.

microArch/scheduler/liquidhandling/executionhelpers.go
```diff
@@ -168,21 +175,17 @@ func aggregatePromptsWithSameMessage(idGen *id.IDGenerator, inss []*wtype.LHInst
...
 			ins.Message = prompter.Message
 			ins.WaitTime = prompter.WaitTime
-			ins.AddOutput(wtype.NewLHComponent(idGen))
 			for _, ins2 := range ar {
-				for _, cmp := range ins2.Inputs {
-					ins.Inputs = append(ins.Inputs, cmp)
-					ins.PassThrough[cmp.ID] = ins2.Outputs[0]
-				}
+				ins.Inputs = append(ins.Inputs, ins2.Inputs...)
+				ins.Outputs = append(ins.Outputs, ins2.Outputs...)
```
msackman-synthace: hmm, the api is a bit unfortunate. We have AddOutput and AddInput as methods, which makes you think that maybe they're doing something useful - validation or something, but then here we're just directly accessing the fields.
I would consider removing AddOutput, and replacing both with `func (inst *LHInstruction) AddOutputs(cmps ... *Liquid) {`

microArch/scheduler/liquidhandling/prompt_test.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+}
+
+var counter int = 0
+
+func getID() string {
+	id := fmt.Sprintf("ID-%d", counter)
+	counter++
+	return id
+}
```
msackman-synthace: Sorry, I really don't like this: I don't like that the test avoids using the public API (NewLHComponent) and reimplements existing functionality in the form of ids. It just makes the test more fragile and requires more implicit knowledge to understand that the test is correct.

msackman-synthace:
Oh, plus CI doesn't appear to like you just yet:
`Step #2 - "docker-build": /go/pkg/mod/github.com/antha-lang/antha@v0.0.1-alpha.0.20190516110822-13cb4293e567/microArch/scheduler/liquidhandling/prompt_test.go:292:15: undefined: buildInstructionChain`
:)

instructions/maker.go
m-i-sadowski: Actually I think I made the conscious decision to remove the whitespace. This makes a bit more sense to me. Wasn't aware we'd reached a consensus on this really - my GoImports has mysteriously broken so I guess I can unbreak it.

microArch/driver/liquidhandling/messageinstruction.go
m-i-sadowski: We still need something equivalent here. Happy to rename it if you prefer.

msackman-synthace: Nah - if the functionality is still required then so be it.

instructions/maker.go
msackman-synthace: I suspect it's consensus by omertà ;)

microArch/scheduler/liquidhandling/executionhelpers.go
m-i-sadowski: Duly noted. However this is out of scope for this PR so I'll consider this separately.

msackman-synthace: Fair enough.

PR #1085, opened 2019-05-16
Antha 2772 file series antha core

levitskovich:
Merged the recent changes from `master` including the `FileSeries` data type definition + changed `File` to `*File` (do I understand right that antha-core uses pointers to files?)

### DIFF
--- a/antha/anthalib/wtype/fileseries.go
+++ b/antha/anthalib/wtype/fileseries.go
@@ -1,0 +1,9 @@
+package wtype
+
+// FileSeries is a datatype representing a set of files for data processing usecases:
+// - the files are related to the same entity (likely to the same device);
+// - the files implicitly form a single dataset - i.e. they likely have the same data schema;
+// A parameter of type FileSeries signals to the UI that it should be set using a special query dialog (rather than manually filling the list of files).
+type FileSeries struct {
+	Files []File
+}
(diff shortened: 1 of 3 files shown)


msackman-synthace:
As discussed on slack, we suspect there need to be changes to migrate, both for the v1.2 json format, and for the simulate request protobuf format.

msackman-synthace:
@levitskovich for testing the protobuf form: *if* you can create a workflow in the ui and run it (old world - not antha-core), then you should be able to grab the request.pb from execution details, or, failing that, if you have the job id then I can show you how to pull it out of the cloud storage stuff.

levitskovich:
Ah, I see, thanks, I'll try doing this - I have an old-style workflow which I can upload to the UI and simulate there to get a request.pb

levitskovich:
Protobuf migration seems to work as well.

Btw, do I understand that merging this is not going to affect non-antha-core simulations at ninja anyway? (remembering that Toby asked not to break ninja because of todays OXB demo)

msackman-synthace:
> Btw, do I understand that merging this is not going to affect non-antha-core simulations at ninja anyway? (remembering that Toby asked not to break ninja because of todays OXB demo)

Yes, this will have no effect on non-antha-core simulations on ninja, don't worry!

PR #1086, opened 2019-05-16
Copy over extra and instance information from a liquid added to a well

chr1sgrant:
Previously, these details were not copied over, even when the well was previously empty.
This change will copy these over when adding to an empty well.

The change has been tested when running a workflow which produced a new plate of liquids containing Extra field and set Instance status using DeclareInstance()
The result led to the expected behaviour where the extra values could be queried and the plate set up correctly; where prior to this change it did not.


antha/anthalib/wtype/lhwell.go
```diff
@@ -296,9 +296,21 @@ func (w *LHWell) AddComponent(c *Liquid) error {
...
 	w.Contents().Mix(c)
 
+	// in cases where a liquid is added to an empty well
+	// copy over Instance status and Extra which are not inherited via Mix
+	if wellEmpty {
+		// copy over instance and extra
+		for key, value := range c.Extra {
+			w.WContents.Extra[key] = value
+		}
+		if c.IsInstance() {
```
m-i-sadowski: instance is set via Extra, so this is unnecessary

antha/anthalib/wtype/lhwell.go
```diff
@@ -296,9 +296,21 @@ func (w *LHWell) AddComponent(c *Liquid) error {
 		//return fmt.Errorf("Cannot add %s to well \"%s\", well already contains %s and maximum volume is %s", c.GetName(), w.GetName(), curVol, maxVol)
 		fmt.Printf("Adding %s to well \"%s\", even though well already contains %s and maximum volume is %s\n", c.Summarize(), w.GetName(), curVol, maxVol)
 	}
+	wellEmpty := w.Contents() == nil
 
 	w.Contents().Mix(c)
 
+	// in cases where a liquid is added to an empty well
+	// copy over Instance status and Extra which are not inherited via Mix
+	if wellEmpty {
```
m-i-sadowski: Please move this to Liquid's mix method instead.

PR #1089, opened 2019-05-20
Rename output file to .tar.gz

mabn:
(this is a PR to feature/TecanUpdates branch)

At the moment the output is in tar.gz format, the file name should match that.
I've removed the second part of the name (instrs[0].BlockId) because it seems to always be blank.

see https://synthace.atlassian.net/browse/TES-25

I've run CA workflow to test it.


arnevdb1:
The build error reported is because Antha is not on the correct branch I think as it complains about 

> /go/pkg/mod/github.com/!synthace/antha-runner@v0.0.0-20190514132415-965ce6435b43/aconv/lhplatetype.go:138:27: not enough arguments in call to wtype.NewLHPlate have (*id.IDGenerator, wtype.PlateTypeName, string, int, int, wtype.Coordinates3D, *wtype.LHWell, float64, float64, float64, float64, float64) want (*id.IDGenerator, wtype.PlateTypeName, string, string, int, int, wtype.Coordinates3D, *wtype.LHWell, float64, float64, float64, float64, float64)

I think we can ignore

PR #1090, opened 2019-05-20
[AC-78] Generally improve all command documentation.

msackman-synthace:
The existing docs are actually pretty reasonable, with one or two oversights.
What's mainly missing is pointers to those docs from the command -h form. So address that.
And generally review and polish all documentation.

### DIFF
--- a/cmd/composer/doc.go
+++ b/cmd/composer/doc.go
@@ -24,8 +24,8 @@
 //  2. Checkout all the known repositories under outdir/src at the
 //     indicated revisions.
 //
-//  3. Transpile the transitive closure of all antha element files
-//     into Go code.
+//  3. Transpile the transitive closure of all required antha element
+//     files into Go code.
 //
 //  4. Use the merged and validated workflow to generate a main.go in
 //     outdir/workflow. This is the entry point for the execution of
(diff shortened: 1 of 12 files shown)


mabn:
These are nice improvements, but I'm not sure if it helps. What I mean is that after seeing:

```
$ ./migrate -from=/Users/marcinbiegan/repo/tecan_accuracy_test/v1.2/a.json
ts=2019-05-17T17:40:46.864556Z fatal="No workflow sources provided."
```

i though that the message is clearly wrong, I have provided a workflow source (a json file) after all.

I think that it'd be better for ./migrate to clearly state that it requires more details, both in --help and in go doc. 

Ideally:

1. Make `migrate --help` show the usage at the beginning of the output, similar to e.g. `bash --help`:
   ```
   $ ./migrate --help
   Usage: migrate [flags] workflow-snippet.json ...

   Longer options / flags:
   (...)
   ```

   It would make it clear that migrate requires at least one argument

2. "No workflow sources provided" should be rephrased to something else in this scenario. E.g. "At least one workflow-snippet with repository configuration is required".

3. I'd change the order of things in doc.go for `migrate` and start with the example and the note that "As a minimum, it is necessary to provide sufficient Repositories such that every element instance in the input historical workflow (version=1.2) can be located within one of the Repositories"

4. The documentation refers to the extra arguments as "paths to workflows" (--help),  "additional information (particularly Repositories) in the SchemaVersion=2.0 format", 
"workflow snippets in the current SchemaVersion=2.0". Maybe just call it "workflow snippet" in all those places.

msackman-synthace:
Ahh, understood. Thank you for the fuller description. Yes, agreed, I shall make further improvements.

msackman-synthace:
All done. (2) is difficult because it's an error that comes from code common to a lot of code paths. So I've just changed it to "No workflow provided".

PR #1091, opened 2019-05-21
Log long lines

haydnKing:
When logging, we previously used `bufio.Scanner` to read lines, but this fails for (actually not very) long lines such as those written when `printInstructions` is true. From the `bufio` docs: `Programs that need more control over error handling or large tokens, or must run sequential scans on a reader, should use bufio.Reader instead.`


composer/compile.go
```diff
@@ -232,11 +232,20 @@ func RunAndLogCommand(cmd *exec.Cmd, logger func(...interface{}) error) error {
 
 func drainToLogger(logger func(...interface{}) error, fh io.ReadCloser, key string) {
 	defer fh.Close()
-	scanner := bufio.NewScanner(fh)
-	for scanner.Scan() {
-		logger(key, scanner.Text())
+
+	reader := bufio.NewReader(fh)
+	// the first call to ReadLine() returns an empty string
+	bs, partial, err := reader.ReadLine()
```
msackman-synthace: Interesting.
Given the godocs around ReadLine, and the complexity here of handling partial (sure, it's not complex, but it would be nice not to have to bother), I'm curious whether we should do as the docs suggest and use ReadString('\n') instead.

haydnKing: Yeah, honestly I'm not too impressed by the bufio API - if we use ReadString('\n') we have to remember to deal with the case that the input isn't terminated by '\n'... though that's probably simpler to I'll change back to that

haydnKing: should be better. It's possible to avoid the while loop, but less readable

msackman-synthace:
This seems totally fine to me.
We have a slightly conflict going on though... which is the subject of another ticket / refactoring, but I just want to make sure we're on the same page here.
Basically, given potential concurrency stuff going on with elements etc, I'm not a fan of formatted output going out of stdout/stderr at all. So to me, ideally stuff like printInstructions should just be dumped straight to a file, and then it can be in a richer format and not have any risk of getting interspersed with other output going on.
I full see the need for structured output for diagnostics and debugging, but I do think that stdout/stderr is the wrong place for it generally, so slowly we should probably move to a different system to get that data safely out.

PR #1092, opened 2019-05-21
add Q Instruments Cool Plate as riser

chr1sgrant:


inventory/testinventory/serialize/devices.go
```diff
@@ -277,6 +281,39 @@ var defaultDevices = map[string]device{
...
+					},
+					{
+						Name:          "strip_tubes_0.2ml",
+						SpecialOffset: 0.0,
+					},
+				},
+			},
+		},
+		Properties: devices.Shaker[qCoolPlate],
+		PositionConstraints: map[string][]string{
```
m-i-sadowski: Be aware these do not work via the UI ... command-line should be fine

PR #1093, opened 2019-05-21
[AC-58] Reenable linter for CI

msackman-synthace:


composer/compile.go
```diff
@@ -222,41 +226,37 @@ func StartAndLogCommand(cmd *exec.Cmd, logger func(...interface{}) error) error
...
 	reader := bufio.NewReader(fh)
 	for {
-		if s, err := reader.ReadString('\n'); err == nil {
-			logger(key, strings.TrimSuffix(s, "\n"))
-		} else if err == io.EOF {
-			if s != "" {
-				logger(key, strings.TrimSuffix(s, "\n"))
+		line, err := reader.ReadString('\n')
+		line = strings.TrimSuffix(line, "\n")
+		if line != "" {
```
haydnKing: so I wondered about this. I'm not too against dropping all empty lines, but is might be a bit confusing if someone adds `fmt.Println("\n\n\nsome debug statement that I can easily find amidst a sea of nonsense")` as I am wont to do on occasion

PR #1094, opened 2019-05-22
Update cumulative times for prompts.

matttgregg:
See AP-3 - Cumulative timing is not being collected for prompts with timings.
As a result in the UI (via exported actions.json) the estimated time for
actions is incorrect for anything following a timed prompt.

Code collecting cumulative timings is per instruction type - needs to be updated
following introduction of timed prompt values.

Test Plan:
  Tested locally, actions.json shows correct cumulative times for large prompt times.
  To be observed in UI following deployment.


matttgregg:
Good idea - I was just wondering what testing support we had around this. I'll add one in.

PR #1096, opened 2019-05-22
Fix Tip Selection in Planner

haydnKing:
Previously, the mechanism through which tips were chosen by the planner was broken in a few ways:
- reported loaded tips were wrong for some devices (esp. Gilson)
- couldn't correctly handle non-contiguous tip loading
- couldn't correctly handle multi-tip-type loading
- was generally pretty intractable and spread between `microArch/driver` and `wtype`

This change reimplements tip selection, abandoning the attempt to predict tip choice in a generic way in favour of small device specific functions. Ultimately, these should be provided as callbacks in response to `Configure`, but until this is technically possible they are implemented and tested in `microArch/scheduler/liquidhandling/choosetips.go`.

Affected JIRA tickets:
- fixes: https://synthace.atlassian.net/browse/AP-6 
- fixes: https://synthace.atlassian.net/browse/ANTHA-2648 
- related: https://synthace.atlassian.net/browse/ANTHA-2046
- a big step towards: https://synthace.atlassian.net/browse/ANTHA-2815

Testing:
- thorough unit tests written for new code (hamilton & gilson implementations)
- actions summary update to verify that new tip ordering shows up in summary
- *haven't* run whatever new-style antha-tester is called because I haven't got it working yet

PS. there's a supporting change here: https://github.com/Synthace/antha-runner/pull/425


composer/generate_mod.go
```diff
@@ -105,9 +108,13 @@ func renderRepositoryMod(w io.Writer, repoName workflow.RepositoryName) error {
...
-	Repositories workflow.Repositories
-	AnthaDir     string
-	ReplaceAntha bool
+	Repositories   workflow.Repositories
+	AnthaDir       string
+	ReplaceAntha   bool
+	ReplaceRunner  bool
+	RunnerDir      string
+	ReplacePlugins bool
+	PluginsDir     string
```
msackman-synthace: hah, I wondered if we'd eventually need these. Fine, but consistency in field ordering would be ideal :)

composer/generate_mod.go
```diff
@@ -121,6 +128,16 @@ func newWorkflowMod(repos workflow.Repositories) *workflowMod {
...
+		runnerDir := filepath.Join(filepath.Dir(wm.AnthaDir), "antha-runner")
+		if _, err := os.Stat(runnerDir); !os.IsNotExist(err) {
+			wm.ReplaceRunner = true
+			wm.RunnerDir = runnerDir
+		}
+		pluginsDir := filepath.Join(filepath.Dir(wm.AnthaDir), "instruction-plugins")
+		if _, err := os.Stat(pluginsDir); !os.IsNotExist(err) {
+			wm.ReplacePlugins = true
+			wm.PluginsDir = pluginsDir
+		}
```
msackman-synthace: Hmm. FWIW, I consider this the "wrong" way to do it.
Ideally, we could modify `AnthaModule()` so that it just takes a package name and searches for that.
But the problem is that in the composer runtime, we don't currently have a dependency on instruction-plugins or antha-runner - we only have a dependency on that stuff when we get to the workflow itself.
We could artificially add a dependency (eg a `_` import), or we could say "sod it, this shouldn't exist for long because we're about to merge these repos".
Up to you.

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+	"fmt"
+	"github.com/pkg/errors"
+	"sort"
+	"strings"
+
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+)
+
+type TipSource struct {
+	TipboxID    string           // the ID of the tipbox from which to take tips
```
msackman-synthace: one day, I'm going to do a sed to get us to standardise on `TipBox` and `TipWaste` and not `Tipbox`... ;)

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+		tipTypes[tt] = true
+		tipType = tt
+	}
+	if len(tipTypes) != 1 {
+		s := make([]string, 0, len(tipTypes))
+		for tt := range tipTypes {
+			s = append(s, string(tt))
+		}
+		sort.Strings(s)
+		return nil, errors.Errorf("Gilson device can only handle one tip type at a time: cannot load %s", strings.Join(s, ", "))
```
msackman-synthace: hmm, this is very similar to the code above in Error(). I wonder if we should have `type TipTypes []TipType` and then have a `SortByTypeName()` method on that, or maybe just a `String()` which does the sort internally

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	// 2: ChannelIndexes must be [0,7]
```
msackman-synthace: ahh, uint3 ;)

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+	invalid := make([]string, 0, len(channelMap))
+	for ci := range channelMap {
+		if ci < 0 || ci > 7 {
+			invalid = append(invalid, fmt.Sprintf("%d", ci))
+		}
+	}
+	if len(invalid) > 0 {
+		sort.Strings(invalid)
+		return nil, errors.Errorf("Gilson device has only 8 channels: invalid channels %s", strings.Join(invalid, ", "))
```
msackman-synthace: eh, fwiw, I think you're being too nice with the error msgs - the extra effort you're going to to sort them (but as strings... so "10" would come before "2" which is probably surprising) I think is probably unnecessary - if it were me I'd just build up []int and then `errors.Errorf` with a `%v`

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	// 3: channels must be contiguous, starting from 0
+	missing := make([]string, 0, len(channelMap))
+	for i := 0; i < len(channelMap); i++ {
+		if _, ok := channelMap[ChannelIndex(i)]; !ok {
+			missing = append(missing, fmt.Sprintf("%d", i))
+		}
+	}
+	if len(missing) > 0 {
```
msackman-synthace: perfectly happy with this - the code is clearly correct and highly readable.

But, in your previous loop, you could have just kept track of the min and max and then here say:
1. min must be 0
2. len must eq max

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+		sort.Strings(missing)
+		return nil, errors.Errorf("Gilson device can only load tips on contiguous channels: skipping channels %s", strings.Join(missing, ", "))
+	}
+
+	// 4: head is 0 or 1
+	if !(head == 0 || head == 1) {
+		return nil, errors.Errorf("Unknown head %d", head)
+	}
+
+	// not let's figure out how to load these tips
```
msackman-synthace: s/not/now/

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// let's try and find a tipbox with enough tips
+	var box *wtype.LHTipbox
+	for _, bx := range tipboxes {
+		if wtype.TipType(bx.Tiptype.Type) == tipType && bx.N_clean_tips() >= len(channelMap) {
+			box = bx
```
msackman-synthace: add break; or, if you intend to select the last, add comment.

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+	}
+	if box == nil {
+		return nil, NewTipNotFoundError(tipType)
+	}
+
+	// loading direction for the heads -- nb. this depends on robot configuration, but likely head 0 is loaded on the left and head 1 on the right
+	direction := map[int]wtype.HorizontalDirection{
```
msackman-synthace: I would prefer just changing the type to []wtype.HorizontalDirection - everything else stays the same. It's just a defence in depth thing - if for some reason head has been changed and is now < 0 or > 1 then we will panic. Whereas as it stands, you'll get the default value

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// some assertions
+
+	// 1. we only support head = 0
+	if head != 0 {
+		return nil, errors.Errorf("head %d not supported", head)
+	}
+
+	// 2. channelIndexes must be [0,7]
+	invalid := make([]string, 0, len(channelMap))
```
msackman-synthace: ahh, uint3 ;)

microArch/driver/liquidhandling/choosetips.go
```diff
@@ -0,0 +1,227 @@
...
+	}
+	if len(invalid) > 0 {
+		sort.Strings(invalid)
+		return nil, errors.Errorf("invalid channel indexes: %s", strings.Join(invalid, ", "))
+	}
+
+	// get unique tiptypes
+	tipTypeMap := make(map[wtype.TipType]bool, len(channelMap))
+	for _, tt := range channelMap {
+		tipTypeMap[tt] = true
```
msackman-synthace: again, nothing at all wrong with this code - clear, correct etc.
I would probably have built both in one pass:
```go
if !tipTypeMap[tt] {
  tipTypeMap[tt] = true
  tipTypes = append(tipTypes, tt)
}
```

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -3099,61 +3065,128 @@ func GetFirstDefined(sa []string) int {
...
 	}
 
-	r := ""
+	// this is how we decide which TipChooser to use
+	// Ideally the function should be provided as a callback and be attahced to params already, currently blocked by the need to serialise
+	tipChooser, ok := map[string]TipChooser{
+		"gilson":   chooseTipsGilson,
+		"hamilton": chooseTipsHamilton,
+		"tecan":    chooseTipsHamilton,
+	}[strings.ToLower(params.Mnfr)]
```
msackman-synthace: hmpf, where do these strings come from? It would be lovely to standardise of the strings and consts in target/mixers.go

msackman-synthace:
And ya build fails:
`Step #4 - "docker-build": /go/pkg/mod/github.com/!synthace/antha-runner@v0.0.0-20190514132415-965ce6435b43/aconv/lhtipbox.go:15:3: cannot use in.Tiptype.Type (type wtype.TipType) as type string in field value`

nb you can do `go get github.com/Synthace/antha-runner@my/branch/name` and commit that here, which should unbreak the build.

microArch/driver/liquidhandling/choosetips.go
haydnKing: I'll admit that wanting to write nice error messages is a bit of a Achilles heel of mine, but I also know someone who likes to check that `err.Error() == expectedString` in unit tests who'd probably appreciate stable error messages ;-)

composer/generate_mod.go
haydnKing: yeah I think given that this should be pretty short lived I'm not too bothered to resolve that

microArch/driver/liquidhandling/choosetips.go
haydnKing: good catch here!

microArch/driver/liquidhandling/compositerobotinstruction.go
haydnKing: I'd forgotten about those consts. seems unfortunately they're only returned by the plugins and aren't stored in the LHParameters... adding them feels a bit out of scope since hopefully this map won't live for long?

haydnKing:
Ok, I think this is worthy of another look. I've just run element test bundles with no changes in behaviour

driver/liquidhandling/objects_test.go
```diff
@@ -163,17 +159,16 @@ func SetUpTipsFor(lab *laboratory.Laboratory, lhp *liquidhandling.LHProperties,
...
+				tip := tb.Tips[0][0]
+				str := tip.Mnfr + string(tip.Type) + tip.MinVol.ToString() + tip.MaxVol.ToString()
+				if seen[str] {
+					return nil
+				}
 
-			seen[str] = true
-			lhp.Tips = append(lhp.Tips, tb.Tips[0][0])
+				seen[str] = true
+				lhp.Tips = append(lhp.Tips, tb.Tips[0][0])
```
msackman-synthace: cosmetic: s/tb.Tips[0][0]/tip/ as you already have the local variable

microArch/driver/liquidhandling/choosetips_test.go
```diff
@@ -0,0 +1,713 @@
...
+		TipType: tct.TipType,
+		Tips:    tips,
+	}
+}
+
+type TipChooserTest struct {
+	Name    string
+	Chooser TipChooser // the function to test
+	// Tipboxes define what tipboxes should be available for the test and what state they should be in.
+	// Tipboxe IDs are replaced with "1","2","3", etc for testing
```
m-i-sadowski: Tipboxe -> Tipbox

microArch/driver/liquidhandling/choosetips_test.go
```diff
@@ -0,0 +1,713 @@
...
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/go-test/deep"
+
+	"github.com/antha-lang/antha/antha/anthalib/wtype"
+)
+
+// TipChooserTipbox defines the state of a tipbox for testing, should be an array of 8 strings each of length 12
```
m-i-sadowski: Generally speaking I love this set of tests. The TipLocations thing is a little bit clunky IMO though. The issue is for some of the more complicated tests I need to count 1s to see where tips are to determine whether or not I agree with the proposed result. 

I'm inclined to suggest we should at some point (not now) improve write- and read-ability by having something like FillFrom(C10, COLS) so that you can easily tell whether, e.g., C10 is supposed to be full or empty.

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -1403,6 +1386,61 @@ func (ins *SuckInstruction) GetParameter(name InstructionParameter) interface{}
 	}
 }
 
+func getMixVolumes(labEffects *effects.LaboratoryEffects, pol wtype.LHPolicy, policyKey string, insVols []wunit.Volume, insWhat []string, tipTypes []wtype.TipType, prms []*wtype.LHChannelParameter) ([]wunit.Volume, error) {
```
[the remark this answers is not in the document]
haydnKing: yeah, I found myself having to make an identical fix to identical code so figured it would save time to just make a function

microArch/driver/liquidhandling/compositerobotinstruction.go
```diff
@@ -3099,61 +3065,127 @@ func GetFirstDefined(sa []string) int {
...
+			return nil, err
+		} else if usetiptracking {
+			// in "DriverTipTracking" mode the user has to refresh the tipboxes themselves, so
+			// find the first non-empty instance of tips of the missing type(s) and refresh them
+		TIPTYPE:
+			for _, tt := range tipErr.Missing {
+				for _, tb := range params.TipboxesByPreference() {
+					// checking for missing tips is _probably_ unnecessary here since we're not likely to be
+					// loading more than one tipbox at a time in the immediate future
+					if tb.Tiptype.Type == tt && tb.N_clean_tips() < tb.NRows()*tb.NCols() {
```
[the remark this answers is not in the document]
haydnKing: So the real reason for this is if there's some case where a full tipbox either can't be used for this specific load for some arbitrary callback-defined reason. 
In that case, refilling the first tipbox isn't necessarily enough because it might not be the cause of the problem.

I think the comment is more referring to the other case where we're loading a 384 head which requires 4 full tipboxes, I'll expand the comment.

microArch/driver/liquidhandling/choosetips_test.go
haydnKing: Yeah, I'm not 100% happy with this either, though I just added some comments which I think make this much better

microArch/driver/liquidhandling/compositerobotinstruction.go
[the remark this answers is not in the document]
m-i-sadowski: On discussion we agreed we'd effectively have the same check but make it the business of the tipbox to return an error if it can't be refreshed for some reason (e.g. it's full)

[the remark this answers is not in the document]
haydnKing: ok, this is now done. `initialize_tips` isn't nice, but at least it's hidden away nicely

m-i-sadowski:
Hurrah for the demise of nasty labyrinths

PR #1100, opened 2019-05-29
Add support for Hx1000F tips

haydnKing:
This also involved correcting the lower bound for each tips, and in the instruction plugin, fixing the lower bound of the head: https://github.com/Synthace/instruction-plugins/pull/29

nb. this is set to be merged into ANTHA-2815 (and then into future_sanity), since it needs the changes in the instruction plugin to work correctly


haydnKing:
Ps. I've tested this by running on the robot, and the 1000f tips were used correctly to transfer 500ul and 1000ul, though it exposed an issue with channel choice

inventory/tipboxes/library.go
```diff
@@ -176,19 +176,27 @@ func makeTecanTipBoxes(idGen *id.IDGenerator) []*wtype.LHTipbox {
 func makeHamiltonTipboxes(idGen *id.IDGenerator) []*wtype.LHTipbox {
 	var ret []*wtype.LHTipbox
 
-	//Filter tips
+	// Filter tips
```
m-i-sadowski: ???

I assume this is copypasta

PR #1102, opened 2019-05-30
Deterministic input plates order

mabn:
I was trying to create a test for Tecan but it was failing because the order
of input plates was randomized. This fixes the issue.


microArch/scheduler/liquidhandling/setupagent.go
```diff
@@ -277,6 +279,23 @@ func BasicSetupAgent(labEffects *effects.LaboratoryEffects, request *LHRequest,
...
 
+func sortedKeys(arg map[string]*wtype.Plate) (result []string) {
+	for k := range arg {
+		result = append(result, k)
+	}
+	sort.Strings(result)
+	return
+}
+
+// where are the generics..
```
haydnKing: I think the tidier way to do this would be
```
// LocationMap maps a component name to a slice of locations where that component is found
type LocationMap map[string][]string

// Keys return the sorted list of keys
func (lm LocationMap) Keys() []string {
   ...
}
```
and similar for `map[string]*Plate`

microArch/scheduler/liquidhandling/setupagent.go
```diff
@@ -277,6 +279,23 @@ func BasicSetupAgent(labEffects *effects.LaboratoryEffects, request *LHRequest,
...
+func sortedKeys2(arg map[string][]string) (result []string) {
```
haydnKing: it's usually better to do `result := make([]string, 0, len(arg))` in this sort of situation - by setting the capacity you can guarantee you won't have to re-initialise the array when you later append to it

microArch/scheduler/liquidhandling/setupagent.go
mabn: I did that, but haven't replaced all occurrences across the code with the new type.

One thing I noticed is that InputAssignments and OutputAssignments are only used during setup, so they could be not exported, or even removed from the LHRequest

haydnKing: removing them from the request would be a good idea actually - it's pretty polluted and it's clear these vars aren't actually a "request" as such!

PR #1103, opened 2019-05-31
Improve ChooseChannels to handle overlapping tips

haydnKing:
`ChooseChannels` determines which head and tips should be used for a particular transfer.

Previously each channel was chosen independently, and it behaved badly when multiple tip types had the same minimum volume.

This Change:
 - only returns one tiptype which is to be used for all channels
 - correctly distinguishes which tip to use for e.g. 0.5ul when two tips [0.5-50.0] and [0.5-300.0] are available
- avoids some floating point errors that were causing tranfering 100ul with 50ul tips to be done as 3x33.3ul transfers

There are some minor changes to an element due to the change to `wunit.MinVolume` / `wunit.MaxVolume`, with this change antha/cmd/element testing ran correctly 

## merge plan
1. ~remove_lhcp_id -> feature/future_sanity  (see https://github.com/antha-lang/antha/pull/1101 )~
2. fix_choosechannel -> feature/future_sanity . (this branch)


haydnKing:
See: https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/824

microArch/driver/liquidhandling/tests/choosechannel_test.go
```diff
@@ -1,174 +1,187 @@
...
-				} else if !prm.Maxvol.EqualTo(mxr) || !prm.Minvol.EqualTo(mnr) || tiptype != tpr {
-					return fmt.Errorf("Incorrect channel choice for volume %v\n\tGot %v %v %v\n\tWant %v %v %v",
-						vol.ToString(), prm.Minvol.ToString(), prm.Maxvol.ToString(), tiptype, mnr.ToString(), mxr.ToString(), tpr)
-				}
-			}
-			return nil
+func TestChannelChoiceGilson(t *testing.T) {
+	idGen := id.NewIDGenerator("ChooseChannelGilson")
+	(&ChooseChannelsTests{
+		Tips: []*wtype.LHTip{
```
m-i-sadowski: I'd like to see behaviour when the mid-range tip type is included here, just for completeness sake
Also a demonstration of the use of the HV tips for LV on Gilson would make sense to check that head / tip combinations work correctly

microArch/driver/liquidhandling/choosechannel.go
```diff
@@ -2,170 +2,106 @@ package liquidhandling
...
-	// we should always make sure we do not send a volume which is too low
-
-	if lhcp.Minvol.GreaterThanRounded(vol, 1) {
-		return 0
+// ScoreChannel score the channels in terms of its suitability to transfer a volume vol
+// Properties of the score function are:
+//   - score is undefined for channels which cannot transfer vol and false is returned
+//   - channels which can transfer vol in fewer transfers score better than those which take more
+//   - for the same number of transfers, the channel which transfers vol while being proportionally closest to its maximum scores better
+func ScoreChannel(vol wunit.Volume, lhcp *wtype.LHChannelParameter) (float64, bool) {
```
m-i-sadowski: As discussed I think this fails to be deterministic when there are tips with exactly the same volume range available
Please can you find a suitable place to check this isn't the case and error if it is?

microArch/driver/liquidhandling/choosechannel.go
```diff
@@ -2,170 +2,106 @@ package liquidhandling
...
+// ChooseChannels decide which combination of head/adaptor/tiptype should be used to transfer the given set of
+// volumes. Currently the same combination will be returned for each channel, except for channels where
+// vol[index] is zero, which are left empty
+// len(vols) should be <= the number of channels of at least one available head.
+func ChooseChannels(vols []wunit.Volume, prms *LHProperties) ([]*wtype.LHChannelParameter, []wtype.TipType, error) {
 
-		err := (mx.RawValue()-tv)/(mx.RawValue()-mn.RawValue()) + extra
+	var bestChannel *wtype.LHChannelParameter
+	var bestTipType wtype.TipType
+	bestScore := -math.MaxFloat64
```
m-i-sadowski: why are we maximising negatives and not minimising positives?

microArch/driver/liquidhandling/choosechannel.go
```diff
@@ -2,170 +2,106 @@ package liquidhandling
...
+		return nil, nil, wtype.LHErrorf(wtype.LH_ERR_VOL, "no tip chosen: volumes %v could not be moved by the liquid handler in this configuration\n\t%s", vols, strings.Join(config, "\n\t"))
 	}
 
-	if headchosen == nil {
-		return nil, nil, TipChosenError(vol, prms)
+	props := make([]*wtype.LHChannelParameter, len(vols))
+	tipTypes := make([]wtype.TipType, len(vols))
+	for i, vol := range vols {
+		if !vol.IsZero() {
+			props[i] = bestChannel.Dup()
```
m-i-sadowski: Not sure we really need to Dup here?

microArch/driver/liquidhandling/tests/choosechannel_test.go
```diff
@@ -1,174 +1,187 @@
...
-				return errors.New("channel was not nil for small volume")
-			}
-			if tip != nil {
-				return errors.New("tip was not nil for small volume")
-			}
-			if err == nil {
-				return errors.New("error not generated for small volume")
-			}
-			return nil
+func TestChannelChoiceOverlap(t *testing.T) {
```
m-i-sadowski: Ask Arne for any Tecan-motivated test cases as well?

I'm wondering about other cases too:

1) non-overlapping ranges (not done)
2) overlap, same minimum (done)
3) overlap, same maximum (not done)
4) overlap, one range contains the other, e.g. 0.5-100, 10-50 (not done)
5) tiling - three distinct, overlapping ranges (done)

in each case with a volume which should be chosen by either tip or none.

microArch/driver/liquidhandling/tests/choosechannel_test.go
```diff
@@ -1,174 +1,187 @@
...
+			NewChooseChannelsSingleTest(2.0, 0, "Low", false),
+			NewChooseChannelsSingleTest(5.0, 0, "Low", false),
+			NewChooseChannelsSingleTest(10.0, 0, "Low", false),
+			NewChooseChannelsSingleTest(20.0, 0, "Low", false),
+			NewChooseChannelsSingleTest(30.0, 1, "High", false),
+			NewChooseChannelsSingleTest(50.0, 1, "High", false),
+			NewChooseChannelsSingleTest(100.0, 1, "High", false),
+			NewChooseChannelsSingleTest(200.0, 1, "High", false),
+			NewChooseChannelsSingleTest(300.0, 1, "High", false),
+			NewChooseChannelsTest([]float64{0.5, 0.5, 0.5, 0.5}, []int{0, 0, 0, 0}, []string{"Low", "Low", "Low", "Low"}, false),
```
m-i-sadowski: couple more cases of multi choice? Given we don't have independence what do we see if we ask it for inconsistent vols?

microArch/driver/liquidhandling/tests/choosechannel_test.go
haydnKing: mid-range tip? I think the difficulty here is that we still test tip compatibility by checking that the tip range is contained by the head range, so any tip which spans the low-high head boundary should never be selected.
Similarly, the HV on LV tests would be essentially the same as this - the LVGilson200 definition looks like: 
`wtype.NewLHTip(idGen, "gilson", "LVGilson200", 1.0, 20.0, "ul", false, shp, 39.3)`
i.e. the volume range is essentially the same as what we're testing here

microArch/driver/liquidhandling/choosechannel.go
haydnKing: good question, will change that

microArch/driver/liquidhandling/tests/choosechannel_test.go
haydnKing: Ok so:
1. done
2. done
3. explicitly disallowed
4. done
5. done
Looks like the tecan case is pretty much identical to the hamilton one

microArch/driver/liquidhandling/tests/choosechannel_test.go
m-i-sadowski: Yeah actually I had in mind the DF30 but looking at it it's basically a DL10 with a higher minimum volume (which, then would be a case of two tips with the same maximum volume if we allowed that!)

haydnKing:
> I still have a tiny reservation about tips which do not exactly fit any available head but I think that has to remain as tech debt.

Yes - my plan with this is to allow the plugin to determine which tips are compatible with which head(s) once the refactor happens, but for now I think the current behaviour is sufficient.

PR #1105, opened 2019-06-03
Add plates '384 Armadillo Skirted' and 'N8010560 in Gilson 496' to reflect ninja updates

svasseur-synthace:
Some new plates added into the hardcoded list since Nuno wishes to use them.

We really have to clean the way plates are handled, this is really insane and a nightmare. There are definitions at multiple places, they do not always match and it's waaaaay too much manual steps to have something 'working', or at least usable by some people.


svasseur-synthace:
I've asked my Portuguese contact and he gave me some links, so adding them in the comments.

PR #1107, opened 2019-06-03
do update subcomponents even in cases where liquids have no sub compo…

chr1sgrant:
…nents


m-i-sadowski:
Can you give some context please? I'm not sure why this is necessary. 

It's currently failing various tests, so that needs sorting out anyway, but without any context it's hard to say whether the tests are right or this change is (although obvs I would guess the tests need updating)

PR #1109, opened 2019-06-04
Generate found test data for tests.

matttgregg:
It's necessary to add test data to files, however it is not trivial
to generate the test data.
By always exporting the test data, it becomes easy to create and
update workflow tests.

Test Plan:
Run standard workflow tests:
  * Check that we do not affect test times (through extra processing)
  * Check that we can easily update test bundles.


matttgregg:
Addition of output doesn't appear to affect runtime of test workflows.
Updates/test creation are fairly easily done via _jq_:

```
jq --slurpfile testData "comparisons.foundresults.json" '.Testing = $testData' workflow/workflow.json
```

matttgregg:
Main concern is whether this is easy enough for element maintainers...

laboratory/compare.go
```diff
@@ -13,6 +14,10 @@ import (
 
 // Compare compares output generated with any supplied test data in the workflow
 func (labBuild *LaboratoryBuilder) Compare() {
+	foundFile := filepath.Join(labBuild.outDir, "comparisons.foundresults.json")
```
mabn: Maybe this could be called "actual" results - when doing assertions the two things being compared are often called "actual" and "expected"

laboratory/compare.go
```diff
@@ -13,6 +14,10 @@ import (
 
 // Compare compares output generated with any supplied test data in the workflow
 func (labBuild *LaboratoryBuilder) Compare() {
+	actualFile := filepath.Join(labBuild.outDir, "comparisons.actual.json")
```
msackman-synthace: https://twitter.com/SeanTAllen/status/1133454305195974657

Sorry, I know this is cosmetic, but seeing `actual` all over the place is disturbing me.

"computed" or "generated" seem preferable to me.

matttgregg: I vaguely think that `actual` has a funny mouth feel if that helps. I'll go for `generated`.

PR #1110, opened 2019-06-05
Fixing the bug in ForeachKey

levitskovich:
Because of the bug in keys comparison, ForeachKey used to split any table into many one-row partitions (shame on me that we didn't have a test for this). So:
* fixed the bug
* rewrote the test in order to 1) check the number of partitions, 2) use Append instead of TableBuilder and 3) fix yet another bug in the test (the previous version  implicitly assumed that key columns are the first columns of a table)


twoodwark:
ugh, fix makes sense, let's fix the build

PR #1111, opened 2019-06-05
fix due to parquet-go interface change

levitskovich:
The parquet-go authors
* renamed packages `ParquetReader` -> `reader`, `ParquetWriter` -> `writer`, `ParquetFile` -> `source`
* all concrete implementations of `ParquetFile` are moved to another repo `github.com/xitongsys/parquet-go-source`


levitskovich:
after this is approved, I'll update antha-runner as well

PR #1112, opened 2019-06-06
Clean LHPolicies in preparation for the API refactor

haydnKing:
This got a bit big, but a lot if it is mechanical. Significant changes are:
- moved `microArch/driver/liquidhandling.GetPolicyFor(*LHPolicyRuleSet, ins RobotInstruction) LHPolicy` to `wtype/LHPolicyRuleSet.GetPolicyFor(ins)`
- safe, typed accessors for policy values
- more consistent overriding of policies with other policies
- LHCategoryCondition is now case insensitive so that setting a liquid type to "protoplasts" or "Protoplasts" both cause the same behaviour

This will enable
- Summary generation to stop depending on the physical simulator by reading from the LHPolicy instead
- `actions.json` to include the list of matching rules&policies and the order in which they were applied
- a step towards rationalising our policy system and maybe one day reading policy and rule info from configuration rather than hardcoding it

I've tested using `cmd/elements`, there's one three line deletion in repos which needs to happen in parallel: https://repos.antha.com/antha-ninja/elements-westeros/merge_requests/831


haydnKing:
@m-i-sadowski has given verbal consent to the outer facing changes in this PR (categories/liquid types not being case sensitive, better accessors), it just needs a code review. Annoyingly github's diff is pretty poor :-/ 

This should set us up to move away from a `map[string]interface{}` approach entirely.

Happy for this to wait until next week's drama has subsided

PR #1113, opened 2019-06-07
 Add partNr to PlateType

mabn:
Tecan plugin uses manufacturer+partNr to match Antha plates with the ones
defined in the EVOware.

Changes:
1. added partNr to PlateType
2. make the conversion from simulaterequest to workflow aware of partNr
3. add partNr to workflow 2.0 schema so it's possible to put it into Inventory
   section.

These changes are extracted from features/TecanUpdates branch, merging this will bring future_sanity and TecanUpdates branches a bit closer.

There are corresponding changes in antha-runner and instrumentation-plugins in feature/part-nr branches.


mabn:
The message above (merged) is about merging to TecanUpdates, not future_sanity, this PR is not merged yet.

mabn:
I assume that these are synonyms. 
Maybe we should call it catalogNumber /catalogNr in the code? But partNr was shorter.

